PR #873, opened 2015-02-25
DS-2474 METS format in OAI includes only the first author

helix84:
https://jira.duraspace.org/browse/DS-2474


aschweer:
as said on Jira, +1

PR #875, opened 2015-02-27
DS-2478 update Google Analytics tracking JS snippet

helix84:
https://jira.duraspace.org/browse/DS-2478
- update the Google Analytics snippet
- move it into an included file in JSPUI
- move it into a separate template in XMLUI
- clean up whitespace


helix84:
This update should be a no-brainer, merging it.

PR #878, opened 2015-03-04
DS-2485: DefaultItemVersionProvider hides an IdentifierException

pnbecker:


pnbecker:
As the patch doesn't change any behavior and the code change is really minor, I merge it directly.

PR #880, opened 2015-03-05
DS-2491 set deletedRecord=transient in OAI Identify

helix84:
https://jira.duraspace.org/browse/DS-2491


aschweer:
As said on Jira, +1

PR #881, opened 2015-03-05
DS-3109: Introduce a VersionedHandleIdentifierProvider that creates new handles for each version

pnbecker:
This PR originally was written against ~~https://jira.duraspace.org/browse/DS-1348~~. DS-1348 was replaced by: https://jira.duraspace.org/browse/DS-3109
This PR also fixes https://jira.duraspace.org/browse/DS-2940

The VersionedHandleIdentifierProvider introduced the idea of an handle
that always refers the newest version (in the sense of an edition) of an
Item. This breaks persistence of handles. This commit changes the
VersionedHandleIdentifierProvider to still create handles containing the
version number, but it does not change the bare handle to always point
on the newest version. This PR does not include the idea of another
handle that always points to the newest version (see the ticket for
further explaination).

### DIFF
--- a/dspace-api/src/main/java/org/dspace/identifier/HandleIdentifierProvider.java
+++ b/dspace-api/src/main/java/org/dspace/identifier/HandleIdentifierProvider.java
@@ -37,8 +37,6 @@ public class HandleIdentifierProvider extends IdentifierProvider {
     /** Prefix registered to no one */
     protected static final String EXAMPLE_PREFIX = "123456789";
 
-    protected String[] supportedPrefixes = new String[]{"info:hdl", "hdl", "http://"};
-
     @Autowired(required = true)
     protected HandleService handleService;
     @Autowired(required = true)
@@ -52,23 +50,21 @@ public class HandleIdentifierProvider extends IdentifierProvider {
     @Override
     public boolean supports(String identifier)
     {
-        for(String prefix : supportedPrefixes){
-            if(identifier.startsWith(prefix))
-            {
-                return true;
-            }
+        String prefix = handleService.getPrefix();
+        String handleResolver = ConfigurationManager.getProperty("handle.canonical.prefix");
+        if (identifier == null)
+        {
+            return false;
         }
-
-        try {
-            String outOfUrl = retrieveHandleOutOfUrl(identifier);
-            if(outOfUrl != null)
-            {
-                return true;
-            }
-        } catch (SQLException e) {
-            log.error(e.getMessage(), e);
+        if (identifier.startsWith(prefix)
+                || identifier.startsWith(handleResolver)
+                || identifier.startsWith("http://hdl.handle.net/")
+                || identifier.startsWith("hdl:")
+                || identifier.startsWith("info:hdl"))
+        {
+            return true;
         }
-
+        
         return false;
     }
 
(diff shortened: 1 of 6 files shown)


pnbecker:
I rebased this PR today. I still have to test this PR to work properly before it can be reviewed.

pnbecker:
This PR can go into review. Testing it is simple: Enable the VersionedHandleIdentifierProvider in [dspace]/config/spring/api/identifier.xml, enable Versioning, create an item and several versions, take a look on the created Handles. They should look like e.g.: 123456789/100, 123456789/100.2, 123456789/100.3 and so on.

This PR must be merged AFTER #900. If the order is changed we get an merge conflict (which shouldn't be hard to resolve). Nevertheless this PR does not directly depend on #900.

pnbecker:
I added a fix against DS-2940. DS-2940 describes two problems with canonical handles and metadata. As the VersionedHandleIdentifierProvider that uses canonical handles is renamed to VersionedHandleIdentifierProviderWithCanonicalHandles within this PR, I added my solution for DS-2940 here to avoid a merge conflict.

pnbecker:
I rebased this today and removed the "needs discussion" label, as all concerns were solved (see the jira issue for details: https://jira.duraspace.org/browse/DS-1348?focusedCommentId=48680&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-48680).

pnbecker:
Rebased this again and fixed again a new bug in the VersionedHandleIdentifierProviderWithCanonicalHandles.

pnbecker:
@benbosman @KevinVdV I tested this. Nevertheless perhaps you take another look especially on the VersionedHandleIdentifierProviderWithCanonicalHandles? As mentioned, rebasing made it necessary to bugfix this again.

KevinVdV:
Tested again with the default setup, but I noticed that the dc.identifier.uri now only gets the handle (123456789/6) as the metadata value and not an actual url to handle (http://hdl.handle.net/123456789/6).

hardyoyo:
I've run a test of this in Vagrant-DSpace, I created a new top level community and a collection within it, then imported 100 metadata-only items to that new collection. I can see the same behavior that KevinVdV mentioned 6 days ago, the dc.identifier.uri gets a handle that looks like 10673/1110 instead of http://hdl.handle.net/10673/1110, in addition, on the simple item view page, the URI is incorrectly linked to http://dspace.vagrant.dev:8080/xmlui/handle/10673/10673/1110, which will not work (notice the double handle prefix). I suspect the incorrect URI is causing the Mirage template to render that link incorrectly.

pnbecker:
I'll take care by the end of this week.

pnbecker:
Rebased, fixed reported problem (damn one liner) and tested.

hardyoyo:
+1 I tested again in Vagrant-DSpace using the latest code, I can now confirm that handles for newly-created items are constructed correctly: http://hdl.handle.net/10673/1111

pnbecker:
@hardyoyo thanks for testing.

kshepherd:
+1 tested in vagrant-dspace, got good handle for new version, and original handle points to original item w/ notice re: new versions

KevinVdV:
+1, tested with the both the old configuration & the new. Since we are at three +1's I'm merging this one.

PR #883, opened 2015-03-06
[DS-2493] "View more" link is shown even when there aren't more items.

rivaldi8:
Fixes https://jira.duraspace.org/browse/DS-2493


helix84:
+1, tested with default values (20 for collection, 5 for homepage) on:
collection with > 20 items: View More displayed
collection with < 20 items: View More not displayed
homepage with > 5 items: View More displayed

PR #884, opened 2015-03-08
DS-2501 fix SQL in REST /items/find-by-metadata-field

helix84:
Fix tested on Postgres.

https://jira.duraspace.org/browse/DS-2501

### DIFF
--- a/dspace-rest/src/main/java/org/dspace/rest/ItemsResource.java
+++ b/dspace-rest/src/main/java/org/dspace/rest/ItemsResource.java
@@ -916,22 +916,22 @@ public class ItemsResource extends Resource
      * Find items by one metadada field.
      * 
      * @param metadataEntry
-     *            Metadata field by which will be searched.
+     *            Metadata field to search by.
      * @param scheme
      *            Scheme of metadata(key).
      * @param value
      *            Value of metadata field.
      * @param headers
-     *            If you want to access to item under logged user into context.
-     *            In headers must be set header "rest-dspace-token" with passed
-     *            token from login method.
-     * @return Return array of founded items.
+     *            If you want to access the item as the user logged into context,
+     *            header "rest-dspace-token" must be set to token value retrieved
+     *            from the login method.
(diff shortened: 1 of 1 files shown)


dspace-rest/src/main/java/org/dspace/rest/ItemsResource.java
```diff
@@ -992,7 +992,7 @@ else if (!AuthorizeManager.authorizeActionBoolean(context, bitstream, org.dspace
             {
                 sql += "QUALIFIER='" + metadata[2] + "' AND ";
             }
-            sql += "dbms_lob.substr(TEXT_VALUE, 40)='" + metadataEntry.getValue() + "' AND ";
+            sql += "TEXT_VALUE='" + metadataEntry.getValue() + "' AND ";
```
ctu-developers: I think that is still a problem with error ORA-00932 (direct compare CLOB and string - see above TODO comment) for oracle db.

helix84: It would probably be best to construct that part of SQL separately for Oracle and Postgres.

The solution for Oracle doesn't seem ideal, isn't there a way to do it without a constant? If this is the best we can do, I'll update the code here.

ctu-developers: I agree. I tried the oracle string conversion functions for VARCHAR2, but without success. Always ended with an error ORA-00932. I think that is no way to use identical code for both DBMS, except change datatype CLOB to VARCHAR2 of column TEXT_VALUE of table METADATAVALUE.

I think that better solution for oracle will be:
```sql
sql += "dbms_lob.compare(TEXT_VALUE, '" + metadataEntry.getValue() + "') = 0 AND ";
```
or maybe
```sql
sql += "dbms_lob.compare(TEXT_VALUE, to_clob('" + metadataEntry.getValue() + "')) = 0 AND ";
```

PR #886, opened 2015-03-11
DS-2020 null check & turning _ to / in handles

kosarko:


aschweer:
I haven't tested this, but +1 based on looking at the code.

aschweer:
https://jira.duraspace.org/browse/DS-2020

peterdietz:
Not tested, but code looks okay. +1

hardyoyo:
I'll note that Tim Donohue gave the original patch a +1 a while ago, so that's +3, I'm going to merge this to master and backport this to 5_x now.

PR #887, opened 2015-03-12
DS-2504: Method getLink() returns incorrect context path

ctu-developers:
This is only quick fix. The solution will be improved in a few days.

https://jira.duraspace.org/browse/DS-2504

### DIFF
--- a/dspace-rest/src/main/java/org/dspace/rest/Resource.java
+++ b/dspace-rest/src/main/java/org/dspace/rest/Resource.java
@@ -11,6 +11,7 @@ import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.List;
 
+import javax.servlet.ServletContext;
 import javax.servlet.http.HttpServletRequest;
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.HttpHeaders;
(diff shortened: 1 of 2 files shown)


mdiggory:
There is no rule that the rest api should be served at the URI "/rest". I would recommend using the servlet context, or externalizing a baseURL.

ctu-developers:
Yes, we are planning to do so. Maybe tomorrow;-)

ctu-developers:
I added static method to class Resource. I dont like that dependency between DSpaceObject and Resource, but i think there is no option. (i do not see any other options).

mwoodiupui:
The <extends Resource>, which knows its context path, could supply that path as a constructor argument to its <extends DSpaceObject>, or DSpaceObject could provide a setter to be used by an <extendsResource>.

Ideally DSpaceObject would just supply the part of the path it knows, and <extends Resource> would concatenate that with the part it knows, but I don't know how to do that with all the JAX-RS magic going on.

ctu-developers:
Good ideas, but if it will be setting by constructor or setter, then many code must be changed. I think this is only way how to do it. Because DSpace Object cannot know context path, because it is not Resource. So, it must get it from Resource.

peterdietz:
+1 I think this looks clean, and gets the job done.

helix84:
Does anyone object to merging this as-is?

PR #889, opened 2015-03-16
DS-2509 Update REST README.md with CRUD

ctu-developers:
I have added some information about REST api from version 5.0 to readme, which was outdated.

https://jira.duraspace.org/browse/DS-2509

### DIFF
--- a/dspace-rest/README.md
+++ b/dspace-rest/README.md
@@ -10,59 +10,174 @@ This REST API is integrated directly into the DSpace code-base.
   * ```<Context path="/rest"  docBase="/dspace/webapps/rest"  allowLinking="true"/>```
 
 
-At this point, this is a READ ONLY API for DSpace, for the anonymous user. Only Anonymous READ Communities, Collections, Items, and Bitstreams are available.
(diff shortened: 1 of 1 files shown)


helix84:
I put the changes into both master and dspace-5_x (for 5.2 release) with some changes:

* changed {rest-endpoint} back to http://localhost:8080/ which is the default, thus can be copy-pasted more easily
* added HTTP 415 status
* fixed some typos

863ff236243d06b08132b88e6a236515aae4e059
6f45627f109cd55badf567e36048c1e04c307f86

ctu-developers:
There is only our code without @helix84 patch https://github.com/DSpace/DSpace/commit/6f45627f109cd55badf567e36048c1e04c307f86 in the 

https://github.com/DSpace/DSpace/tree/master/dspace-rest
https://github.com/DSpace/DSpace/tree/dspace-5_x/dspace-rest

Merge was done correctly?

ctu-developers:
Why was replaced "{rest-endpoint}" with "http://localhost:8080"? Was merging done correctly?

helix84:
Yes, I explained the changes in comments above. You can see the commits on the branch here:

https://github.com/DSpace/DSpace/commits/master/dspace-rest

It seemed to me more understandable to put a concrete address there, especially because it was not explained what {rest-endpoint} stands for, but also for ease of copying/pasting (even though the address may change, this is the default).

PR #890, opened 2015-03-16
[DS-2513] Improve multipart header parsing

aschweer:
The original code broke when files were uploaded whose name contains
the semicolon-space sequence.

http://jira.duraspace.org/browse/DS-2513


helix84:
+1

I tested with the "; " sequence, some pangrams and 4-byte UTF-8 sequences. I tested on Chrome 37 and Iceweasel 24. I did no filename length tests.

PR #893, opened 2015-03-20
[DS-2212] Ignore _version_ field when sharding solr statistics

aschweer:


aschweer:
This should probably be tested before merging.

terrywbrady:
@aschweer , this worked for me.  I copied my production stats to my test server and counted the number of records in each shard before and after the sharding process.  I was able to successfully shard my 2014 statistics.

If you have a stackoverflow account, feel free to post your solution to http://stackoverflow.com/questions/26941260/normalizing-solr-records-for-sharding-version-issues.  This might be helpful to others in the future.

Thanks you for finding this solution.

aschweer:
Thanks for testing and giving feedback, @terrywbrady . It looks like this will be included in DSpace 5.2. I've added an answer to your SO question.

PR #899, opened 2015-03-30
DS-2529 CSV import bugfix for fields under authority control with a language

antoine-atmire:
https://jira.duraspace.org/browse/DS-2529

### DIFF
--- a/dspace-api/src/main/java/org/dspace/app/bulkedit/MetadataImport.java
+++ b/dspace-api/src/main/java/org/dspace/app/bulkedit/MetadataImport.java
@@ -1169,10 +1169,8 @@ public class MetadataImport
      */
     private static boolean isAuthorityControlledField(String md)
     {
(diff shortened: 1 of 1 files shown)


dspace-api/src/main/java/org/dspace/app/bulkedit/MetadataImport.java
```diff
@@ -1169,10 +1169,8 @@ private static int displayChanges(List<BulkEditChange> changes, boolean changed)
      */
     private static boolean isAuthorityControlledField(String md)
     {
-        int pos = md.indexOf("[");
-        String mdf = (pos > -1 ? md.substring(0, pos) : md);
-        pos = md.indexOf(":");
-        mdf = (pos > -1 ? md.substring(pos+1) : md);
```
bram-atmire: The old 1175 is where the bug resides: it should substring the mdf string and NOT the original md string.

PR #901, opened 2015-04-01
DS-2531 New entries for the robots hostname list

bram-atmire:
http://jira.duraspace.org/browse/DS-2531


aschweer:
Not sure how many +1s we need for this sort of thing -- you definitely have mine!

PR #903, opened 2015-04-01
DS-2508: Retrieve of collections by findAll() crash with SQL error

ctu-developers:
https://jira.duraspace.org/browse/DS-2508
Mergeable to _master_ and _dspace-5_x_ (backport for _dspace-4_x_ see #904)!


helix84:
+1, tested the collections endpoint on master with Postgres, doesn't break anything

PR #907, opened 2015-04-02
[DS-2534] Fix a bug in the "Jump to" browse feature

kutsurak:
When computing the offset for the "jump to" feature at
SolrBrowseDAO.doOffsetQuery we should take into account if we are
browsing a subset of the items (e.g. we are viewing the items that have
a specific subject) and not all of them (e.g. by title).


robintaylor:
I don't really know this area of the codebase, but I have applied the fix locally and our users tell me its now working as it should, so its +1 from me.

PR #914, opened 2015-04-14
DS-1965: Adds date fields to the form used to edit policies (JSPUI).

pnbecker:


aschweer:
Transferring my comment from https://github.com/DSpace/DSpace/pull/915

Tested locally, +1.

I would like to see some more error handling (eg putting "201" into the date field currently leads to the date being set to 0201-01-01), but XMLUI doesn't do error handling either, so that is a separate issue.

PR #920, opened 2015-04-16
Minor: fix year and capitalization in displayed copyright

cwilper:
I just noticed these were wrong in Mirage after installing 5.1.

(Please let me know if I should be creating JIRA issues for stuff like this an I'll be glad to.)


peterdietz:
This looks good to me. Its so minor, I don't think its necessary to fire a jira.

PR #925, opened 2015-04-20
DS-2550: fix ImageMagick/Ghostscript problems with transparent pdfs

pnbecker:
http://jira.duraspace.org/browse/DS-2550


pnbecker:
#926 is a cherry-pick of this commit against dspace-5_x branch.

PR #926, opened 2015-04-21
DS-2550: fix ImageMagick/Ghostscript problems with transparent pdfs

pnbecker:
PR against dspace-5_x branch.


pnbecker:
this is a cherry-pick on dspace-5_x of #925

PR #927, opened 2015-04-21
DS-2551: JSPUI show thumbnails only if user has read permission

pnbecker:


pnbecker:
#928 is a cherry-pick of this commit on dspace-5_x.

PR #708, opened 2014-10-25
DS-2218: Unable to use command "update-handle-prefix"

ctu-developers:
https://jira.duraspace.org/browse/DS-2218

Use the command "bin/dspace update-handle-prefix <old handle> <new handle>" ends with the "ORA-00911: invalid character". The problem is the extra semicolon in sql command.

I made changes in formatting SQL.

PR #929, opened 2015-04-21
DS-2218: Unable to use command "update-handle-prefix"

ctu-developers:
https://jira.duraspace.org/browse/DS-2218

Mergeable to _master_ and _dspace-5_x_ (backport for _dspace-4_x_ see #708)!


helix84:
+1, tested on master with Postgres

PR #930, opened 2015-04-22
DS-2218: Unable to use command "update-handle-prefix"

helix84:
Replaces #708 which had wrong target branch.


helix84:
+1, tested on postgres.

PR #936, opened 2015-04-29
DS-1493 : Extends PR#190, makes Social Export module work with DSpace5

olli-gold:
I have taken João's work and prepared it to work with DSpace5. I have successfully tested it locally.

It would be great to see it getting merged into master.

Surely it's not yet perfect, and I am willing to improve it if necessary. Tim Donohue's comments on https://jira.duraspace.org/browse/DS-1493 have not been integrated yet. I have written another comment there.


olli-gold:
I'm closing this PR before merge, because I have just upgraded to DSpace 5.3 and was not able to port this PR to the new version. The extension is to heavy dependent on XOAI 2.2.9, which is not in use any more in newer DSpace versions. I guess I won't be able to support this feature, and so I'm canceling this PR. Sorry for any inconvenience.

PR #937, opened 2015-04-30
DS-2545: Show collections the user can submits items to only.

pnbecker:
The JSPSelectColletionStep and JSPStartSubmissionLookupStep already set
the collections -the user can submit to- as request attribute. The JSPs
us these argument already. This commit lets the SelectCollectionTag use
this attribute too instead of looking for the collections on itself.


pnbecker:
#938 is a cherry-pick of this PR against dspace-5_x branch.

hardyoyo:
tested in Vagrant-DSpace, works as expected: I created a new collection with only my own login as an allowed submitter, then I logged in as dspacedemo+submit@gmail.com, and started a new submission. My new collection was not listed as an option. Logged out, logged in as my own user, and started a submission. That user *can* submit to the new collection. +1

PR #938, opened 2015-04-30
DS-2545: Show collections the user can submit items to only.

pnbecker:
The JSPSelectColletionStep and JSPStartSubmissionLookupStep already set
the collections -the user can submit to- as request attribute. The JSPs
us these argument already. This commit lets the SelectCollectionTag use
this attribute too instead of looking for the collections on itself.


pnbecker:
Cherry-pick of #937 for dspace-5_x branch.

hardyoyo:
We don't really need a cherry-pick PR, if the original PR (see #937) is accepted, we just cherry-pick and push to master. I recommend this PR be closed.

PR #939, opened 2015-04-30
DS-2449 mirage2 template label

aschweer:
https://jira.duraspace.org/browse/DS-2449


aschweer:
https://jira.duraspace.org/browse/DS-2449

PR #941, opened 2015-05-05
[DS-2527] Disable collection authorisation enumeration optimisation by default

rradillen:
Disable collection authorisation enumeration optimisation by default so LDAP and Shibboleth may work out of the box.

### DIFF
--- a/dspace-api/src/main/java/org/dspace/content/Collection.java
+++ b/dspace-api/src/main/java/org/dspace/content/Collection.java
@@ -1564,7 +1564,7 @@ public class Collection extends DSpaceObject
 
     public static Collection[] findAuthorizedOptimized(Context context, int actionID) throws java.sql.SQLException
     {
-        if(! ConfigurationManager.getBooleanProperty("org.dspace.content.Collection.findAuthorizedPerformanceOptimize", true)) {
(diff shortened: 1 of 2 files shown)


peterdietz:
Please use the existing findAuthorizedPerformanceOptimize config
https://github.com/rradillen/DSpace/blob/patch-9/dspace/config/dspace.cfg#L843-L846

rradillen:
I moved the default to the Collection following dspace's pattern of not specified is default.

kshepherd:
+1, tested on DSpace-Vagrant. Didn't test LDAP or Shib auth, but can verify from DEBUG logs that the findAuthorizedPerformanceOptimize SQL isn't running by default.

(small nitpick that is probably out of scope for this PR, I like to keep the commented "example/default" config line the same as the *default* behaviour, not the opposite. eg. mail.server.disabled. Harder to accidentally reverse behaviour, user has to deliberately change "the default". I thought this was a convention from the old days but I might not be remembering correctly. Anyway, this PR is just following the convention the previous config lines had of including "example/default" line that is opposite of default behaviour so as I said, probably out of scope for this PR, just my 2c and I might think about a tidyup PR for all those boolean defaults if others agree `</nitpick>`)

PR #944, opened 2015-05-13
DS-2461 Escape some colons in queries

aschweer:
This allows searching for titles with colons while still allowing fielded searches


aschweer:
https://jira.duraspace.org/browse/DS-2461

hardyoyo:
This is a tiny change, if you're running it in production, I don't think much review is required. We trust you.

aschweer:
Thanks @hardyoyo -- sorry, I should have mentioned this on here too not just on the ticket, but I think this can still be improved a bit more. Not sure it should be merged now. I'll see if I can make the changes that I think would be good.

aschweer:
I've changed the code to escape the colons in the back end. Tested locally. I'll push the button.

PR #945, opened 2015-05-14
DS-2575 Ensure pooled/owned workflow task are listed in fixed order

aschweer:


aschweer:
https://jira.duraspace.org/browse/DS-2575

PR #951, opened 2015-05-22
[DS-2590] Fix multiple issues with distributed archives.

mwoodiupui:
Include dspace/modules/*/src/main/webapps so build succeeds.
Avoid damaging a sample ZIP archive by munging "line endings".
Upgrade to maven-assembly-plugin 2.5.4 (which uncovered the line ending problem).

### DIFF
--- a/dspace/src/main/assembly/release.xml
+++ b/dspace/src/main/assembly/release.xml
@@ -46,8 +46,10 @@
          <!-- Ensure line endings in all files are Unix (LF) line endings-->
          <lineEnding>unix</lineEnding>
       </fileSet>
-      <!-- Ensure the presence of "empty" directories for build steps
-           that need them. -->
+      <!-- Ensure the presence of "empty" overlay directories for build
+           steps that need them.
+           By default our ".gitignore" files are excluded by
+           "useDefaultExcludes=true". -->
       <fileSet>
          <directory>..</directory>
          <useDefaultExcludes>false</useDefaultExcludes>
--- a/dspace/src/main/assembly/release.xml
+++ b/dspace/src/main/assembly/release.xml
@@ -46,6 +46,19 @@
          <!-- Ensure line endings in all files are Unix (LF) line endings-->
          <lineEnding>unix</lineEnding>
       </fileSet>
+      <!-- Ensure the presence of "empty" directories for build steps
+           that need them. -->
+      <fileSet>
+         <directory>..</directory>
+         <useDefaultExcludes>false</useDefaultExcludes>
+         <includes>
+            <include>dspace/modules/**/.gitignore</include>
+         </includes>
+         <excludes>
+            <exclude>**/target/**</exclude>
+         </excludes>
+         <lineEnding>unix</lineEnding>
+      </fileSet>
    </fileSets>
 
 </assembly>
--- a/dspace/src/main/assembly/src-release.xml
+++ b/dspace/src/main/assembly/src-release.xml
@@ -38,7 +38,8 @@
          <!-- Ensure line endings in all files are Unix (LF) line endings-->
          <lineEnding>unix</lineEnding>
       </fileSet>
-      <!-- Do not destroy ZIP archives by changing "line" endings! -->
+      <!-- Do not destroy ZIP archives by changing "line" endings!
+           (e.g. SWORD's example.zip) -->
       <fileSet>
          <directory>..</directory>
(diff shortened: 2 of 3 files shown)


mwoodiupui:
If accepted, this should be cherrypicked to master as well.

hardyoyo:
Tested, both zip files produced by these changed configurations build DSpace perfectly, +1, we can merge whenever Travis decides to get out of the way.

hardyoyo:
I know Travis is being pretty insistent about this build failure, but I cannot duplicate the error, tests pass on my machine. So, I remain +1 on this.

tdonohue:
Travis seems to be implying an issue with initializing the Unit Test framework. I haven't had a chance to try this myself yet, but looks to be running into a recursion bug...

Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.985 sec <<< FAILURE! - in org.dspace.util.MultiFormatDateParserTest
org.dspace.util.MultiFormatDateParserTest  Time elapsed: 0.985 sec  <<< ERROR!
java.lang.IllegalArgumentException: ConfigurationManager: Too many levels of recursion in configuration property variable interpolation, property=dspace.name
	at org.dspace.core.ConfigurationManager.interpolate(ConfigurationManager.java:837)
	at org.dspace.core.ConfigurationManager.interpolate(ConfigurationManager.java:863)
	at org.dspace.core.ConfigurationManager.interpolate(ConfigurationManager.java:863)
	at org.dspace.core.ConfigurationManager.interpolate(ConfigurationManager.java:863)
	at org.dspace.core.ConfigurationManager.interpolate(ConfigurationManager.java:863)
	at org.dspace.core.ConfigurationManager.interpolate(ConfigurationManager.java:863)
	at org.dspace.core.ConfigurationManager.interpolate(ConfigurationManager.java:863)
	at org.dspace.core.ConfigurationManager.interpolate(ConfigurationManager.java:863)
	at org.dspace.core.ConfigurationManager.interpolate(ConfigurationManager.java:863)
	at org.dspace.core.ConfigurationManager.interpolate(ConfigurationManager.java:863)
	at org.dspace.core.ConfigurationManager.loadConfig(ConfigurationManager.java:668)
	at org.dspace.AbstractUnitTest.initOnce(AbstractUnitTest.java:99)

pom.xml
```diff
@@ -124,7 +124,7 @@
              </plugin>
              <plugin>
                 <artifactId>maven-assembly-plugin</artifactId>
-                <version>2.4.1</version>
+                <version>2.5.4</version>
```
hardyoyo: From my testing, this change here is the likely cause of the test failures on Travis. Since this change is not necessary to the bug fix, let's leave the maven-assembly-plugin version number at 2.4.1, and leave the wondering what went wrong for another day?

dspace/src/main/assembly/release.xml
```diff
@@ -46,6 +46,19 @@
          <!-- Ensure line endings in all files are Unix (LF) line endings-->
          <lineEnding>unix</lineEnding>
       </fileSet>
+      <!-- Ensure the presence of "empty" directories for build steps
+           that need them. -->
```
tdonohue: I'd like this comment to be enhanced slightly (as I'm not sure it's descriptive enough of the problem):

dspace/src/main/assembly/src-release.xml
```diff
@@ -33,10 +33,35 @@
          <excludes>
             <exclude>**/target/**</exclude>
             <exclude>.*</exclude>
+            <exclude>**/*.zip</exclude>
          </excludes>
          <!-- Ensure line endings in all files are Unix (LF) line endings-->
          <lineEnding>unix</lineEnding>
       </fileSet>
+      <!-- Do not destroy ZIP archives by changing "line" endings! -->
```
tdonohue: This could use more explanation as it's a bit of a misleading comment. Perhaps:

Do not destroy source code ZIP archives by changing "line" endings! (e.g. SWORD's example.zip)

tdonohue:
Besides the comments I've added above (which just ask for a bit more documentation in the form of XML comments), I'm +1 these changes. They should be merged to all branches which had PR #874 applied. Those branches include:
* master
* dspace-5_x
* dspace-4_x
* dspace-3_x

mwoodiupui:
My initial guess is that m-assembly-p 2.5 is not filtering things that 2.4 did.

PR #952, opened 2015-05-27
DS-2594 Long file names overlap the second column of item metadata in…

rivaldi8:
… Mirage 2

Fixes https://jira.duraspace.org/browse/DS-2594


helix84:
+1

PR #953, opened 2015-05-27
[DS-2502] Correct some dependencies to prevent pulling in servlet-api

mwoodiupui:
This needs some testing.  It leaves servlet-api.jar out of all webapp.s but includes it in [DSpace]/lib.

### DIFF
--- a/dspace-api/pom.xml
+++ b/dspace-api/pom.xml
@@ -604,12 +604,20 @@
             <artifactId>google-api-services-analytics</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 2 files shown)


peterdietz:
+1 I was testing a jspui instance on tomcat8 and it wouldn't appear to start. I cherry-picked this patch, and JSPUI is working fine on tomcat8 now.

hardyoyo:
+1, I've been running this in production for months, I appreciate not hearing about servlet-api every time I watch the log files.

PR #916, opened 2015-04-14
DS-2539: REST: Export Metadata Schema and Field Registries

terrywbrady:
https://jira.duraspace.org/browse/DS-2539

Export the DSpace Metadata Schema and Field Registries.

This exported file can be used in conjunction with the FileAnalyzer to validate metadata field while generating DSpace Ingest Folders.

https://github.com/Georgetown-University-Libraries/File-Analyzer/wiki/Create-Ingest-Folders-for-a-Set-of-Files

The FileAnalyzer is a desktop application containing a number of automation tasks for library and digitization projects. A set of DSpace related tasks are included in the code base.

PR #954, opened 2015-05-27
DS-2583 Additional REST Verbs to Support Reporting Tools

terrywbrady:
# Notes about this PR
- See https://jira.duraspace.org/browse/DS-2583
- This code was built for DSpace 4.3.  Submitting 4.3 compatible code which will fail in the build
- Hardy Pottinger suggested submitting as is.
- This code contains code also present in https://github.com/DSpace/DSpace/pull/916
- **Question**: Should the sample client code be included in this PR?
- **Question**: If client code is included, should it be served up from /xmlui/... or from /rest/...?
- **Question**: Can Jersey be configured to return html/css/js files?
# Proposed REST API based reporting tool for DSpace

A proposed reporting tool for DSpace based on the DSpace 4.x REST api.
## Demonstration Video

[Demonstration Video of Possible Report Tools](http://www.youtube.com/watch?v=1Lxzqpq_3x4)
## Collections Filters

_ItemsFilterDefs.java_ contains a set of filter definitions that match specific item use cases.

```
public enum ItemFilterDefs implements ItemFilterTest {
    is_not_withdrawn("Withdrawn Items") {
        public boolean testItem(Context context, Item item) {
            return !item.isWithdrawn();
        }        
    },
    is_withdrawn("Available Items - Not Withdrawn") {
        public boolean testItem(Context context, Item item) {
            return item.isWithdrawn();
        }        
    },
    is_discoverable("Discoverable Items - Not Private") {
        public boolean testItem(Context context, Item item) {
            return item.isDiscoverable();
        }        
    },
    is_not_discoverable("Private Items - Not Discoverable") {
        public boolean testItem(Context context, Item item) {
            return !item.isDiscoverable();
        }        
    },
    has_multiple_originals("Item has Multiple Originals") {
        public boolean testItem(Context context, Item item) {
            int count = 0;
            try {
                for(Bundle bundle: item.getBundles("ORIGINAL")){
                    count += bundle.getBitstreams().length;
                }
            } catch (SQLException e) {
            }
            return count > 1;
        }        
    },
    ...
```

This code illustrates a mechanism to filter DSpace items by use case and item properties.
### Sample Collections Filter Client

_The purpose of this client code is to illustrate how quickly a useful client application could be built by extending the REST API to support the ability to filter collections._
- [Sample Filter Client Page](https://github.com/Georgetown-University-Libraries/DSpaceRestQCReports/tree/master/sampleClient/index.html)
- [Sample Filter Client JS](https://github.com/Georgetown-University-Libraries/DSpaceRestQCReports/tree/master/sampleClient/restCommon.js)
- [Sample Client Screen Shots](https://github.com/Georgetown-University-Libraries/DSpaceRestQCReports/releases/tag/v0.1)
## Metadata Query Tool

This tool allows a user to query for items based on the presence of one or more metadata fields within an item.  The tool supports 8 operators: exists, doesn't exist, equals, not equals, like, not like, matches, doesn't match.  More than one operation can be provided.
### Sample Metadata Query Tool Client

_The purpose of this client code is to illustrate how quickly a useful client application could be built by extending the REST API to support the ability to query item metadata._
- [Sample Query Tool Client Page](https://github.com/Georgetown-University-Libraries/DSpaceRestQCReports/tree/master/sampleClient/query.html)
- [Sample Query Tool Client JS](https://github.com/Georgetown-University-Libraries/DSpaceRestQCReports/tree/master/sampleClient/restQuery.js)
- [Sample Query Tool Client Screen Shots](https://github.com/Georgetown-University-Libraries/DSpaceRestQCReports/releases/tag/v0.2)
## New API Calls Needed to Support These Tools
- **/metadataregistry** - Return the Metadata Registry
- **/filters** - Return a set of pre-defined filters for items of interest
- **/filtered-collections** - Return items / item counts within a collection that satisfy a filter
- **/filtered-items** - Return items that match a metadata query

### DIFF
--- a/dspace-rest/src/main/java/org/dspace/rest/FilteredCollectionsResource.java
+++ b/dspace-rest/src/main/java/org/dspace/rest/FilteredCollectionsResource.java
@@ -1,0 +1,115 @@
+/**
+ * The contents of this file are subject to the license and copyright
+ * detailed in the LICENSE and NOTICE files at the root of the source
+ * tree and available online at
+ *
+ * http://www.dspace.org/license/
+ */
+package org.dspace.rest;
+
+
+import org.apache.log4j.Logger;
+import org.dspace.authorize.AuthorizeManager;
+import org.dspace.content.DSpaceObject;
+import org.dspace.core.ConfigurationManager;
+import org.dspace.core.Constants;
+import org.dspace.rest.common.Collection;
+import org.dspace.usage.UsageEvent;
+import org.dspace.utils.DSpace;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.ServletContext;
+import javax.ws.rs.*;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.HttpHeaders;
+import java.sql.SQLException;
+import java.util.ArrayList;
+
+/*
+The "Path" annotation indicates the URI this class will be available at relative to your base URL.  For
+example, if this web-app is launched at localhost using a context of "hello" and no URL pattern is defined
+in the web.xml servlet mapping section, then the web service will be available at:
+
+http://localhost:8080/<webapp>/collections
+ */
+@Path("/filtered-collections")
+public class FilteredCollectionsResource {
+    private static Logger log = Logger.getLogger(FilteredCollectionsResource.class);
+
+    @javax.ws.rs.core.Context ServletContext servletContext;
+    
+    private static final boolean writeStatistics;
+	
+	static{
+		writeStatistics=ConfigurationManager.getBooleanProperty("rest","stats",false);
+	}
+
+    @GET
+    @Path("/{collection_id}")
+    @Produces({MediaType.APPLICATION_JSON, MediaType.APPLICATION_XML})
+    public org.dspace.rest.common.FilteredCollection getCollection(@PathParam("collection_id") Integer collection_id, @QueryParam("expand") String expand, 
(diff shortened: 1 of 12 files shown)


helix84:
WRT your questions: I don't think client code belongs to the DSpace codebase. I suggest you post it as a separate project to https://github.com/DSpace-Labs

There are two proposed ideas for serving HTML for the REST API:
1) serve a client-side XSLT stylesheet for the XML response (just like we do with OAI since DSpace 3)
2) put up a Swagger description of the API and allow a petshop-style client to be autogenerated

terrywbrady:
@helix84 , thank you for the suggestions.  Regarding suggestion #1, I presume that an institution that wanted to use that approach would need to modify jersey code within the rest module to embed the client stylesheet onto the return XML or they would need to implement a custom verb that would return the stylesheet + XML.

terrywbrady:
@helix84 , I am trying to implement your recommendation to return xml with an xml-stylesheet pi from the rest service.  While looking for an example of this, I see some dialog on StackOverflow that uses the XmlHeader annotation: http://stackoverflow.com/questions/14739303/jax-rs-jersey-text-xml-can-i-specify-a-xml-stylesheet/14739593#14739593 .  This does not appear to compile in DSpace 5x.  Do you happen to know a mechanism that can do this?

terrywbrady:
I found https://jersey.java.net/documentation/latest/media.html#d0e8844 which answered this question.  My plan is to allow report resources to be defined in a config file.

terrywbrady:
Closing this PR.  A fresh version of the PR: https://github.com/DSpace/DSpace/pull/1062

helix84:
Sorry, I missed your questions. Thanks for the updates.

PR #955, opened 2015-05-29
DS-2560 - Did you mean option missing in Mirage2

Mini-Pillai:


hardyoyo:
demonstrated as working during live demo at OR2015, ~50 peer reviewers, including 4 commiters, gave their plus one, merging

PR #956, opened 2015-06-02
DS-2598 Correct XPATH for available date in mets format xoai

aschweer:
This ensures that dc.date.available is shown when using the mets metadata
format in OAI-PMH. Previously, the dateAvailable element was present but empty.


aschweer:
https://jira.duraspace.org/browse/DS-2598

helix84:
+1, tested

PR #964, opened 2015-06-16
DS-2618: Process mail.server.disabled in test-email

rradillen:
https://jira.duraspace.org/browse/DS-2618

### DIFF
--- a/dspace-api/src/main/java/org/dspace/core/Email.java
+++ b/dspace-api/src/main/java/org/dspace/core/Email.java
@@ -7,38 +7,21 @@
  */
 package org.dspace.core;
 
-import java.io.BufferedReader;
-import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.IOException;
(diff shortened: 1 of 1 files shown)


KevinVdV:
+1

dspace-api/src/main/java/org/dspace/core/Email.java
```diff
@@ -7,38 +7,21 @@
...
-import javax.mail.Transport;
+import javax.mail.*;
 import javax.mail.internet.InternetAddress;
 import javax.mail.internet.MimeBodyPart;
 import javax.mail.internet.MimeMessage;
 import javax.mail.internet.MimeMultipart;
-import org.apache.log4j.Logger;
-import org.dspace.services.EmailService;
-import org.dspace.utils.DSpace;
+import java.io.*;
```
helix84: I understand star imports may be a sort of a religious issue with percieved pros and cons on both sides, but this is a bug fix, so let's not touch these.

rradillen: right that only confuses the issue. Removing

helix84:
+1

PR #965, opened 2015-06-18
DS-2620 typo in the word cocoon

kosarko:


mwoodiupui:
+1 obviously correct

hardyoyo:
+1 as well, obviously correct, I will note that this PR is against the maintenance branch for DSpace 5, and not the master branch, it will therefore need to be cherry-picked to master.

bram-atmire:
cherry picked to master & merging here in the maintenance branch.

PR #966, opened 2015-06-18
Adding CSS classes identifying each field by its metadata type

nteike:
Adding CSS classes identifying each field by its metadata type. This allows unique layout changes for each metadata type in the detail view.


helix84:
Great idea.

nteike:
I made a mistake. I will make a new pull request.

PR #967, opened 2015-06-18
DS-2624: Adding CSS classes identifying each field by its metadata type.

nteike:
Adding CSS classes identifying each field by its metadata type. This allows unique layout changes for each metadata type in the detail view.

### DIFF
--- a/dspace-jspui/src/main/java/org/dspace/app/webui/jsptag/ItemTag.java
+++ b/dspace-jspui/src/main/java/org/dspace/app/webui/jsptag/ItemTag.java
@@ -471,7 +471,7 @@ public class ItemTag extends TagSupport
 		// We use underscore as separator and no wildcard qualifier
 		// because dots and asterisks are forbidden as CSS class names.
 		String metadataNameClass = "";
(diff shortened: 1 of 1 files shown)


dspace-jspui/src/main/java/org/dspace/app/webui/jsptag/ItemTag.java
```diff
@@ -467,7 +467,19 @@ else if (eq.length > 2)
             
             if (values.length > 0)
             {
-                out.print("<tr><td class=\"metadataFieldLabel\">");
+                // Create CSS class to identify fields by their metadata name.
+		// We use underscore as separator and no wildcard qualifier
+		// because dots and asterisks are forbidden as CSS class names.
+		String metadataNameClass = "";
+		if (qualifier == null || qualifier == Item.ANY
```
KevinVdV: qualifier == Item.ANY => Item.ANY.equals(qualifier)

pnbecker:
@helix84 commented on PR #966 that he likes the idea. I do as well!
It is possible to add multiple classes, so perhaps to stay backward compatible we could keep the old CSS class and add the new one.

PR #984, opened 2015-07-08
DS-2645: teach IPMatcher to deal with  IpAddrs like  0:0:0:0:0:0:0:1

akinom:
https://jira.duraspace.org/browse/DS-2645

### DIFF
--- a/dspace-api/src/main/java/org/dspace/authenticate/IPMatcher.java
+++ b/dspace-api/src/main/java/org/dspace/authenticate/IPMatcher.java
@@ -279,7 +279,11 @@ public class IPMatcher
             }
             catch (UnknownHostException e)
             {
-                throw new IPMatcherException("Malformed IPv6 address ",e);
+                throw new IPMatcherException("Malformed IPv6 address: " + ipIn,e);
+            }
+            catch (IllegalArgumentException ia)
+            {
+                throw new IPMatcherException("Very Malformed IPv6 address: " + ipIn , ia);
             }
 
         for (int i = 0; i < netmask.length; i++)


mwoodiupui:
+1 by inspection.

dspace-api/src/main/java/org/dspace/authenticate/IPMatcher.java
```diff
@@ -279,7 +279,11 @@ public boolean match(String ipIn) throws IPMatcherException
             }
             catch (UnknownHostException e)
             {
-                throw new IPMatcherException("Malformed IPv6 address ",e);
+                throw new IPMatcherException("Malformed IPv6 address: " + ipIn,e);
+            }
+            catch (IllegalArgumentException ia)
```
terrywbrady: IllegalArgumentException  does not seem to be generated from the code within the try block.  Should the call to ip4ToIp6() be included in the try block?

akinom: For some reason which I can not explain  things are working just fine for '0:0:0:0:0:0:0:1’  - that is
IPMatcher.match  log.debugs info about the misinformed ip address and returns false

I tested against /dspace-5_x

So i suppose disregard this now

Monika

—
Monika Mevenkamp
Digital Repository Infrastructure Developer
Princeton University
Phone: 609-258-4161
Skype: mo-meven

On May 16, 2016, at 5:28 PM, Terry Brady <notifications@github.com<mailto:notifications@github.com>> wrote:


In dspace-api/src/main/java/org/dspace/authenticate/IPMatcher.java<https://github.com/DSpace/DSpace/pull/984#discussion_r63429162>:

> @@ -279,7 +279,11 @@ public boolean match(String ipIn) throws IPMatcherException
>              }
>              catch (UnknownHostException e)
>              {
> -                throw new IPMatcherException("Malformed IPv6 address ",e);
> +                throw new IPMatcherException("Malformed IPv6 address: " + ipIn,e);
> +            }
> +            catch (IllegalArgumentException ia)


IllegalArgumentException does not seem to be generated from the code within the try block. Should the call to ip4ToIp6() be included in the try block?

PR #987, opened 2015-07-09
DS-2650: JSPUI Recent Submission header only when needed

akinom:
see https://jira.duraspace.org/browse/DS-2650


KevinVdV:
Obvious fix, test & it works, so +1. Merging now.

Thanks @akinom !

PR #1007, opened 2015-08-06
DS-2692 REST-API /handle not reflecting updates

kosarko:
https://jira.duraspace.org/browse/DS-2692

Extend Handle from Resource and use createContext similarly to other *Resource classes

as pointed out in email by @peterdietz 

> I see that there was a code refactoring during the development, which successfully hit the DSpaceObject endpoints ( /items, /collections, /communities), but missed the /handle endpoint). Unintentional caching, a bug.
> The old way we were obtaining a context in REST is this (visible in handle code). We cache a context, hold on to it, try to use it as a singleton-ish-thing. It was supposed to be a way to help performance at one point, your not going back to the DB to get a new context for each request, so you could scale higher. 
>         try {
>             if(context == null || !context.isValid() ) {
>                 context = new Context();
>                 //Failed SQL is ignored as a failed SQL statement, prevent: current transaction is aborted, commands ignored until end of transaction block
>                 context.getDBConnection().setAutoCommit(true);
>             }
> The problem is that you hold the same DB connection for a very long time, and similar to a transaction, that DB connection will use the state of the database at the time it got the connection / context. So, other edits to objects will be missed, its caches / hold stale connection, and just sucks. After several days of tomcat running, it could also just run into some weird issue.
> The new DB connection that we figured out (Hat tip to Richard Rodgers at OR2013 in PEI), is to properly get a new context for each request, and properly close your context after each request. Wrapping everything in a try, and finally helps ensure success.
>         try
>         {
>             context = createContext(getUser(headers));
>             ...
>         finally
>         {
>             processFinally(context);   ### essentially just context.abort() 
>         }

### DIFF
--- a/dspace-rest/src/main/java/org/dspace/rest/HandleResource.java
+++ b/dspace-rest/src/main/java/org/dspace/rest/HandleResource.java
@@ -10,14 +10,16 @@ package org.dspace.rest;
 import org.apache.log4j.Logger;
 import org.dspace.authorize.AuthorizeManager;
 import org.dspace.core.Constants;
-import org.dspace.core.Context;
 import org.dspace.handle.HandleManager;
 import org.dspace.rest.common.Collection;
 import org.dspace.rest.common.Community;
 import org.dspace.rest.common.DSpaceObject;
 import org.dspace.rest.common.Item;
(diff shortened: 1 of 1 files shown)


peterdietz:
+1 Looks good to me

PR #1015, opened 2015-08-13
DS-2702 set mail.extraproperties before obtaining session

rradillen:
https://jira.duraspace.org/browse/DS-2702


mwoodiupui:
+1, the old code is clearly out of order.

PR #1020, opened 2015-08-18
DS-2124 Move LNI to a separate GitHub project

ghost:
Deleting the LNI since it is not going to be refactored as part of the Services API project. A copy has already been saved as a separate Github project.

https://jira.duraspace.org/browse/DS-2124
https://github.com/DSpace/dspace-lni


hardyoyo:
+1 (looks good to me, Travis says it compiles, which is all the testing we require for a module deletion)

hardyoyo:
In case it helps, here's the "change of address" for the LNI module's new home: https://github.com/DSpace/dspace-lni

mwoodiupui:
+1 looks thorough.

tdonohue:
Looks good. Merging this into the Services API "feature branch". It'll be merged over to "master" after the entire Services API refactor is completed (in time for 6.0)

PR #1021, opened 2015-08-19
DSpace refactored service api

KevinVdV:
Pull request for the service based api, see https://wiki.duraspace.org/display/DSPACE/DSpace+Service+based+api & https://jira.duraspace.org/browse/DS-2701 for more information


dspace-api/src/main/java/org/dspace/app/util/AuthorizeUtil.java
```diff
@@ -410,15 +414,15 @@ else if (!AuthorizeManager.isAdmin(context))
...
+        List<Community> parentCommunities = collection.getCommunities();
         if (AuthorizeConfiguration
                 .canCommunityAdminManageCollectionAdminGroup()
-                && parentCommunities != null && parentCommunities.length > 0)
+                && parentCommunities != null && parentCommunities.size() > 0)
         {
-            AuthorizeManager.authorizeAction(context, collection
-                    .getCommunities()[0], Constants.ADMIN);
+            authorizeService.authorizeAction(context, collection
+                    .getCommunities().get(0), Constants.ADMIN);
```
peterdietz: Is this the proper Service way, or needs another pass?
collection.getCommunities().get(0)

I can't find collectionService.getCommunities(collection), do I need to create it? I see getParentObject

KevinVdV: @peterdietz This is the correct way, you can get the communities from from a collection by using collection.getCommunities(). This is a database mapping that is a simple getter. These can still be made from the original database object.

Hope this helps.

PR #1025, opened 2015-08-20
DS-2701 Sword v1 and v2

aschweer:
This is purely to get the two modules to compile. There are no unit tests for either module and I haven't done any local tests.

### DIFF
--- a/dspace-sword/src/main/java/org/dspace/sword/ItemDepositor.java
+++ b/dspace-sword/src/main/java/org/dspace/sword/ItemDepositor.java
@@ -8,21 +8,23 @@
 package org.dspace.sword;
 
 import org.apache.commons.lang.StringUtils;
+import org.dspace.authorize.AuthorizeException;
 import org.dspace.content.*;
 import org.dspace.content.factory.ContentServiceFactory;
 import org.dspace.content.service.BitstreamFormatService;
 import org.dspace.content.service.BitstreamService;
 import org.dspace.content.service.BundleService;
 import org.dspace.content.service.ItemService;
-import org.dspace.core.Context;
 import org.dspace.core.ConfigurationManager;
-import org.dspace.authorize.AuthorizeException;
+import org.dspace.core.Context;
 import org.purl.sword.base.Deposit;
-import org.purl.sword.base.SWORDErrorException;
 import org.purl.sword.base.ErrorCodes;
+import org.purl.sword.base.SWORDErrorException;
 
-import java.io.*;
+import java.io.FileInputStream;
+import java.io.IOException;
 import java.sql.SQLException;
+import java.util.Iterator;
 import java.util.List;
 
 public class ItemDepositor extends Depositor
@@ -178,9 +180,12 @@ public class ItemDepositor extends Depositor
 			// from them.  This will ensure that the bitstream is physically
 			// removed from the disk.
 			Bitstream bs = result.getBitstream();
-			List<BundleBitstream> b2bs = bs.getBundles();
-			for (BundleBitstream b2b : b2bs) {
-				Bundle bundle = b2b.getBundle();
+			Iterator<BundleBitstream> b2bs = bs.getBundles().iterator();
+			while (b2bs.hasNext())
+			{
+				BundleBitstream bundleBitstream = b2bs.next();
+				b2bs.remove();
+				Bundle bundle = bundleBitstream.getBundle();
 				bundleService.removeBitstream(sc.getContext(), bundle, bs);
 				bundleService.update(sc.getContext(), bundle);
 			}
--- a/dspace-sword/src/main/java/org/dspace/sword/ItemDepositor.java
+++ b/dspace-sword/src/main/java/org/dspace/sword/ItemDepositor.java
@@ -2,7 +2,7 @@
  * The contents of this file are subject to the license and copyright
  * detailed in the LICENSE and NOTICE files at the root of the source
  * tree and available online at
- *
+ * <p>
  * http://www.dspace.org/license/
  */
 package org.dspace.sword;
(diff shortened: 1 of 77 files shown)


dspace-sword/src/main/java/org/dspace/sword/ItemDepositor.java
```diff
@@ -168,16 +172,17 @@ public void undoDeposit(DepositResult result) throws DSpaceSWORDException
...
 			Bitstream bs = result.getBitstream();
-			Bundle[] bundles = bs.getBundles();
-			for (int i = 0; i < bundles.length; i++)
-			{
-				bundles[i].removeBitstream(bs);
-				bundles[i].update();
+			List<BundleBitstream> b2bs = bs.getBundles();
+			for (BundleBitstream b2b : b2bs) {
+				Bundle bundle = b2b.getBundle();
+				bundleService.removeBitstream(sc.getContext(), bundle, bs);
```
KevinVdV: I think that we will get an "ConcurrentModificationException" here. Because the "BundleBitstream" list is persisted, so when we delete an item from a list we are iterating it will alter the list.
The best way to fix this is by using an iterator:
Bitstream bs = result.getBitstream();
Iterator<BundleBitstream> b2bs = bs.getBundles().iterator();
while (b2bs.hasNext()) {
	BundleBitstream bundleBitstream = b2bs.next();
	b2bs.remove();
	Bundle bundle = bundleBitstream.getBundle();
	bundleService.removeBitstream(sc.getContext(), bundle, bs);
	bundleService.update(sc.getContext(), bundle);
}

dspace-swordv2/src/main/java/org/dspace/sword2/VersionManager.java
```diff
@@ -10,75 +10,90 @@
...
+			this.archiveBundle(context, item, source);
 		}
 
         // remove all the bitstreams from the bundle
-		for (Bitstream bitstream : source.getBitstreams())
+		List<BundleBitstream> bundleBitstreams = source.getBitstreams();
+		for (BundleBitstream bundleBitstream : bundleBitstreams)
 		{
-			source.removeBitstream(bitstream);
+			bundleService.removeBitstream(context, source, bundleBitstream.getBitstream());
```
KevinVdV: I think that we will get an "ConcurrentModificationException" here. Because the "BundleBitstream" list is persisted, so when we delete an item from a list we are iterating it will alter the list.
The best way to fix this is by using an iterator (see comment above for example)

dspace-swordv2/src/main/java/org/dspace/sword2/VersionManager.java
```diff
@@ -10,75 +10,90 @@
...
-		Bundle[] bundles = bitstream.getBundles();
-		for (Bundle bundle : bundles)
+		List<BundleBitstream> bundleBitstreams = bitstream.getBundles();
+		for (BundleBitstream bundleBitstream : bundleBitstreams)
 		{
-			if (exempt != null && bundle.getID() != exempt.getID())
+			if (exempt != null && bundleBitstream.getBundle().getID() != exempt.getID())
 			{
-				bundle.removeBitstream(bitstream);
+				bundleService.removeBitstream(context, bundleBitstream.getBundle(), bitstream);
```
KevinVdV: I think that we will get an "ConcurrentModificationException" here. Because the "BundleBitstream" list is persisted, so when we delete an item from a list we are iterating it will alter the list.
The best way to fix this is by using an iterator (see comment above for example)

dspace-swordv2/src/main/java/org/dspace/sword2/ContainerManagerDSpace.java
```diff
@@ -824,22 +781,22 @@ protected void doContainerDelete(SwordContext swordContext, Item item, AuthCrede
...
 			{
 				InProgressSubmission wfi = wft.getWorkflowItem(context, item);
-				wfi.deleteWrapper();
+				workflowItemService.deleteWrapper(context, wfi);
 			}
 
 			// then delete the item
-			Collection[] collections = item.getCollections();
+			List<Collection> collections = item.getCollections();
 			for (Collection collection : collections)
```
KevinVdV: This will probably give a concurrentModificationException, best to use itemService.delete(context, item);

dspace-swordv2/src/main/java/org/dspace/sword2/MediaResourceManagerDSpace.java
```diff
@@ -779,13 +732,14 @@ private void removeContent(SwordContext swordContext, Item item, AuthCredentials
...
 			// remove content only really means everything from the ORIGINAL bundle
 			VersionManager vm = new VersionManager();
-			Bundle[] originals = item.getBundles("ORIGINAL");
-			for (Bundle original : originals)
+			List<Bundle> bundles = item.getBundles();
+			for (Bundle bundle : bundles)
 			{
-				vm.removeBundle(item, original);
+                if (Constants.CONTENT_BUNDLE_NAME.equals(bundle.getName()))
+				vm.removeBundle(swordContext.getContext(), item, bundle);
```
KevinVdV: This will cause a concurrent Modification exception, best use an iterator.

aschweer:
Thanks a lot for the review @KevinVdV! I'll make the changes you suggest tomorrow (or on the weekend if tomorrow gets away from me).

dspace-swordv2/src/main/java/org/dspace/sword2/VersionManager.java
```diff
@@ -10,75 +10,95 @@
...
+	protected ItemService itemService = ContentServiceFactory.getInstance().getItemService();
+	protected BundleService bundleService = ContentServiceFactory.getInstance().getBundleService();
+	protected BitstreamService bitstreamService = ContentServiceFactory.getInstance().getBitstreamService();
+
+	public void removeBundle(Context context, Item item, String name)
 			throws SQLException, AuthorizeException, IOException
 	{
 		boolean keep = ConfigurationManager.getBooleanProperty("swordv2-server", "versions.keep");
-		Bundle[] bundles = item.getBundles(name);
+		List<Bundle> bundles = item.getBundles();
```
aschweer: @KevinVdV , is the list ok here (since the remove call doesn't involve the service)? Or is this another problem spot for concurrent modification exceptions?

KevinVdV: @aschweer It will also cause an issue here, I didn't yet notice this one apparently. The problem is that in the call below: "this.removeBundle(context, item, b, keep)" actually calls an "itemService.removeBundle(context, item, source)" method.
Which will result in an alteration of the "bundles" list which are now looping over. This will lead to the ConcurrentModificationException.

The problem isn't related to a service call or not, it occurs because hibernate returns persisted lists and deleting an object results in a list alteration.
We used to work with just value based arrays which didn't have this problem since the object would still exist, when we delete an object using persisted lists the object is destroyed when it is deleted (so it is really gone). Since the object is gone the list loop will crash since the list picks up on this change.

Hope this helps.

aschweer: It does, thanks @KevinVdV! I've amended my pull request.

aschweer:
I think this is as far as it can go without being able to run end-to-end tests. I've added a final commit to re-format everything according to the [official DSpace coding conventions](https://wiki.duraspace.org/display/DSPACE/Code+Contribution+Guidelines#CodeContributionGuidelines-CodingConventions) but I'm happy to take that commit back out if people think it makes review too hard. 

So from my point of view this can be merged -- @KevinVdV (or anyone else), is there anything more you'd like to see in this PR first?

dspace-swordv2/src/main/java/org/dspace/sword2/WorkflowTools.java
```diff
@@ -44,10 +60,16 @@ public boolean isItemInWorkflow(Context context, Item item)
     {
         try
         {
-            if(ConfigurationManager.getProperty("workflow","workflow.framework").equals("xmlworkflow")){
-                return XmlWorkflowItem.findByItem(context, item) != null;
-            }else{
-                return BasicWorkflowItem.findByItem(context, item) != null;
+            if (ConfigurationManager
```
KevinVdV: Perhaps replace this by: WorkflowServiceFactory.getInstance().getWorkflowService().findByItem(context, item) ? This will work for both workflows at once & leave out the need for the old if this then that.

dspace-swordv2/src/main/java/org/dspace/sword2/WorkflowTools.java
```diff
@@ -94,10 +116,15 @@ public InProgressSubmission getWorkflowItem(Context context, Item item)
     {
         try
         {
-            if(ConfigurationManager.getProperty("workflow","workflow.framework").equals("xmlworkflow")){
-                return XmlWorkflowItem.findByItem(context, item);
-            }else{
-                return BasicWorkflowItem.findByItem(context, item);
+            if (ConfigurationManager
```
KevinVdV: Perhaps replace this by: WorkflowServiceFactory.getInstance().getWorkflowService().findByItem(context, item) ? This will work for both workflows at once & leave out the need for the old if this then that.

dspace-swordv2/src/main/java/org/dspace/sword2/WorkflowTools.java
```diff
@@ -144,38 +171,36 @@ public void startWorkflow(Context context, Item item)
...
             // kick off the workflow
-            boolean notify = ConfigurationManager.getBooleanProperty("swordv2-server", "workflow.notify");
-            if (ConfigurationManager.getProperty("workflow", "workflow.framework").equals("xmlworkflow")) {
-                if (notify) {
-                    XmlWorkflowServiceImpl.start(context, wsi);
-                } else {
-                    XmlWorkflowServiceImpl.startWithoutNotify(context, wsi);
+            boolean notify = ConfigurationManager
+                    .getBooleanProperty("swordv2-server", "workflow.notify");
+            if (ConfigurationManager
```
KevinVdV: Another case where the if this workflow then that can be replaced by WorkflowServiceFactory.getInstance().getWorkflowService().start(), WorkflowServiceFactory.getInstance().getWorkflowService().startWithoutNotify()

dspace-swordv2/src/main/java/org/dspace/sword2/CollectionListManagerDSpace.java
```diff
@@ -67,134 +77,145 @@ public Feed listCollectionContents(IRI colIRI, AuthCredentials authCredentials,
...
+                    {
+                        collectionItems.add(item);
+                        break;
+                    }
+                }
+            }
+
+            // finally get the ones out of the workflow
+            List<BasicWorkflowItem> wfis = basicWorkflowItemService
+                    .findByOwner(sc.getContext(), person);
```
KevinVdV: This is a small mistake (which I forgot to document until now: https://wiki.duraspace.org/display/DSPACE/DSpace+Service+based+api%3A+API+Changelist#DSpaceServicebasedapi:APIChangelist-WorkflowItemrefactoredinto:Interface(oldusagemovedto:BasicWorkflowItem)). This should use the "WorkflowServiceFactory.getInstance().getWorkflowItemService().findBySubmitter() method. This will create an abstraction from the workflow & return WorkflowItems which also have a getCollection() & getItem() method.

KevinVdV:
Almost there, just a few more fixes on the workflow side & this is ready to merge !

aschweer:
Excellent, thanks a lot @KevinVdV, I hadn't spotted the workflow related methods. Makes perfect sense to have one that's smart enough to figure out which type of workflow is actually being used. I hope I can get to the changes this week, but it may have to be the weekend.

aschweer:
All done with the workflow I think -- there are now no references to basic/xml workflow in either of the sword modules (just references to the generic workflow), which is great.

Done from my point of view (still with the caveat of not having run any tests).

KevinVdV:
Took a quick look around & couldn't find any more blockers. Haven't tested it though, but I assume that testing will occur once all the modules compile.

PR #1036, opened 2015-08-28
DS-1924 Include current locale in calculating cache key

aschweer:
https://jira.duraspace.org/browse/DS-1924

This ensures that the value of the 'current locale' page meta element reflects locale switches made by the user.


helix84:
+1, tested

PR #1037, opened 2015-08-28
[DS-2729] Ensure SiteService extends DSpaceObjectService

KevinVdV:


dspace-api/src/main/java/org/dspace/content/SiteServiceImpl.java
```diff
@@ -50,4 +58,33 @@ public Site createSite(Context context) throws SQLException {
...
+    public void updateLastModified(Context context, Site dso) throws SQLException, AuthorizeException {
+        //Not used at the moment
+    }
+
+    @Override
+    public void update(Context context, Site dso) throws SQLException, AuthorizeException {
+        if(!authorizeService.isAdmin(context)){
+            throw new AuthorizeException();
+        }
+        siteDAO.save(context, dso);
```
pnbecker: We should trigger an event here. IIRC at least the RDFConsumer would consume the event.

pnbecker:
Beside the line comment above I think we should override the getName(dso) method to return ${dspace.name}. Everything else looks almost the same I was doing yesterday evening, but you were faster. Thanks for that PR!

I haven't tested this PR yet, but after resolving the two things I mentioned above this should be merged asap.

dspace-api/src/main/java/org/dspace/content/SiteServiceImpl.java
KevinVdV: I agree fixed it so that siteService update() fires modified events.

KevinVdV:
The 2 comments mentioned above should be fixed, ensured that site.getName() links through to siteService.getName();
https://github.com/DSpace/DSpace/pull/1037/files#diff-5a810dd2bf287a4ab237a7c8014771c4R45

This way the getName() method cannot differ between service & data object call.

pnbecker:
Without having tested the code (but reviewed the changes): +1

pnbecker:
We can think about removing the link from org.dspace.content.Site.getName() to org.dspace.content.SiteService.getName(Site) as I've moved this into DSpaceObject in #1039.

pnbecker:
I tested this PR in the meantime, DSpace runs fine with it.

PR #1040, opened 2015-08-28
DS-2701: cleanup or org.dspace.rdf

pnbecker:
I reviewed our code within org.dspace.rdf and all packages bellow. Here is the result. Our Linked (Open) Data support needs #1037 to run properly. With #1037 and this PR our LOD support works fine again.


KevinVdV:
Changes look good, merging.

PR #1042, opened 2015-08-31
fix typo in method name (forwarder -> forwarded)

helix84:
Minor change in method name and all its uses (1). Since we're already doing a large API break in DSpace 6, let's fix method names, too.


helix84:
Fixed in master (for 6.0): 3829e1f5eecf34555f8990371c41d2d515de55e7

PR #1046, opened 2015-09-07
DS-2740 upload step fails when Sherpa/Romeo doesn't hold any copyrigh…

helix84:
…t information

https://jira.duraspace.org/browse/DS-2740


peterdietz:
Hi @helix84, 

How about this small tweak:
https://github.com/LongsightGroup/DSpace/commit/505c37e786757b122b68165ebf46ae9a1053c688

Show the journal, and then don't show the publisher section if its not present. Not just skip this result entirely if its missing both.

helix84:
@peterdietz Yes, that's better. Closing this, opened #1098.

PR #1052, opened 2015-09-11
DS-2748: Reduce Cocoon 404 log noise

cwilper:
This change uses a less noisy approach to making Cocoon send 404s when needed. When applied, the large traces will stop appear whenever a non-existing page is requested of DSpace, and 404s will be consistently returned for subsequent requests.

https://jira.duraspace.org/browse/DS-2748

Rather than throwing a ResourceNotFoundException, this goes back to setting the http response status directly. This alone isn't enough, however, because Cocoon will by default cache the response and begin providing it to subsequent requests with a 200 status code. To avoid that, this change also explicitly disables caching of such responses by returning null from getKey().

Note: This PR is compatible with PR #1050, but both are based on master and touch similar parts of PageNotFoundTransformer, so some minor manual merging may be required.

### DIFF
--- a/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/general/PageNotFoundTransformer.java
+++ b/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/general/PageNotFoundTransformer.java
@@ -7,15 +7,14 @@
  */
 package org.dspace.app.xmlui.aspect.general;
 
+import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 import java.io.Serializable;
 import java.sql.SQLException;
 
 import org.apache.cocoon.ResourceNotFoundException;
 import org.apache.cocoon.caching.CacheableProcessingComponent;
-import org.apache.cocoon.environment.ObjectModelHelper;
-import org.apache.cocoon.environment.Request;
-import org.apache.cocoon.util.HashUtil;
+import org.apache.cocoon.environment.http.HttpEnvironment;
 import org.apache.excalibur.source.SourceValidity;
 import org.apache.excalibur.source.impl.validity.NOPValidity;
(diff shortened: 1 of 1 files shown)


aschweer:
I haven't had a chance to test this yet, but "this change also explicitly disables caching of such responses" sounds very interesting. I've long suspected that 404s consume a lot of heap space but never had the time to confirm / track this down properly. Not caching in these situations may help fix that problem if it does indeed exist.

cwilper:
I think there's a good chance that more caching is occurring in Cocoon than intended, in general, with the various our of box transformers. One thing I noticed a while ago, not directly related to this issue, is that a lot of them are by default returning the string "0" as the key when trying to signal to Cocoon that caching should not occur. That seemed odd to me because the javadocs clearly state that you should be returning null in such cases.

Regarding this particular issue, one thing I'm not fully certain about is whether my change to uniformly return nulls from getKey in PageNotFoundTransformer has the effect of disabling cache where it shouldn't be.

cwilper:
Upon further investigation, I'm withdrawing this PR because it actually has the effect of consuming *more* cocoon cache than the existing approach of throwing ResourceNotFoundException.

I determined this by setting a breakpoint in org.apache.cocoon.caching.impl.CacheImpl#store, and examining the state of the cache after starting DSpace with a clear Cocoon cache, and hitting a missing page. The approach of throwing ResourceNotFoundException caused zero calls to add items to the cache. But the approach in this PR caused several, and I verified they were left in the cache upon a subsequent request by looking at CacheImpl.store.cache.memoryStore.map. These appeared to be cached components from earlier parts of the pipeline, but were cached as unique items as their keys contained the (bad) request path.

So, some useful information was gathered along the way. 1) Verified that the ResourceNotFoundException approach does not cause Cocoon cache to grow, and 2) Determined how to monitor the size of the cache in a debugger.

PR #1065, opened 2015-09-18
[DS-2701-rest-api-fixes] Small bugfixes for rest api

KevinVdV:
Small bugfixes concerning comparing of objects, missing update call & ConcurrentModificationException


KevinVdV:
Going to merge this directly because these are small issues.

peterdietz:
Hi Kevin, 

Thanks for catching these. I wasn't thinking of UUID's as objects, but as primitives, "==" will be wrong always, so equals is better. 

Also, thanks for catching the AuthorizeException, I didn't commit it, because I thought it was some debugging code of mine. (I had to catch runtime exceptions to catch the findByIdOrLegacy issue)

PR #1069, opened 2015-09-22
Update xoai.xml

josekarvalho:
Corrected the filter for Driver context and the prefix for OpenAIRE context

### DIFF
--- a/dspace/config/crosswalks/oai/xoai.xml
+++ b/dspace/config/crosswalks/oai/xoai.xml
@@ -373,7 +373,7 @@
             <Configuration>
                 <string name="field">dc.rights</string>
                 <string name="operator">contains</string>
-                <list>
+                <list name="values">
                     <string>open access</string>
                     <string>openAccess</string>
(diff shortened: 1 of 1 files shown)


helix84:
+1, obvious configuration fixes

kosarko:
What makes the prefix wrong, ie why can't it stay EC/FP. Why include only EC funding and not all the funding streams mentioned at http://api.openaire.eu/#cha_projects_http ?

josekarvalho:
We need to change some configurations there based on the evolution of the OpenAIRE guidelines from version 2 to 3 (see: https://guidelines.openaire.eu/en/latest/literature/introduction.html) and the definition of the content of the "openaire" set - https://guidelines.openaire.eu/en/latest/literature/use_of_oai_pmh.html - (and not the "openaire context" of the OAI of DSpace).

We will suggest some new changes to allow DSpace to be compatible with OpenAIRE 3.0 guidelines and also be 2.0 compatible.

PR #1074, opened 2015-09-23
[DS-2792] Add dc.creator to the output of qdc.xsl

bdewar:
This will improve the metadata that is harvested via OAI when the selected metadataFormat is qdc.  Previously, there was no attempt to include a dc:creator element.

### DIFF
--- a/dspace/config/crosswalks/oai/metadataFormats/qdc.xsl
+++ b/dspace/config/crosswalks/oai/metadataFormats/qdc.xsl
@@ -26,6 +26,11 @@
 			<xsl:for-each select="doc:metadata/doc:element[@name='dc']/doc:element[@name='title']/doc:element/doc:field[@name='value']">
 				<dc:title><xsl:value-of select="." /></dc:title>
 			</xsl:for-each>
+			<xsl:for-each select="doc:metadata/doc:element[@name='dc']/doc:element[@name='creator']/doc:element/doc:field[@name='value']">
+				<dc:creator>
+					<xsl:value-of select="." />
+				</dc:creator>
+			</xsl:for-each>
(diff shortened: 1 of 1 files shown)


dspace/config/crosswalks/oai/metadataFormats/qdc.xsl
```diff
@@ -26,6 +26,16 @@
 			<xsl:for-each select="doc:metadata/doc:element[@name='dc']/doc:element[@name='title']/doc:element/doc:field[@name='value']">
 				<dc:title><xsl:value-of select="." /></dc:title>
 			</xsl:for-each>
+			<xsl:for-each select="doc:metadata/doc:element[@name='dc']/doc:element[@name='creator']/doc:element/doc:field[@name='value']">
+				<dc:creator>
+					<xsl:value-of select="." />
+				</dc:creator>
+			</xsl:for-each>
+			<xsl:for-each select="doc:metadata/doc:element[@name='dc']/doc:element[@name='contributor']/doc:element[@name='author']/doc:element/doc:field[@name='value']">
```
hardyoyo: This change does more than add dc.creator, it also adds dc.contributor.author... which is probably a necessary change/bugfix, but has not yet been mentioned as an issue.

hardyoyo:
+1 to merge, it would be good if the QDC view of OAI mentioned authors. :-)

dspace/config/crosswalks/oai/metadataFormats/qdc.xsl
bdewar: Hi.

I made this patch.

It doesn’t exactly add a dc.contributor.author (there is no such thing in QDC).

The issue is only that there is no dc.creator element in the output.   The fact that it pulls data from the source document’s dc.creator and from the source document’s dc.contributor.author in order to populate the output document’s dc.creator doesn’t make it a separate issue, I don’t think.

Brad

PR #1062, opened 2015-09-17
DS-2583: Quality Control Reporting via REST API (v2)

terrywbrady:
Extend the REST API to support quality control reports. 
- GET /reports - Return a list of report tools built on the rest api
- GET /reports/{nickname} - Return a redirect to a specific report
- GET /filters - Return a list of use case filters available for quality control reporting
- GET /filtered-collections - Return collections and item counts based on pre-defined filters
- GET /filtered-collections/{collection_id} - Return items and item counts for a collection based on pre-defined filters
- GET /filtered-items - Retrieve a set of items based on a metadata query and a set of filters
## GET /reports

Configured in dspace/config/modules/rest.cfg

```
# return report tool html by name
#rest-report-url.collections = ../static/rest/index.html
#rest-report-url.item-query = ../static/rest/query.html
```
## GET /filters

Configured in dspace/config/modules/rest.cfg.  
These filters are referenced in the other reporting actions.
Each of these classes contains an enumerated list of useful ways to filter DSpace items.

```
# configurable filters for filter-collections and filtered-items
plugin.sequence.org.dspace.rest.filter.ItemFilterList = \
        org.dspace.rest.filter.ItemFilterDefs,\
        org.dspace.rest.filter.ItemFilterDefsMisc,\
        org.dspace.rest.filter.ItemFilterDefsPerm
```
## Sample Reporting Application

The following JavaScript application illustrates how these new REST actions can provide quality control reporting to repository managers.

https://github.com/Georgetown-University-Libraries/DSpace-Rest-Reports

PR #1085, opened 2015-10-01
DS-2583: Quality Control Reporting via REST API (v3)

terrywbrady:
Extend the REST API to support quality control reports. 
- GET /reports - Return a list of report tools built on the rest api
- GET /reports/{nickname} - Return a redirect to a specific report
- GET /filters - Return a list of use case filters available for quality control reporting
- GET /filtered-collections - Return collections and item counts based on pre-defined filters
- GET /filtered-collections/{collection_id} - Return items and item counts for a collection based on pre-defined filters
- GET /filtered-items - Retrieve a set of items based on a metadata query and a set of filters
## GET /reports

Configured in dspace/config/modules/rest.cfg

```
# return report tool html by name
#rest-report-url.collections = ../static/rest/index.html
#rest-report-url.item-query = ../static/rest/query.html
```
## GET /filters

Configured in dspace/config/modules/rest.cfg.  
These filters are referenced in the other reporting actions.
Each of these classes contains an enumerated list of useful ways to filter DSpace items.

```
# configurable filters for filter-collections and filtered-items
plugin.sequence.org.dspace.rest.filter.ItemFilterList = \
        org.dspace.rest.filter.ItemFilterDefs,\
        org.dspace.rest.filter.ItemFilterDefsMisc,\
        org.dspace.rest.filter.ItemFilterDefsPerm
```
## Sample Reporting Application

The following JavaScript application illustrates how these new REST actions can provide quality control reporting to repository managers.

https://github.com/Georgetown-University-Libraries/DSpace-Rest-Reports
## Note

This is a replacement for https://github.com/DSpace/DSpace/pull/1062


terrywbrady:
Closing... I seemed to have incorrectly created the branch that was the origin of this PR.  Resubmitting from a clean repo

dspace-api/src/main/java/org/dspace/content/dao/impl/ItemDAOImpl.java
```diff
@@ -103,6 +118,111 @@
         return iterate(query);
     }
 
+    enum OP {equals,not_equals,like,not_like,contains,doesnt_contain,exists,doesnt_exist,matches,doesnt_match;}
+    
+    @Override
+    public Iterator<Item> findByMetadataQuery(Context context, List<String> query_field, List<String> query_op, List<String> query_val, List<String> collection_ids, String regexClause) throws SQLException {
```
KevinVdV: There is to much "business" logic in this DAO method, the DAO should just be about running a database query. The transformation of strings to UUIDs, splitting of metadata value should go to the service layer.
The reason for this being if anybody else wants a different DAO then hibernate they would have to reimplement all this logic as well.

dspace-api/src/main/java/org/dspace/content/dao/impl/ItemDAOImpl.java
```diff
@@ -103,6 +118,111 @@
...
+        	}
+        	
+        	DetachedCriteria subcriteria = DetachedCriteria.forClass(MetadataValue.class,"mv");
+        	subcriteria.add(Property.forName("mv.dSpaceObject").eqProperty("item.id"));
+        	subcriteria.setProjection(Projections.property("mv.dSpaceObject"));
+        	
+        	if (!allFields) {
+        		DetachedCriteria  fieldCriteria = DetachedCriteria.forClass(MetadataField.class,"mfr");
+        		fieldCriteria.add(Property.forName("mv.metadataField").eqProperty("mfr.id"));
+        		fieldCriteria.setProjection(Projections.property("mfr.id"));
```
KevinVdV: Instead of splitting up the schema metatafield string & using it in a query, perhaps lookup the metadatafield itself. The MetadataField & MetadataSchema rows are cached across sessions so a query on MetadataField doesn't even hit the database but rather the cache.
It also results in a lot less code.
Make sure to perform the lookup from the service side.

dspace-rest/src/main/java/org/dspace/rest/FilteredCollectionsResource.java
```diff
@@ -0,0 +1,198 @@
...
+            throw new WebApplicationException(Response.Status.INTERNAL_SERVER_ERROR);
+        } catch (ContextException e) {
+            log.error(String.format("Could not read collection %d.  %s", collection_id, e.getMessage()));
+            throw new WebApplicationException(Response.Status.INTERNAL_SERVER_ERROR);
+		} finally {
+            if(context != null) {
+                try {
+                    context.complete();
+                } catch (SQLException e) {
+                    log.error(e.getMessage() + " occurred while trying to close");
```
KevinVdV: Perhaps also log the stacktrace, it is useful in a lot of cases.

dspace-rest/src/main/java/org/dspace/rest/FilteredItemsResource.java
```diff
@@ -0,0 +1,156 @@
...
+            throw new WebApplicationException(Response.Status.INTERNAL_SERVER_ERROR);
+        } catch (ContextException e) {
+            log.error("Unauthorized filtered item query. " + e.getMessage());
+            throw new WebApplicationException(Response.Status.INTERNAL_SERVER_ERROR);
+		} finally {
+            if(context != null) {
+                try {
+                    context.complete();
+                } catch (SQLException e) {
+                    log.error(e.getMessage() + " occurred while trying to close");
```
KevinVdV: Perhaps also log the stacktrace, it is useful in a lot of cases.

PR #1098, opened 2015-10-08
DS-2740 Conditionally show publisher for sherpa romeo, if present

helix84:
https://jira.duraspace.org/browse/DS-2740


helix84:
+1 to Peter's solution, tested

PR #1102, opened 2015-10-10
DS-2784 Update REST API to Jersey2

akotynski:
https://jira.duraspace.org/browse/DS-2784

### DIFF
--- a/dspace-rest/src/main/java/org/dspace/rest/DSpaceRestApplication.java
+++ b/dspace-rest/src/main/java/org/dspace/rest/DSpaceRestApplication.java
@@ -7,7 +7,6 @@ import org.springframework.web.filter.RequestContextFilter;
 public class DSpaceRestApplication extends ResourceConfig {
 
     public DSpaceRestApplication() {
-        register(RequestContextFilter.class);
         register(JacksonFeature.class);
         packages("org.dspace.rest");
     }
--- a/dspace-rest/src/main/java/org/dspace/rest/DSpaceRestApplication.java
+++ b/dspace-rest/src/main/java/org/dspace/rest/DSpaceRestApplication.java
@@ -1,3 +1,10 @@
+/**
+ * The contents of this file are subject to the license and copyright
+ * detailed in the LICENSE and NOTICE files at the root of the source
+ * tree and available online at
+ *
+ * http://www.dspace.org/license/
+ */
 package org.dspace.rest;
 
 import org.glassfish.jersey.jackson.JacksonFeature;
--- a/dspace-rest/src/main/java/org/dspace/rest/DSpaceRestApplication.java
+++ b/dspace-rest/src/main/java/org/dspace/rest/DSpaceRestApplication.java
@@ -1,0 +1,14 @@
+package org.dspace.rest;
+
+import org.glassfish.jersey.jackson.JacksonFeature;
+import org.glassfish.jersey.server.ResourceConfig;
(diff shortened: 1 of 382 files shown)


dspace-rest/src/main/java/org/dspace/rest/DSpaceRestApplication.java
```diff
@@ -0,0 +1,13 @@
+package org.dspace.rest;
```
peterdietz: Please add DSpace license header.

dspace-rest/src/main/java/org/dspace/rest/common/DSpaceObject.java
```diff
@@ -112,4 +114,8 @@ public String getUUID() {
     public void setUUID(String uuid) {
         this.uuid = uuid;
     }
+
+    private String createLink(ServletContext context){
+        return context.getContextPath() + "/" + English.plural(getType()) + "/" + getId();
+    }
```
peterdietz: I see what you mean, the link will refer to the ID, and if ID is integer, as opposed to the new UUID, then the link will be limited as new objects will not have old integer ID, but all objects will have uuid. To ensure compatibility, perhaps we would need to leave getLink to return the integerID, but getLinkUUID could use the new identifier. (this is because we don't otherwise have rest versioning, and don't want to break contracts).

peterdietz:
This looks good to me. Please resolve the missing license header issue. Are you aware of any changes in jersey 1x to 2x that are worth noting? While developing DSpace's REST API, jersey 2 hadn't yet been released.

peterdietz:
#1100 is for 5x, small jersey bump 1.17.1 to 1.19, and #1102 is bigger jersey bump 1.7 to 2.22.1

dspace-rest/src/main/java/org/dspace/rest/common/DSpaceObject.java
akotynski: Sure, I agree with you. But for now ID is always null.

akotynski:
jersey 2.22.1 comes with jackson 2.5.4 (1.x with 1.9.2), new DI freamwork (hk2) - jersey 1.x use own legacy freamwork and big [performance improvements](https://blogs.oracle.com/japod/entry/jersey_2_performance)

peterdietz:
+1 I've run and tested this code, and it everything appears to be working normally. I didn't notice any issues from the Jersey upgrade.

akotynski:
how about null ID in generated link?

dspace-rest/src/main/java/org/dspace/rest/RestIndex.java
```diff
@@ -37,12 +38,13 @@
  * @author Rostislav Novak (Computing and Information Centre, CTU in Prague)
  * 
  */
+@Component
```
akotynski: ups. sorry for that

peterdietz:
Sorry for the delay. You can change the link to have the UUID then. The client will already have to parse that field as a string, and it is the best url to access the object.

akotynski:
Ok, I removed legacy ID

helix84:
@peterdietz is it good to merge yet? I'd like to close some tickets that depend on this.

peterdietz:
I have merged this PR. REST API is now Jersey 2. Master + REST is now working. There was a momentary issue with PubMed (dspace-api, bringing jersey2), and REST using jersey1. This is all working now

PR #1105, opened 2015-10-14
[Task: DS-2807] Bundle bitstream ordering fix for hibernate

KevinVdV:
https://jira.duraspace.org/browse/DS-2807

This pull request with update the bundle2bitstream columns to properly use the order that hibernate requires. Some additional testing of this pull request would be great.
For postgres testing this on a larger database to make sure that the query is optimal, for oracle testing it on an actual oracle database with more then 2 items in it (ran a local oracle DB so it should work).


KevinVdV:
Bug can be fixed, but I still need to drop the "bitstream_order_legacy" column, which is just there for legacy purposes. Once the PR has been verified I will drop the column.

peterdietz:
+1, I've tested this, and a large site that wouldn't complete an ant update (from pre-hibernate to hibernate) even letting it run over a weekend, now completes the update in under 10 minutes. 

Bitstream order appears correct.

peterdietz:
Another situation where this PR came to help is: 
ERROR: column "bitstream_order" contains null values

During an ant update on a pre-6 DB. After I cherry-picked this commit, it then ant update's successfully.

helix84:
+1, this helped with https://jira.duraspace.org/browse/DS-3012

PR #1108, opened 2015-10-15
[DS-2813] fixed NullPointer in submission lookup if pubmed is down.

christian-scheible:
https://jira.duraspace.org/browse/DS-2813

If PubMed is down there will be a NullPointerException in PubmedOnlineDataLoader. To fix this return an empty list instead of null


helix84:
+1, tested

PR #1111, opened 2015-10-15
[DS-2592] Fix OpenUrl description.xml link

aschweer:
https://jira.duraspace.org/browse/DS-2592


helix84:
+, tested, now fixed in all themes

PR #1114, opened 2015-10-16
DS-2829 Add the logo setter to the Community DTO

arnodb:
https://jira.duraspace.org/browse/DS-2829

That fixes potential deserialization issues.
This is covered by a test on the generated JAXB schema.

Background: I generated a REST client with wadl2java and there was an issue with the logo property of the Community DTO. That might not be the only one issue but that is a start. Pull request for branch dspace-5_x is coming...

### DIFF
--- a/dspace-rest/src/main/java/org/dspace/rest/common/Community.java
+++ b/dspace-rest/src/main/java/org/dspace/rest/common/Community.java
@@ -181,6 +181,10 @@ public class Community extends DSpaceObject{
         return logo;
     }
 
+    public void setLogo(Bitstream logo) {
+        this.logo = logo;
+    }
+
     // Renamed because of xml annotation exception with this attribute and getSubCommunities.
     @XmlElement(name = "subcommunities", required = true)
 	public List<Community> getSubcommunities() {
--- a/dspace-rest/src/test/java/org/dspace/rest/common/TestJAXBSchema.java
+++ b/dspace-rest/src/test/java/org/dspace/rest/common/TestJAXBSchema.java
@@ -1,0 +1,70 @@
+/**
+ * The contents of this file are subject to the license and copyright
+ * detailed in the LICENSE and NOTICE files at the root of the source
+ * tree and available online at
+ *
+ * http://www.dspace.org/license/
+ */
+package org.dspace.rest.common;
+
+import static org.junit.Assert.assertEquals;
+
+import java.io.InputStream;
+import java.io.IOException;
+import java.io.StringWriter;
+import java.io.Writer;
+
+import javax.xml.bind.JAXBContext;
+import javax.xml.bind.SchemaOutputResolver;
+import javax.xml.transform.Result;
+import javax.xml.transform.stream.StreamResult;
+
+import org.apache.commons.io.IOUtils;
+import org.junit.Test;
+
+public class TestJAXBSchema {
+
+    private static class TestSchemaOutputResolver extends SchemaOutputResolver {
+
+        private final Writer output;
+
+        public TestSchemaOutputResolver(Writer output) {
+            this.output = output;
+        }
+
+        public Result createOutput(String namespaceUri, String suggestedFileName) throws IOException {
+            StreamResult result = new StreamResult(output);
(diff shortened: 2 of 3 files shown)


helix84:
Please, describe the issue and tell us how to reproduce it.

arnodb:
Hi,

the problem is:

* https://demo.dspace.org/rest/communities returns JSON with some logo properties in community objects (they are null but they are there)
* https://demo.dspace.org/rest/application.wadl/xsd0.xsd doesn't contain the logo property because there is no setter in the Community DTO

So I added the setter and covered this with a test that generates the JAXB schema.

Note that I am working on packaging a DSpace REST client, so I am currently patching the schema I get from the demo website before I use it to generate the client code.

On the PRs, I'm afraid the DTOs have changed from 5_x to master so that is why I picked the commit and patched it slightly. I need to rebase my 5_x PR now.

arnodb:
If you really want to reproduce the issue then have look at https://github.com/arnodb/dspace-rest-client, I've just patched xsd0.xsd in it, otherwise some test won't pass.

arnodb:
Hi @helix84 would you prefer me to log an issue somewhere in order to keep track of the issue?

helix84:
Yes, that's what we normally do, we keep track of bugs in [Jira](jira.duraspace.org/browse/DS) (has more features than GitHub) and only to PRs here. But don't worry too much, anything that has a PR won't get lost, either.

arnodb:
OK, I've just signed up Jira and created https://jira.duraspace.org/browse/DS-2829.

helix84:
Thank you. Can you please add the missing standard DSpace license header to the new files? Travis runs with license checks enabled, that's why the test build fails.

arnodb:
Hi, done. However I'm not sure what I did with the merge because there seem to be 20 commits in the PR whereas there is only one in reality.

helix84:
The correct way to do this would be to checkout arnodb/rest-dto-sanity, add a commit and push to arnodb (origin). Then another commit would show up here.

You can still fix up a PR by undoing as many commits as you need (git reset --hard HEAD^) and then force-pushing.

In any case, we can just cherry-pick a single commit, so no worries: d5b4cf4

arnodb:
Here we are. branch resynchronized, license header added. Over to you :-).

peterdietz:
Thank you Arnaud, I wasn't actually aware that REST/Jersey generated the wadl out of the box. I can recall a day that I was searching for how to get REST to output something akin to a WSDL, but for REST, a WADL.

I see that you have added a setter for the collection logo, however, I don't believe you could actually put the collection logo in DSpace REST. Would you mind going the extra mile, and uncomment this, and test it?  https://github.com/DSpace/DSpace/blob/77f696dad2aeae98f566e53f09780d6e070d3526/dspace-rest/src/main/java/org/dspace/rest/CollectionsResource.java#L452
```// dspaceCollection.setLogo(collection.getLogo()); // TODO Add this option.```

And test if you can set the collection logo. Also, thank you for adding the test case.

By the way, your other projects look interesting:
https://github.com/idmgroup/dspace-rest-client
https://github.com/idmgroup/DSpace-REST-Client-Perl

I think that lack of DSpace (REST) SDK's for languages is a barrier to people building integrations. And, at the same time, we can be refining the REST API itself. So, thanks for the help, on both fronts.

arnodb:
Hi Peter, yes I knew that we could generate a WADL descriptor but I was not aware that it was enabled by default. So I also discovered something. I'm doing that for my current employer.

If you need my email to discuss anything, you can dig in the commits log.

I will implement what you ask me to ASAP so that this change goes upstream seamlessly.

Note: I spotted that when I create an item, its name it not saved, only its metadata. I may report a bug soon about that.

arnodb:
Hi Peter, actually I don't know how to convert a REST Bitstream object into an API Bitstream object. I have just looked at addItemBitstream and it seems far from being trivial.

peterdietz:
Okay, we can leave the setting of logo / bitstream off. Yes, non-trivial indeed.
Overall, +1

PR #1117, opened 2015-10-20
DS-2637 configurable pagination for discovery

jonas-atmire:
This contribution adds the possibility to configure the pagination for discovery on the fly.
Initially, the pagination for discovery was hard-set to 10 results per page.
This fix adds the same configurability to the discovery pagination that was already in place in the standard browse by pages, making it possible to change the number of results per page through a similar gear icon.
This contribution is based on the following JIRA ticket - https://jira.duraspace.org/browse/DS-2637
Atmire (http://www.atmire.com) has been commissioned by CGIAR (http://www.cgiar.org/) to fix this particular JIRA ticket as wel as helping improve the usability of DSpace by contributing it to the community.

### DIFF
--- a/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/artifactbrowser/ConfigurableBrowse.java
+++ b/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/artifactbrowser/ConfigurableBrowse.java
@@ -121,7 +121,8 @@ public class ConfigurableBrowse extends AbstractDSpaceTransformer implements
 
     /** The options for results per page */
     private static final int[] RESULTS_PER_PAGE_PROGRESSION = {5,10,20,40,60,80,100};
-    
+    private int currentOffset = 0;
+
     /** Cached validity object */
     private SourceValidity validity;
 
@@ -293,7 +294,9 @@ public class ConfigurableBrowse extends AbstractDSpaceTransformer implements
 
         // If there are items to browse, add the pagination
         int itemsTotal = info.getTotal();
-        if (itemsTotal > 0) 
+        currentOffset=info.getOffset();
+
+        if (itemsTotal > 0)
         {
             //results.setSimplePagination(itemsTotal, firstItemIndex, lastItemIndex, previousPage, nextPage)
             results.setSimplePagination(itemsTotal, browseInfo.getOverallPosition() + 1,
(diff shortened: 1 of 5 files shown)


dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/discovery/SearchFacetFilter.java
```diff
@@ -452,13 +469,14 @@ private String getNextPageURL(SearchFilterParam browseParams, Request request) t
 
     private String getPreviousPageURL(SearchFilterParam browseParams, Request request) throws UnsupportedEncodingException, UIException {
         //If our offset should be 0 then we shouldn't be able to view a previous page url
-        if (0 == queryArgs.getFacetOffset() && Util.getIntParameter(request, "offset") == -1)
+        boolean currentOffsetSmallerOrEqualTO0 = currentOffset <= 0;
```
aschweer: It would be better to use a variable name that's tied to the meaning -- something like `canGoBack` or `linkToPreviousPage`. At the very least, avoid having the upper-case letter O right next to the digit 0.

dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/discovery/SearchFacetFilter.java
```diff
@@ -344,6 +348,19 @@ public void addBody(Body body) throws SAXException, WingException, UIException,
         }
     }
 
+    private void updateQueryResultsAndOffset(Request request, DSpaceObject dso) {
```
aschweer: Perhaps reformat this method to follow the DSpace [coding conventions](https://wiki.duraspace.org/display/DSPACE/Code+Contribution+Guidelines#CodeContributionGuidelines-CodingConventions).

dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/artifactbrowser/ConfigurableBrowse.java
```diff
@@ -792,6 +794,20 @@ else if (bi.isItemIndex() && !bi.isInternalIndex())
         return params;
     }
 
+    private void updateOffset(Request request, BrowseParams params) {
```
aschweer: Perhaps reformat this method to follow the DSpace [coding conventions](https://wiki.duraspace.org/display/DSPACE/Code+Contribution+Guidelines#CodeContributionGuidelines-CodingConventions).

aschweer:
I've noted a few style issues; it would be great if these could be addressed. Apart from that, this has my +1 - tested using Mirage 1 and 2. Since this is a relatively small change but a UX improvement, I'm happy to have this accepted into 5.4 already.

PR #1121, opened 2015-10-22
DS-2830 add proper synchronization in TokenHolder

arnodb:
https://jira.duraspace.org/browse/DS-2830

Hi, I have just figured out that the token management in the REST API is not thread-safe. So if the API is consumed concurrently by multiple scripts then one may encounter some erroneous 401/403 statuses even if the token is present and valid. 

In TokenHolder.java everything is static and the access to the tokens and persons hashes should be synchronized.

### DIFF
--- a/dspace-rest/src/main/java/org/dspace/rest/TokenHolder.java
+++ b/dspace-rest/src/main/java/org/dspace/rest/TokenHolder.java
@@ -61,19 +61,21 @@ public class TokenHolder
             context = new org.dspace.core.Context();
             EPerson dspaceUser = EPerson.findByEmail(context, user.getEmail());
 
-            if ((dspaceUser == null) || (!dspaceUser.checkPassword(user.getPassword())))
-            {
-                token = null;
-            }
-            else if (tokens.containsKey(user.getEmail()))
-            {
-                token = tokens.get(user.getEmail());
-            }
-            else
-            {
-                token = generateToken();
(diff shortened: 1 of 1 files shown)


peterdietz:
Looks good to me. Would need to apply to master branch as well.

peterdietz:
I see that master has been changed to use Guava, so maybe the relevant collection there would be to change that to a synchronizedBiMap.
http://docs.guava-libraries.googlecode.com/git-history/v13.0.1/javadoc/com/google/common/collect/Maps.html#synchronizedBiMap(com.google.common.collect.BiMap)

arnodb:
Possibly, I didn't have a look at master actually. I will try to ASAP. By the way, I spotted that REST logout cannot be used as it logs out the user regardless of any kind of session. At least in 5_x, I don't know for master. That means that one must not try to clean things up by logging out of the REST API.

arnodb:
I've just had a look and it is probably better to put specific synchronization because there are multiple calls that should operate without the bimap to change state.

arnodb:
Do you want the PR for https://github.com/arnodb/DSpace/commit/f0adf320d0b5df0b4879b12733d5b9f7743e2b69 ?

peterdietz:
I have created a PR to master from your above commit / branch. Thanks

arnodb:
Thank you @peterdietz.

PR #1127, opened 2015-10-27
DS-2845 Handle case where metadata field does not map to a valid schema

terrywbrady:
See https://jira.duraspace.org/browse/DS-2845

This seemed to resolve the problem that I encountered.


peterdietz:
I would be tempted to do some debug logging to see why some metadata field doesn't fit. Maybe there are some nulls in the metadatavalue table, or other mismatching.

terrywbrady:
@peterdietz , I added some more diagnostics to the JIRA ticket

terrywbrady:
@KevinVdV identified that the underlying problem was related to the database migration script.  Closing this PR as unnecessary.

PR #1133, opened 2015-10-28
DS-2827: Fix parent/sub community mapping

terrywbrady:
See https://jira.duraspace.org/browse/DS-2827


peterdietz:
+1 I have tested the before and after of this, and this is a major fix. Before, the XMLUI community-list was in a completely unfamiliar hierarchy sub-communities at the top level, so more like anarchy, and now it is the typical hierarchy. 

I have noticed that the sub-communities are not ordered A-Z, but rather Z-A

PR #1134, opened 2015-10-29
DS-2831 connections cleanup, context reuse (master)

arnodb:
Hi,

this is the master version of https://github.com/DSpace/DSpace/pull/1122.

Plus a couple of REST context completion fixes.


peterdietz:
+1, The REST fixes are good. I must have assumed that using services would commit the context. And the EventService changes already were reviewed by proxy by Tim in #1122

peterdietz:
Tested the 5x version #1122, still is very similar, and to master, so merging.

PR #1148, opened 2015-11-04
DS-2867: Obvious fix for JSPUI edit item form

pnbecker:
This PR can be tested simply: create an item in JSPUI and add at least two files to that item. Publish it, login as administrator if not logged in as one already and click the edit button in item view. Without this PR this produces an Internal System Error, with this patch the edit item form is opened without any problems.


pnbecker:
I merged it as the Bug was obvious, the solution simple and it was obviously a result from changing ID type from integer to UUID.

PR #1150, opened 2015-11-04
DS-2830 (Master) add proper synchronization in TokenHolder

peterdietz:
Master version to synchronize /login /logout token usage of REST
https://jira.duraspace.org/browse/DS-2830

5x version:
https://github.com/DSpace/DSpace/pull/1121


peterdietz:
We have already merged DSPR #1121 to 5x, and this goes to master.

PR #1153, opened 2015-11-06
DS-2871 mobile theme: correct order of imports

helix84:
https://jira.duraspace.org/browse/DS-2871


helix84:
This one is very obvious, I'll just merge it.

PR #1164, opened 2015-11-12
[DS-2883] Resurrect TabFileUsageEventListener

mwoodiupui:
Someone found a use for this, so I recovered it from history (and tightened up the code a wee bit).  This would give DSpace a simple, compact primary source for usage cases, letting Solr be what it wants to be: a cache representing a primary source. It's also a handy format for feeding to dedicated statistical tools like SAS or R, for doing exploratory statistics. These cases would not be lost in the clutter of the general dspace.log file, allowing us to separate data with long-term value from data with only short-term (i.e. diagnostic) value. 

Lightly tested with XMLUI.

https://jira.duraspace.org/browse/DS-2883

### DIFF
--- a/dspace-api/src/main/java/org/dspace/usage/TabFileUsageEventListener.java
+++ b/dspace-api/src/main/java/org/dspace/usage/TabFileUsageEventListener.java
@@ -15,16 +15,20 @@ import java.io.PrintWriter;
 import java.text.SimpleDateFormat;
 import java.util.Date;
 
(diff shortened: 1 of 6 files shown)


hardyoyo:
+1 it will be nice to have the option to use this to keep a permanent log of usage data. That is work for another day, but this is an important first step. Thanks!

pnbecker:
+1 by code inspection, haven't tested it.

PR #1169, opened 2015-11-13
DS-2890: Prevent NPE in Util.getUUIDParameter(....).

pnbecker:


pnbecker:
I merged this as the code was simple and the reason for the NPEs obvious.

PR #1170, opened 2015-11-14
[DS-2763] EHCache throwing a NotSerializableException: org.dspace.content.CollectionServiceImpl

mwoodiupui:
https://jira.duraspace.org/browse/DS-2763

Thanks to FindBugs for pointing out numerous difficulties with Serializable.

### DIFF
--- a/dspace-api/src/main/java/org/dspace/app/bulkedit/DSpaceCSV.java
+++ b/dspace-api/src/main/java/org/dspace/app/bulkedit/DSpaceCSV.java
@@ -9,12 +9,8 @@ package org.dspace.app.bulkedit;
 
 import org.apache.commons.lang3.StringUtils;
 import org.dspace.authority.AuthorityValue;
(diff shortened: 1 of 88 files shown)


mwoodiupui:
I think this is about as Serializable-clean as can be done without more radical restructuring.  A few classes *must* be Serializable but cannot be due to holding references to non-Serializable classes from non-DSpace projects.

I would appreciate some inspection of JSPUI from someone more familiar with Servlet coding.  I had to make some more substantial changes to many Servlets and tag classes.

hardyoyo:
+1, builds, everything still runs (i.e. "smoke test passed"), and I can confirm the serialization error messages are now gone from dspace.log. Unit tests are not currently working in my environment, however I can see the green checkmark above indicating that TravisCI was able to run them successfully.

KevinVdV:
Did an inspection of the code & it looks good, so +1.

helix84:
Cool, glad we can finally squash this bug. Thank you!

PR #1178, opened 2015-11-20
[DS-2905] Fix format bugs with the EPerson CLI scripts & UUIDs

KevinVdV:
Replacing %d by %s & adding a toString to the UUID retrieval should fix these issues.

### DIFF
--- a/dspace-api/src/main/java/org/dspace/eperson/EPersonCLITool.java
+++ b/dspace-api/src/main/java/org/dspace/eperson/EPersonCLITool.java
@@ -191,7 +191,7 @@ public class EPersonCLITool {
         try {
             ePersonService.update(context, eperson);
             context.complete();
-            System.out.printf("Created EPerson %d\n", eperson.getID());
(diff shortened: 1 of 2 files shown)


dspace-api/src/main/java/org/dspace/eperson/EPersonCLITool.java
```diff
@@ -416,7 +416,7 @@ private static int cmdList(Context context, String[] argv)
         try {
             for (EPerson person : ePersonService.findAll(context, EPerson.EMAIL))
             {
-                System.out.printf("%d\t%s/%s\t%s, %s\n",
+                System.out.printf("%s\t%s/%s\t%s, %s\n",
                         person.getID(),
```
mwoodiupui: Don't we need .toString() here as well?

mwoodiupui:
After considering the above line note, +1.  Obviously necessary.

KevinVdV:
You are correct, fixed.

PR #1179, opened 2015-11-20
[DS-2906] Remove duplicate null check

KevinVdV:
https://jira.duraspace.org/browse/DS-2906


dspace-jspui/src/main/java/org/dspace/app/webui/components/VersioningItemHome.java
```diff
@@ -91,26 +91,24 @@ public void process(Context context, HttpServletRequest request,
 				throw new PluginException(e.getMessage());
 			}
 
-			if (latestVersion != null) {
```
mwoodiupui: I don't see where the balancing close-bracket was removed.

KevinVdV: Might be warped in the github view, the removal bracket is shown in IntelliJ....

helix84:
+1 by inspection

PR #1181, opened 2015-11-20
[DS-2908] Exception is created but not thrown in the CollectionServiceImpl

KevinVdV:


mwoodiupui:
+1 obviously correct.

hardyoyo:
+1 by inspection

PR #1182, opened 2015-11-20
[DS-2909] Fix EPerson netId setter

KevinVdV:
Was a self assigned variable so the netid could never be set.


mwoodiupui:
+1 obviously correct.

hardyoyo:
+1 by inspection

PR #1183, opened 2015-11-20
[DS-2910] SherpaService timeout variable has no default value

KevinVdV:
https://jira.duraspace.org/browse/DS-2910


mwoodiupui:
This changes the default socket timeout for a Sherpa connection from zero (infinite wait) to 5 seconds, which is probably a good idea.  +1

PR #1184, opened 2015-11-20
[DS-2901] Fix broken equals methods found by FindBugs

KevinVdV:


mwoodiupui:
+1 clearly needed.  The failure is due to the demise of the Codehaus repository.

hardyoyo:
+1 obviously correct

PR #1185, opened 2015-11-20
[DS-2911] Inspect and fix unused variables

KevinVdV:
https://jira.duraspace.org/browse/DS-2911

### DIFF
--- a/dspace-rest/src/main/java/org/dspace/rest/BitstreamResource.java
+++ b/dspace-rest/src/main/java/org/dspace/rest/BitstreamResource.java
@@ -165,7 +165,6 @@ public class BitstreamResource extends Resource
         {
             context = createContext(getUser(headers));
             org.dspace.content.Bitstream dspaceBitstream = findBitstream(context, bitstreamId, org.dspace.core.Constants.READ);
-            //TODO why isn't the above used...
             policies = new Bitstream(dspaceBitstream,"policies", context).getPolicies();
 
             context.complete();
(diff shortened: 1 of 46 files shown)


KevinVdV:
The following differences points to potential bugs, the other changes are mere code cleanup:
https://github.com/DSpace/DSpace/pull/1185/files#diff-c027cc8129f5476cfe253043fcb66971R1263
https://github.com/DSpace/DSpace/pull/1185/files#diff-36833b35a6f2d36c81f0d875d3bcd41aR680
https://github.com/DSpace/DSpace/pull/1185/files#diff-a17391bd2bf6fed372045f6ddf202430R182

dspace-rest/src/main/java/org/dspace/rest/BitstreamResource.java
```diff
@@ -165,7 +165,6 @@ public Bitstream getBitstream(@PathParam("bitstream_id") String bitstreamId, @Qu
         {
             context = createContext(getUser(headers));
             org.dspace.content.Bitstream dspaceBitstream = findBitstream(context, bitstreamId, org.dspace.core.Constants.READ);
-            List<org.dspace.authorize.ResourcePolicy> resourcePolicies = authorizeService.getPolicies(context, dspaceBitstream);
             //TODO why isn't the above used...
```
mwoodiupui: Perhaps the comment should be removed as well, since it appears to have referred to the removed assignment and would now be confusing.

KevinVdV: Good point, comment has been removed.

PR #1186, opened 2015-11-20
[DS-2912] Sub daily script is broken

KevinVdV:
https://jira.duraspace.org/browse/DS-2912


mwoodiupui:
This was clearly wrong.  +1

PR #1187, opened 2015-11-23
[DS-2913] Fix community parent logging infinite loop

KevinVdV:
https://jira.duraspace.org/browse/DS-2913


terrywbrady:
+1.  This fix is very helpful.

PR #1191, opened 2015-11-24
DS-2916 BitstreamDAOImpl contains incorrect database linking

jonas-atmire:
The implementation to find by Collection/Community/Item contains relations that don't exists between the database tables


KevinVdV:
+1, this was a leftover of the old BundleBitstream object that was removed a while ago, merging.

PR #1193, opened 2015-11-25
[DS-2921] Item registration ignores bundles

mwoodiupui:
https://jira.duraspace.org/browse/DS-2921

Briefly tested atop DS-79 (which I was testing when I tripped over this).


KevinVdV:
+1

PR #1194, opened 2015-11-26
DS-2852 Discovery search labels show authority key instead of author name

tmtvl:
https://jira.duraspace.org/browse/DS-2852

Adjusted the discovery theming of Mirage 2 to show author names instead
of authority values.

### DIFF
--- a/dspace-xmlui/src/main/java/org/dspace/app/xmlui/utils/XSLUtils.java
+++ b/dspace-xmlui/src/main/java/org/dspace/app/xmlui/utils/XSLUtils.java
@@ -7,6 +7,12 @@
  */
 package org.dspace.app.xmlui.utils;
 
+import org.dspace.authority.AuthorityValue;
+import org.dspace.authority.AuthorityValueFinder;
+import org.dspace.core.Context;
+
+import java.sql.SQLException;
+
 /**
  * Utilities that are needed in XSL transformations.
  *
@@ -53,4 +59,21 @@ public class XSLUtils {
         return string.substring(0, targetLength) + " ...";
 
     }
+
+	public static String getDiscoveryFilterDisplay(String display) throws SQLException {
+		Context ctx = new Context();
+
+		if (display != null) {
+			if (display.length() == 36 && display.contains("-")) {
+				AuthorityValue authorityValue = new AuthorityValueFinder().findByUID(ctx, display);
+				if (authorityValue != null) {
+					display = authorityValue.getValue();
+				}
+			}
+		}
+
+		ctx.complete();
+
+		return display;
+	}
 }
--- a/dspace-xmlui-mirage2/src/main/webapp/templates/discovery_simple_filters.hbs
+++ b/dspace-xmlui-mirage2/src/main/webapp/templates/discovery_simple_filters.hbs
@@ -8,5 +8,5 @@
 
 -->
 {{#each orig_filters}}
-    <label href="#" class="label label-primary" data-index="{{index}}">{{query}}&nbsp;&times;</label>
+    <label href="#" class="label label-primary" data-index="{{index}}">{{display_query}}&nbsp;&times;</label>
 {{/each}}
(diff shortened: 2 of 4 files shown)


hardyoyo:
I've been testing this PR in combination with my institutions DSpace-5_x-based fork of DSpace, and after manually applying the changes to our Mirage2 overlays, rebuilding the Discovery index and the Authority index, I still see this text on the author browse pages: Browsing by Author "c34b248d-7873-40ea-95f4-9f77cac0c51b" which may simply mean there are more templates that need changing. I'll see if I can figure it out.

hardyoyo:
It looks like [this line](https://github.com/DSpace/DSpace/blob/9d747c7c08a90296c1fea1c16b1dd83ef246b314/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/artifactbrowser/ConfigurableBrowse.java#L909) is where the code responsible for the "Browsing by Author" text is generated. Now to figure out how to make that text work with the authority index.

aschweer:
Presumably the ORCID browse doesn't return true for `isAuthorityIndex` and/or a different way for obtaining the label for a given authority key is needed? Since it looks like there is logic just underneath the line linked to by @hardyoyo that has special logic for authority-controlled metadata fields.

hardyoyo:
Aha: https://wiki.duraspace.org/display/DSDOC5x/ORCID+Integration I missed the lines that state:

    The browse indexes need to point to the new authority-controlled index: webui.browse.index.2 = author:metadata:dc.contributor.*,dc.creator:text should become webui.browse.index.2 = author:metadataAuthority:dc.contributor.author:text

I will try that configuration and will re-index.

hardyoyo:
Hmm... with the suggested configuration noted above:

webui.browse.index.2 = author:metadataAuthority:dc.contributor.author:text

And a full rebuild of the Discovery index, the author browse is now full of 00058c59-3b1b-49f1-b215-2f70ceee403e and 00062e32-8ada-4cb0-9d14-5f68fbc066b2 (etc. etc.) where before at least the main browse pages included names. This probably doesn't have anything to do with this PR... but, I'm frankly kind of lost at this point. I'm going to try rebuilding the Authority cache (dspace index-authority) just to see if that impacts anything.

bram-atmire:
Browse heading should be a different problem normally
https://jira.duraspace.org/browse/DS-2401

dspace-xmlui/src/main/java/org/dspace/app/xmlui/utils/XSLUtils.java
```diff
@@ -53,4 +59,21 @@ public static String shortenString(String string, int targetLength, int maxDevia
         return string.substring(0, targetLength) + " ...";
 
     }
+
+	public static String getDiscoveryFilterDisplay(String display) throws SQLException {
+		Context ctx = new Context();
```
rradillen: Still needs some handling of the context so it doesn't leak on exception.

hardyoyo:
I think part of the problem here is that the documentation is wrong. See this page:

https://wiki.duraspace.org/display/DSDOC5x/ORCID+Integration#ORCIDIntegration-Configuration

 (my fault, I copied/pasted suggested text into the docs without verifying it). However, even with a proper configuration for an authority-controlled browse index --a good example of such is already present in the dspace.cfg, just commented out: https://github.com/DSpace/DSpace/blob/master/dspace/config/dspace.cfg#L1116

... I still see the authority key in author browse headings.

I am now looking into the possibility that the patch suggested in this thread:
http://dspace.2283337.n4.nabble.com/Browse-authority-controlled-author-names-by-text-value-not-by-authority-key-td4672708.html
may be required to address the problem of the authority key showing up in browse headings.

mwoodiupui:
Reviewing in IRC 29-Mar-2017, any progress?

hardyoyo:
Much has transpired since then. I now am employed elsewhere, however, we intend to use the ORCID integration for one of our instances of DSpace. It's a simple matter to apply Keiji Suzuki's patch and make a PR against this PR, I'll tackle that right now.

hardyoyo:
Keiji Suzuki's patch https://gist.githubusercontent.com/zuki/11365437/raw/9a646a65ab95ee90bdf8404d5ac53740324a7f38/SolrBrowseCrateDAO.java.diff

Modifies a file that does not exist in DSpace 6.x. Which leads me to ask: have we verified this bug still exists in 6.x?

tomdesair:
@tdonohue This PR can be closed in favor of https://github.com/DSpace/DSpace/pull/1701

PR #1196, opened 2015-11-27
DS-2923: Update DataCite default configuration.

pnbecker:


pnbecker:
https://jira.duraspace.org/browse/DS-2923

PR #1199, opened 2015-11-30
DS-2893 Mirage2: printing an item page includes the URL to the bitstreams

tmtvl:
https://jira.duraspace.org/browse/DS-2893

Quick fix for printing URLs.


hardyoyo:
+1, tested, solves the problem, thanks!

helix84:
+1, looks sane, merging

PR #1203, opened 2015-12-03
DS-2873

PhilipVis:
The Group Edit Form didn't show groups or epersons because SelectEPersonTag.setSelected() expects an EPerson object or array, and it was receiving a List.  

https://jira.duraspace.org/browse/DS-2873


KevinVdV:
+1 by inspection

helix84:
+1, tested

PR #1204, opened 2015-12-03
DS-2851 Administrators cannot create a new user through JSPUI

PhilipVis:
Attribute canLogIn is null when a new EPerson is being created, so when EPerson.canLogIn() was called from eperson-edit.jsp this caused a nullpointer exception. Now method BooleanUtils.isTrue() from apache commons is used to prevent this. 

https://jira.duraspace.org/browse/DS-2851


marsaoua:
I've tested the fix and it works fine.

helix84:
+1, tested

PR #1209, opened 2015-12-06
DS-2936 REST-API /handle endpoint broken

bram-atmire:
https://jira.duraspace.org/browse/DS-2936

After the lookup, the context is now being closed. As a result, the context won't be closed in the finalizer anymore. The finalizer also potentially messes with the return statement, so the return was moved to the bottom of the method. The code is tested and works, but I'm not fully sure it corresponds to best practice.


KevinVdV:
+1, will also need to be cherry picked in the master.

PR #760, opened 2014-11-22
DS-2315: export Simple Archive Format without bitstreams

kohts:
https://jira.duraspace.org/browse/DS-2315

PR #1210, opened 2015-12-07
DS-2315: export Simple Archive Format without bitstreams

kohts:
Replacement for https://github.com/DSpace/DSpace/pull/760

https://jira.duraspace.org/browse/DS-2315

### DIFF
--- a/dspace-api/src/main/java/org/dspace/app/itemexport/ItemExportCLITool.java
+++ b/dspace-api/src/main/java/org/dspace/app/itemexport/ItemExportCLITool.java
@@ -71,6 +71,10 @@ public class ItemExportCLITool {
         options.addOption("z", "zip", true, "export as zip file (specify filename e.g. export.zip)");
         options.addOption("h", "help", false, "help");
 
+        // as pointed out by Peter Dietz this provides similar functionality to export metadata
+        // but it is needed since it directly exports to Simple Archive Format (SAF)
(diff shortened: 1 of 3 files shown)


pnbecker:
This PR is just a new clean PR of #760. #760 was reviewed by helix84 in November 2014 and mhwood in June 2015. @kohts rebase his PR against the service API.

I think we should add another small review and test as the service API is a major change. Nevertheless the original PR of course existed before our PR deadline for DSpace 6.0.

kohts:
This one has been tested and I confirm it works as expected.

The only thing it needs to make sense is DS-2844; to be more specific, one index from it
-- CREATE INDEX metadatavalue_object on metadatavalue(dspace_object_id)

Without this index whole export (either with -x or without it) takes much more time,
than expected because of slow Postgres (this is not really an individual issue
of this PR, but rather a problem for a lot of other places).

Please merge.

helix84:
+1, I re-tested the PR and it still works fine after the migration to service api.

I'd have one minor request tangential to this PR - please include item handle in the message:
Exporting Item 19474d60-9dba-47ea-b8d2-a5a7e6bf43b5 to /tmp/export/69

https://github.com/DSpace/DSpace/blob/master/dspace-api/src/main/java/org/dspace/app/itemexport/ItemExportServiceImpl.java#L125

kohts:
Added logging of item handle during export.

PR #1213, opened 2015-12-10
[DS-2942] Fix (template) item metadata edit

KevinVdV:
http://jira.duraspace.org/browse/DS-2942

This pull request not only fixes the -1 casting problem but also ensures that the edit collection template item page works properly. This page was broken by using a "==" to compare UUID's instead of equals.


mwoodiupui:
+1.  Works, code is what I would expect.

helix84:
+1 tested, works correctly

PR #1221, opened 2015-12-16
DS-2945 Submitting to a collection without any workflow as normal user ends in an Exception

PhilipVis:
https://jira.duraspace.org/browse/DS-2945


KevinVdV:
Haven't tested it yet but looks sensible, so +1.

KevinVdV:
Tested the PR & it resolves the issue.

PR #1225, opened 2015-12-21
DS-2962: Robots.txt disallows site /discovery but not /handle/*/discover

rradillen:
https://jira.duraspace.org/browse/DS-2962

### DIFF
--- a/dspace-xmlui/src/main/webapp/static/robots.txt
+++ b/dspace-xmlui/src/main/webapp/static/robots.txt
@@ -11,11 +11,14 @@ Sitemap: ${dspace.url}/htmlmap
 User-agent: *
 # Disable access to Discovery search and filters
 Disallow: /discover
+Disallow: /handle/${handle.prefix}/*/discover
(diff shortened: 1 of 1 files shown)


KevinVdV:
+1, but if we accept for master it might collide with: https://github.com/DSpace/DSpace/pull/1104

helix84:
Just a note about asterisk in URL - it's not supported by the robots exclusion protocol (http://www.robotstxt.org/robotstxt.html), although it may be supported by the Google crawler (https://en.wikipedia.org/wiki/Robots_exclusion_standard#Universal_.22.2A.22_match).

bram-atmire:
Google, Bing, Yahoo and Ask support * and $ according to this information:
https://developers.google.com/webmasters/control-crawl-index/docs/robots_txt?hl=en#url-matching-based-on-path-values

hardyoyo:
+1 I've been running this patch in production from the day it was submitted, 12/21/2015

PR #1231, opened 2015-12-23
DS-2965 - Allow cache clear during re-index

terrywbrady:
The PR has enabled me to successfully index tens of thousands of items in my DSpace 6 instance.

For method visibility purposes, I added the clearCache method to the Context object.

I wonder if a more generic implementation could be placed on AbstractHibernateDAO.iterate() to manage the cache size when pulling a large number of objects in different contexts.

https://jira.duraspace.org/browse/DS-2965

### DIFF
--- a/dspace-api/src/main/java/org/dspace/core/Context.java
+++ b/dspace-api/src/main/java/org/dspace/core/Context.java
@@ -20,6 +20,7 @@ import org.dspace.event.factory.EventServiceFactory;
 import org.dspace.event.service.EventService;
 import org.dspace.storage.rdbms.DatabaseConfigVO;
 import org.dspace.utils.DSpace;
+import org.hibernate.Session;
 import org.springframework.util.CollectionUtils;
 
 /**
@@ -594,5 +595,9 @@ public class Context
 
     public void shutDownDatabase() throws SQLException {
         dbConnection.shutdown();
+    }
+    
+    public void clearCache() throws SQLException {
+    	((Session)this.getDBConnection().getSession()).clear();
     }
 }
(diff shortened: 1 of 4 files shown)


dspace-api/src/main/java/org/dspace/core/Context.java
```diff
@@ -595,4 +596,8 @@ protected void finalize() throws Throwable
     public void shutDownDatabase() throws SQLException {
         dbConnection.shutdown();
     }
+    
+    public void clearCache() throws SQLException {
+    	((Session)this.getDBConnection().getSession()).clear();
```
KevinVdV: I think we should extend the DBConnection interface to have a clearCache() & have the hibernate implementation call this clear.
Having it in here ties the context object to close to the hibernate implementation (which I want to avoid if we ever want to drop hibernate).

KevinVdV:
Aside from my note above a +1 from me !

A small note on the iterator usage, I don't think that the problem is solely limited to the iterator. I improved the iterator here: https://github.com/DSpace/DSpace/pull/1131, but when indexing in discovery all resource policies, bundles, bitstream, .... are also loaded in the cache.
I believe these other objects are the root cause.

dspace-api/src/main/java/org/dspace/core/Context.java
terrywbrady: I am just getting back to this.  getDBConnection() has package access, so I will keep the clearCache method in Context.

terrywbrady:
Recreated in https://github.com/DSpace/DSpace/pull/1276

PR #1232, opened 2015-12-24
DS-2975 - remove item from dspace

akotynski:
https://jira.duraspace.org/browse/DS-2975


KevinVdV:
Would it not be better to call: itemService.delete(), because I fear that this might lead to a ConcurrentModificationException.

akotynski:
Thanks for advice. PR updated

KevinVdV:
Tested this PR locally, it works so merging.

Thanks @akotynski !

PR #1236, opened 2016-01-07
Ds 2629 excel media filter

KevinVdV:
DS-2629 Add ability to filter Excel (xls and xlsx) files for full text searching

https://jira.duraspace.org/browse/DS-2629

Replaces pull request: https://github.com/DSpace/DSpace/pull/98, differences: 
- Ensured compilation against latest master
- Removed the verbose print extracted text as it would really clutter the output
- Compressed the string to inputStream from 3 lines into a single one.
- Removed obsolete constructor call to the "ExtractorFactory"
- Removed a TODO that I verified is correct

### DIFF
--- a/dspace-api/src/main/java/org/dspace/app/mediafilter/ExcelFilter.java
+++ b/dspace-api/src/main/java/org/dspace/app/mediafilter/ExcelFilter.java
@@ -7,15 +7,17 @@
  */
 package org.dspace.app.mediafilter;
 
-import java.io.ByteArrayInputStream;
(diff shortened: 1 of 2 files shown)


KevinVdV:
I tested this code locally & it works as advertised, so a +1 from me. @hardyoyo was interested in testing the old pull request, would you mind testing this updated one ?

peterdietz:
+1 Looks good to me by inspection.I haven't run it / verified, but looks good.

PR #1238, opened 2016-01-11
DS-2985: Use correct arg name to make update-doi and delete-doi execute

cwilper:
https://jira.duraspace.org/browse/DS-2985


KevinVdV:
+1, straight forward fix. Merging.

PR #1240, opened 2016-01-14
[DS-2989] View workspace item broken in JSPUI

KevinVdV:
https://jira.duraspace.org/browse/DS-2989

Create a list instead of an array for the collections to be passed along to the jspui.


helix84:
+1, this fixed the problem

PR #1242, opened 2016-01-18
Fix for DS-2990 Unable to create subcommunites

cjuergen:
https://jira.duraspace.org/browse/DS-2990

Adds parent community id, so that subcommunities are created as subcommunities and not top level communities


KevinVdV:
Is this pull request complete ? I can reproduce the problem but the commits don't appear to fix the issue.

I believe that this IF statement might be wrong: https://github.com/cjuergen/DSpace/blob/DS-2990/dspace-jspui/src/main/java/org/dspace/app/webui/servlet/admin/EditCommunitiesServlet.java#L155

cjuergen:
A sorry you are right the change in the servlet did not belong to this, the change in the jsp alone is ok and fixes the bug.  I'll delete the pull request, fix it and create a new one.

PR #1243, opened 2016-01-18
Fix for DS-2991

cjuergen:
Browse by communities and collections, display the complete tree instead of only the top level communities


KevinVdV:
Tested this locally & this fixed the issue.

PR #1255, opened 2016-01-28
[DS-2992] Selection list for metadata in edit item or item template is not sorted alphabetically anymore

KevinVdV:
https://jira.duraspace.org/browse/DS-2992

This pull request not only fixes the metadata retrieval order in the edit item view but also on the metadata field registry edit pages.


helix84:
+1, tested

PR #1256, opened 2016-01-29
[DS-3008] Fix metadata order storage

KevinVdV:
Added update() method to the DSpaceObjectService, this will (if the metadata has changed) update the "place" column.
All DSpaceObjectServices call this method using super(), this solves the issue.

https://jira.duraspace.org/browse/DS-3008


helix84:
+1 tested in XMLUI submission forms, fixes the problem

PR #1262, opened 2016-02-02
DS-2879/XmlWorkflow: keep the old behavior as default

pnbecker:
This PR harmonizes the default behavior of the old workflow system and
XMLWorkflow system regarding whether notifications for tasks returned
to the task pool should be send or not.


mwoodiupui:
+1 by inspection.  Very simple:  only changes a configuration default value from _false_ to _true_.

mwoodiupui:
See #1201

PR #1264, opened 2016-02-02
DS-3016 upgrade dependencies before 6.0 release (2)

helix84:
https://jira.duraspace.org/browse/DS-3025

Supersedes #1260. Only upgrades dependencies which don't cause build failures.

### DIFF
--- a/dspace-api/pom.xml
+++ b/dspace-api/pom.xml
@@ -493,7 +493,7 @@
         <dependency>
             <groupId>com.maxmind.geoip</groupId>
             <artifactId>geoip-api</artifactId>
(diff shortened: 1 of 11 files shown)


helix84:
Passes smoke test - browsing, viewing, searching in XMLUI and JSPUI, couple of endpoints in REST and in OAI.

pnbecker:
+1 by inspection. Travis tells us, that it is building and helix84 tested it. I think it is important to merge before testathon!

hardyoyo:
+1 by inspection

helix84:
Discussed in today's DevMtg.

PR #1268, opened 2016-02-03
DS-2987 Solved metadata loss by qualdrop fields in DescribeStep

juanmanuelcata:
https://jira.duraspace.org/browse/DS-2987
I've modified DescribeStep and DCInputSet to avoid metadata loss
in type-binded qualdrop fields


KevinVdV:
Tested this PR & it works, so +1 from me.

Thanks @juanmanuelcata !

PR #1270, opened 2016-02-03
DS-3030 operations with bitstream policies fail in JSPUI

PhilipVis:
Updated class EPersonDAOImpl to prevent a list containing a 'null' object from being created. 

Updated class AuthorizeAdminServlet to get the item ID from the request, because policy.getdSpaceObject() returns a bitstream when deleting a bitstream policy.

jira ticket:
https://jira.duraspace.org/browse/DS-3030


helix84:
+1, all three operations now work

PR #1272, opened 2016-02-03
DS-2619 add DSpace version to /rest/status

helix84:
https://jira.duraspace.org/browse/DS-2619

Poor man's implementation of versioning - allow querying for API version (DSpace major version).

Discussed and approved in today's DevMtg: http://irclogs.duraspace.org/index.php?date=2016-02-03


helix84:
Replaced by #1281

PR #1276, opened 2016-02-04
DS-2965 - Allow cache clear during re-index (V2)

terrywbrady:
This replaces https://github.com/DSpace/DSpace/pull/1231.  The branch that was used for the prior PR is corrupted.

I am testing in a repository with 250,000+ items.

Without this fix, I am unable to re-index a repository.  With this fix, index-discovery succeeds.

@KevinVdV , I applied the suggestion that you made in the prior version of this PR.


KevinVdV:
+1, merging

Thanks for the work @terrywbrady, I will look around & attempt to find other places in the code where this function could be useful (filter media ?)

PR #1286, opened 2016-02-08
DS-3050 XOAI wrong URL encoding

cjuergen:
https://jira.duraspace.org/browse/DS-3050

use `org.dspace.app.util.Util` instead of `java.net.URLEncoder` for URL encoding in XOAI


helix84:
+1, I confirmed that it works

PR #1293, opened 2016-02-12
[DS-3046] NPE upon first LDAP login with autoregistration

KevinVdV:
https://jira.duraspace.org/browse/DS-3046

By not initializing our authentication methods when the "AuthenticationServiceImpl" bean is created will ensure that the service factories can return the service objects required by the authentication methods.
This will fix the issue for LDAP but also for other authentication methods that might require a service.

The root of the problem is that services are not returned by the factory since the kernel hasn't started up yet.


helix84:
+1, tested

KevinVdV:
@terrywbrady Since I assume shibboleth might have the same issues could you test & see if this also fixes issues that you might be having ?

PR #1295, opened 2016-02-12
Full config migration

tim-atmire:
Quick fix for the OAI settings.


tim-atmire:
Will just do the entirety of OAI while I'm at it.

PR #1300, opened 2016-02-15
DS-2517 replace erroneous sql column with correct column

DylanMeeus:
https://jira.duraspace.org/browse/DS-2517


helix84:
+1, obvious fix

DylanMeeus:
@cjuergen You are correct about that. A resource_id is not unique an it seems that the query does not provide sufficient checks to guarantee that it would return an item. Thanks :)

helix84:
@cjuergen Thanks for catching that, obviously I overlooked it. I'll fix it in the master branch.

PR #1307, opened 2016-02-17
DS-3020 upgrade httpclient

helix84:
https://jira.duraspace.org/browse/DS-3020


terrywbrady:
+1 Maven build successful.  Are there any specific tests I should run after re-build?

kshepherd:
Travis CI tests now pass fine: https://travis-ci.org/DSpace/DSpace/builds/109823856

+1, no build or runtime problems that I can see. I will merge now.

PR #1308, opened 2016-02-17
DS-3067 upgrade Maven plugins

helix84:
https://jira.duraspace.org/browse/DS-3067


terrywbrady:
+1 Maven build successful.  Are there any specific tests I should run after re-build?

kshepherd:
+1. builds fine, quick application tests all pass too. will merge now.

PR #1315, opened 2016-02-23
DS-2941 PasswordAuthentication fix

DylanMeeus:
A fix for this JIRA ticket (in the latest version of DSpace but if correct I will apply it in previous versions of DSpace as well)

https://jira.duraspace.org/browse/DS-2941


DylanMeeus:
Saw a mistake, will create new PR soon

PR #1322, opened 2016-02-26
Feature/DS-3004 isAdmin performance problems

tomdesair:
This pull request fixes the performance issues mentioned in https://jira.duraspace.org/browse/DS-3004

Please check the comments of the JIRA ticket for an explanation of the changes.

### DIFF
--- a/dspace-api/src/main/java/org/dspace/app/util/MetadataExposureServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/app/util/MetadataExposureServiceImpl.java
@@ -7,19 +7,15 @@
  */
 package org.dspace.app.util;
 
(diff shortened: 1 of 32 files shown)


mwoodiupui:
Have not tried it yet, but I did read it and this looks good.

marsaoua:
I have tested this PR for the DS-3096 purposes. It resolves it, therefore the DS-3096 issue should be closed.

hardyoyo:
I have tested this in Vagrant-DSpace. Without this patch applied, searching while logged in (using a single word search), takes about a minute to complete. With this patch applied, searching while logged in takes 1-4 seconds. This is an *excellent* fix, +1

PR #1324, opened 2016-02-29
Correct Sherpa/Romeo ungraded journal (gray) error message

AmberPoo1:
https://jira.duraspace.org/browse/DS-3085

Resolved: DSpace DS-3085 (Ticket in the Space Issue Tracker),
Sherpa/Romeo ungraded journal (gray) shows error


helix84:
+1 tested

PR #1335, opened 2016-03-15
Bump jQuery version from 1.6.2 to 1.6.4 in XMLUI themes, excluding Mirage 2

isido:
Ref: https://jira.duraspace.org/browse/DS-3099

Here's a patch that bumps jQuery version from 1.6.2 to 1.6.4 in Classic, Mirage, Reference and template themes. I've briefly tested all the affected themes, and they seem to work - although I don't use those themes so I might have missed something. I've had the similar fix in my own custom theme, which seems to be working as intended.


terrywbrady:
I took a quick look at this PR and it raised a couple questions for me.  1. Do we have a dependency that requires us to stick with jQuery 1.6.x.  If so, what is it?  Is there something about the DSpace install base that requires jQuery to be distributed with DSpace rather than from a hosted site?

isido:
The initial reason for this patch was to silence our security scanner, which identified a potential issue with jQuery 1.6.2. (https://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2011-4969) I didn't explore the issue further to see if it is actually exploitable, but made a minimal patch to bump the version in our custom theme. This patch is just an expansion of that patch to the shipped themes.

Probably jQuery use and bundling would need larger overhaul (I noticed at least two different versions in shipped and links to cdn-versions as well), but I'm not familiar enough with the UI code and it seem that with the version 7 UI in the horizon, there's not that much incentive to work with old themes.

kshepherd:
+1, no jquery errors caught in my testing. This is a small, low-risk PR which fixes potential security issues in older jquery 1.6.x versions, so I am going to merge it now.
I agree that jquery 1.6.x is pretty old now and that it could do with a major version update at some stage, but if the new UI is coming along with the next DSpace release we might not need to worry about it as an XMLUI dependency.

PR #1336, opened 2016-03-16
DS-3102 Fix collection UUID in DSpaceChoiceLookup

aprilherron:
https://jira.duraspace.org/browse/DS-3102


KevinVdV:
Tested locally & it works, without this pull request you get a javascript exception.

So +1 from me & merging.

PR #1348, opened 2016-04-09
DS-3122: Make DataCite XML namespace configurable.

pnbecker:
https://jira.duraspace.org/browse/DS-3122

This must be merged after #1197 as it already uses the schema version 3.

I don't think this really needs to be tested as the code change are so small an obvious. Just in case a short description on how to test this: take the XSLT from #1197, configure DSpace to mint DOIs with DataCite (see DSpace documentation), publish an Item and run [dspace-install]/bin/dspace doi-organiser -u. If it tells you that it successfully updated a DOI everything is fine.


hardyoyo:
+1 by inspection

pnbecker:
rebased.

mwoodiupui:
+1 by inspection, and it *is* small and obvious.

PR #1349, opened 2016-04-12
[DS-3124] XMLUI: Let users add language tags in sumission's edit meta…

mateuszneumann:
https://jira.duraspace.org/browse/DS-3124

This PR adds XMLUI functionality that is already implemented in JSPUI ([DS-2888](https://jira.duraspace.org/browse/DS-2888), #1166)


tdonohue:
Linking this up with another PR that implements the same feature : #1365

isido:
I made a rebased version of this PR and tested it. Apparently, I need to modify XMLUI themes to see this in action?

mateuszneumann:
AFAIK that should not be necessary.  However you will need to configure those fields in input-forms.xml

isido:
I tested this briefly, and it seems to work as intended. There is a small cosmetic detail in presenting the language field: here is a field with content "Other language" and metadatalanguage "(Other)", which is presented like this:

Other language, ((Other))

I'd probably drop the comma between field and metadatavalue and perhaps the parentheses. PR #1365 deals with this by showing the value as a drop-down menu.

In Mirage 2, the language selection menu gets equal width with the input field. I didn't check where the style for that gets generated.

screenshot from 2016-09-12 11-43-47

tdonohue:

PR #1351, opened 2016-04-14
Fix [DS-3107] - add all news locales files added in the local.cfg.

ghost:
A fix to the bug [DS-3107]
https://jira.duraspace.org/projects/DS/issues/DS-3107?filter=allopenissues

### DIFF
--- a/dspace-api/src/main/java/org/dspace/core/NewsServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/core/NewsServiceImpl.java
@@ -16,9 +16,11 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.OutputStreamWriter;
 import java.io.PrintWriter;
+import java.util.ArrayList;
 import java.util.List;
 
 import org.dspace.core.service.NewsService;
+import org.dspace.services.ConfigurationService;
 import org.dspace.services.factory.DSpaceServicesFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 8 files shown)


pnbecker:
I made a code review based on @LuizClaudioSantos comment in the ticket. I just have two simple suggestions: don't make the method private and don't rely that we have one and only one dot in the filename. I wrote some code that needs to be tested in Jira.

Ones those two issues are resolved I would test this and I'm confident we could merge this soon.

dspace-api/src/main/java/org/dspace/core/NewsServiceImpl.java
```diff
@@ -35,8 +37,23 @@
 	private List<String> acceptableFilenames;
 	
 	public void setAcceptableFilenames(List<String> acceptableFilenames) {
-		this.acceptableFilenames = acceptableFilenames;
-	}
+        this.acceptableFilenames = addLocalesToAcceptableFilenames(acceptableFilenames);
+    }
+
+    private List<String> addLocalesToAcceptableFilenames(List<String> acceptableFilenames){
+        ConfigurationService configurationService =  DSpaceServicesFactory.getInstance().getConfigurationService();
```
KevinVdV: The NewsServiceImpl is a bean you can autowire the configuration service in (you don't have to use the factory to retrieve an instance.

An example of autowiring: https://github.com/DSpace/DSpace/blob/master/dspace-api/src/main/java/org/dspace/content/ItemServiceImpl.java#L55

LuizClaudioSantos:
@pnbecker I changed the access modifier to protected and fix the issue about to use the last point.

@KevinVdV I'm using dependency injection now, instead create a new instance of ConfigurationService myself.

One, question unit test, should I make a unit test for it?

pnbecker:
@LuizClaudioSantos You added two "merge" commits. Beside those the code looks fine now to me and does what I do expected (reviewed + tested).

Can you please remove the two merge commits (by resetting your git repository to the commit before the mergers and pushing this to github using push -f). Then I would be +1 on this.

tdonohue:
+1, code looks good to me. I agree there's a few messy commits here, but the end result looks good.

pnbecker:
As this is +2 we could merge it. I'm closing this one in preference for #1363 which contains the same commits except the two empty merge commits.

LuizClaudioSantos:
Hi @pnbecker,
I need help here, should I delete the blanks commits or not?

PR #1352, opened 2016-04-15
[DS-3125] Submitters cannot delete bistreams of workspaceitems

KevinVdV:
https://jira.duraspace.org/browse/DS-3125

This pull requests make some changes to the authorization system of workspace items. When you delete an Item/Bundle/Bitstream a check is made on the DELETE rights.

In previous DSpace versions this didn't exists because you couldn't call the "delete()" method directly, you had to call the collection.removeItem(), item.removeBundle(), ....

When a new workspace item is created DELETE rights are granted to delete the item, when created the bundle, bitstreams will inherit this from Item.

An SQL script is included (for postgres & H2, oracle to follow later) that will grant DELETE rights to anybody who already has remove rights on these records (since anybody who had remove rights in the past could delete the objects).

Below you can find a small test plan to make sure we test everything related to these changes.

**Test plan**
- [x] Delete bundle/bitstream as submitter
- [x] Delete workspace item as a submitter
- [x] Delete bundle/bitstream as a reviewer in the basic workflow
- [x] Delete bundle/bitstream as a reviewer in the configurable workflow

### DIFF
--- a/dspace-api/src/main/java/org/dspace/content/BitstreamServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/content/BitstreamServiceImpl.java
@@ -238,7 +238,7 @@ public class BitstreamServiceImpl extends DSpaceObjectServiceImpl<Bitstream> imp
     @Override
     public void delete(Context context, Bitstream bitstream) throws SQLException, AuthorizeException {
 
-        // changed to a check on remove
+        // changed to a check on delete
         // Check authorisation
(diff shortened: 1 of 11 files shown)


KevinVdV:
Added oracle database script. Could somebody with oracle take this for a test run to ensure it works properly ? (I just verified the query).

hardyoyo:
I have tested with both PostreSQL and Oracle, and this patch works. +1

pnbecker:
+1 by code review.

PR #1370, opened 2016-04-27
DS-3144 Update of publication-lookup config

jonas-atmire:
For PR https://jira.duraspace.org/browse/DS-3144
Fix for a forgotten update of a config
(as required for https://github.com/DSpace/DSpace/pull/1104)

(The retrieval was updated, but the config itself was forgotten)

### DIFF
--- a/dspace/config/modules/publication-lookup.cfg
+++ b/dspace/config/modules/publication-lookup.cfg
@@ -1,1 +1,1 @@
-publication.url = ${publication.url}
+publication-lookup.publication.url = ${publication.url}
--- a/dspace/config/modules/publication-lookup.cfg
+++ b/dspace/config/modules/publication-lookup.cfg
@@ -1,1 +1,0 @@
-publication-lookup.publication.url = ${publication.url}
--- a/dspace-api/src/main/java/org/dspace/submit/step/XMLUIStartSubmissionLookupStep.java
+++ b/dspace-api/src/main/java/org/dspace/submit/step/XMLUIStartSubmissionLookupStep.java
@@ -77,7 +77,7 @@ public class XMLUIStartSubmissionLookupStep extends AbstractProcessingStep {
 
     public String getPublicationUrl(){
         if(publicationUrl==null){
-            publicationUrl=configurationService.getProperty("publication-lookup.publication.url");
+            publicationUrl=configurationService.getProperty("publication.url","*");
         }
         return publicationUrl;
     }
--- a/dspace-api/src/main/resources/spring/spring-dspace-addon-import-services.xml
+++ b/dspace-api/src/main/resources/spring/spring-dspace-addon-import-services.xml
@@ -24,6 +24,14 @@
 
     </bean>
 
+    <!--If multiple importServices have been configured here but only one is to be used during the lookup step (StartSubmissionLookupStep),
+        this can be accomplished by specifying the property "publication.url" to the baseAddress of the required importService
+        So for example
+        publication.url=http://eutils.ncbi.nlm.nih.gov/entrez/eutils/
+        Will result in using the PubmedImportService for the lookup step
+        Omitting this property will default to searching over all configured ImportService implementations
+    -->
+
     <bean id="PubmedImportService" class="org.dspace.importer.external.pubmed.service.other.ImportMetadataSourceServiceImpl" scope="singleton">
         <property name="metadataFieldMapping" ref="PubmedMetadataFieldMapping"/>
         <property name="baseAddress" value="http://eutils.ncbi.nlm.nih.gov/entrez/eutils/"/>
--- a/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/submission/submit/JSONLookupSearcher.java
+++ b/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/submission/submit/JSONLookupSearcher.java
@@ -163,7 +163,7 @@ public class JSONLookupSearcher extends AbstractGenerator {
 
     public String getLookupURI() {
         if(lookupURI ==null){
-            lookupURI = DSpaceServicesFactory.getInstance().getConfigurationService().getProperty("publication-lookup.publication.url");
+            lookupURI = DSpaceServicesFactory.getInstance().getConfigurationService().getProperty("publication.url","*");
         }
         return lookupURI;
     }
--- a/dspace-xmlui/src/main/java/org/dspace/app/xmlui/cocoon/ThemeResourceReader.java
+++ b/dspace-xmlui/src/main/java/org/dspace/app/xmlui/cocoon/ThemeResourceReader.java
@@ -30,7 +30,7 @@ public class ThemeResourceReader extends SafeResourceReader
         implements CacheableProcessingComponent, Configurable
 {
     // Default whitelist of file extensions that are allowed in an XMLUI theme
-    protected String[] DEFAULT_WHITELIST = new String[]{"css", "js", "json", "gif", "jpg", "jpeg", "png", "ico", "bmp", "htm", "html", "svg", "ttf", "woff"};
+    protected String[] DEFAULT_WHITELIST = new String[]{"css", "js", "json", "gif", "jpg", "jpeg", "png", "ico", "bmp", "htm", "html", "svg", "ttf", "woff", "hbs"};
     protected ConfigurationService configurationService = DSpaceServicesFactory.getInstance().getConfigurationService();
 
     @Override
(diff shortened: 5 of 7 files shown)


dspace/config/modules/publication-lookup.cfg
```diff
@@ -1 +1 @@
-publication.url = ${publication.url}
\ No newline at end of file
+publication-lookup.publication.url = ${publication.url}
```
tdonohue: Is there a reason that this defaults to a non-existent setting? `publication.url` doesn't exist in any other configuration file, so this means the default value will always be null.  Is there a more appropriate default?

I guess I'm wondering whether this should be set to `*` by default (which I think is the recommended setting?), instead of requiring users to add a new `publication.url` field elsewhere.

jonas-atmire: @tdonohue Sorry for the delay in replying.

The reason for this property is the ability to have it configurable which "importSource" to use (and thus base the retrieved results on).
With the `*` configured all configured "ImportService" are used. (Upon searching records, all results are concatenated) But if we want to focus on one specific Service, it's possible to specify a publication.url (which is the base-address that is configured in spring-dspace-addon-import-services.xml), and only the results of that ImportService are retrieved.

For example, here we have only 1 ImportService subclass specified, so it doesn't matter if the actual base-address is filled in, or simply the asterisk `*`. Since it will return the same results.
https://github.com/DSpace/DSpace/blob/master/dspace-api/src/main/resources/spring/spring-dspace-addon-import-services.xml#L27

But if a second config is added, it could be useful to have an option to only retrieve from a certain source, or from all of them.

I do however, agree that the default could be set to the `*`, but still retaining the possibility to specify the ImportServices for aforementioned reasons.

jonas-atmire:
I've updated this PR to remove the extra config file with the single property.
Defaulted this to an asterisk, retaining the possibility to override it through the local.cfg using
publication.url
(Removed the publication-lookup prefix that was added because of the module name)

aschweer:
I've tried this out, but either I'm still doing something wrong or it's still not working.

Firefox web console:
1370-pubmed-errors

When I click on the first error ("junk after document element"), it jumps to view-source:http://localhost:8080/themes/Mirage/templates/publication_records.hbs

When I click on the second error ("SyntaxError..."), it jumps to the very beginning of submission-lookup.js.

I have no time to look into this but I thought I'd mention it here in case it helps someone figure out what's going on. Maybe the licence headers are throwing something off?

aschweer:
I've had another quick look and it appears the configuration for the lookup is working fine. Just something in the lookup isn't. Apologies if this is already being discussed elsewhere.

The message I encounter looks suspiciously like this:
http://stackoverflow.com/questions/6084978/firefox-ajax-junk-after-document-element

I tried (a) removing the licence header and (b) putting a common root div around the two top-level elements in publication_records.hbs and that then gives a "not well-formed" message. Anyway, I can now see that the look-up response actually gives a 500 status with the following message:

````
org.apache.axiom.om.OMException: Cannot create OMDocType because the XMLStreamReader doesn't support the DTDReader extension
	at org.apache.axiom.om.impl.builder.StAXOMBuilder.createDTD(StAXOMBuilder.java:480)
	at org.apache.axiom.om.impl.builder.StAXOMBuilder.next(StAXOMBuilder.java:245)
	at org.apache.axiom.core.CoreParentNodeSupport.ajc$interMethod$org_apache_axiom_core_CoreParentNodeSupport$org_apache_axiom_core_CoreParentNode$buildNext(CoreParentNodeSupport.aj:96)
	at org.apache.axiom.om.impl.llom.OMDocumentImpl.buildNext(OMDocumentImpl.java:1)
	at org.apache.axiom.core.CoreParentNodeSupport.ajc$interMethodDispatch1$org_apache_axiom_core_CoreParentNodeSupport$org_apache_axiom_core_CoreParentNode$buildNext(CoreParentNodeSupport.aj)
	at org.apache.axiom.core.CoreChildNodeSupport.ajc$interMethod$org_apache_axiom_core_CoreChildNodeSupport$org_apache_axiom_core_CoreChildNode$coreGetNextSibling(CoreChildNodeSupport.aj:115)
	at org.apache.axiom.om.impl.llom.OMNodeImpl.coreGetNextSibling(OMNodeImpl.java:1)
	at org.apache.axiom.core.CoreChildNodeSupport.ajc$interMethodDispatch1$org_apache_axiom_core_CoreChildNodeSupport$org_apache_axiom_core_CoreChildNode$coreGetNextSibling(CoreChildNodeSupport.aj)
	at org.apache.axiom.core.CoreDocumentSupport.ajc$interMethod$org_apache_axiom_core_CoreDocumentSupport$org_apache_axiom_core_CoreDocument$coreGetDocumentElement(CoreDocumentSupport.aj:47)
	at org.apache.axiom.om.impl.llom.OMDocumentImpl.coreGetDocumentElement(OMDocumentImpl.java:1)
	at org.apache.axiom.om.impl.common.AxiomDocumentSupport.ajc$interMethod$org_apache_axiom_om_impl_common_AxiomDocumentSupport$org_apache_axiom_om_impl_intf_AxiomDocument$getOMDocumentElement(AxiomDocumentSupport.aj:32)
	at org.apache.axiom.om.impl.llom.OMDocumentImpl.getOMDocumentElement(OMDocumentImpl.java:1)
	at org.apache.axiom.om.impl.builder.StAXOMBuilder.getDocumentElement(StAXOMBuilder.java:554)
	at org.apache.axiom.om.impl.builder.StAXOMBuilder.getDocumentElement(StAXOMBuilder.java:550)
	at org.dspace.importer.external.pubmed.service.other.ImportMetadataSourceServiceImpl.getSingleElementValue(ImportMetadataSourceServiceImpl.java:135)
	at org.dspace.importer.external.pubmed.service.other.ImportMetadataSourceServiceImpl.access$600(ImportMetadataSourceServiceImpl.java:36)
	at org.dspace.importer.external.pubmed.service.other.ImportMetadataSourceServiceImpl$GetNbRecords.call(ImportMetadataSourceServiceImpl.java:126)
	at org.dspace.importer.external.pubmed.service.other.ImportMetadataSourceServiceImpl$GetNbRecords.call(ImportMetadataSourceServiceImpl.java:101)
	at org.dspace.importer.external.service.other.MetadataSource.retry(MetadataSource.java:119)
	at org.dspace.importer.external.pubmed.service.other.ImportMetadataSourceServiceImpl.getNbRecords(ImportMetadataSourceServiceImpl.java:44)
	at org.dspace.importer.external.service.ImportService.getNbRecords(ImportService.java:94)
	at org.dspace.app.xmlui.aspect.submission.submit.JSONLookupSearcher.generate(JSONLookupSearcher.java:85)
````

This is Tomcat 8, Java 1.8.0_91 (OpenJDK), running on Ubuntu -- just in case it's a problem with the environment.

hardyoyo:
> This is Tomcat 8, Java 1.8.0_91 (OpenJDK), running on Ubuntu -- just in case it's a problem with the environment.

I don't think DSpace has been officially approved for use with Java 8 / Tomcat 8, yet, however [DS-2653](https://jira.duraspace.org/browse/DS-2653) seems to indicate that everything _should_ work.

aschweer:
Thanks @hardyoyo , I hadn't realised that -- I've been running my dev instances of DSpace 5.x in that same environment for months without problems. I'll see if I can find some time to re-test on Java 7 / Tomcat 7, but I think in fact I don't even have Tomcat 7 installed on my machine!

jonas-atmire:
@aschweer I couldn't reproduce the second error that is visible in your screenshot with any combination of java 1.7/1.8 + tomcat 7/8 + mirage 1/2. 

I personally haven't tested the OpenJDK versions of java, but I don't think these could be the cause of the problem.

I did however reproduce the "Junk after document element" error, but this does not hinder the retrieval of the records as I can still import the searched publications and go on with the submission. I will look into why the error message is present though. I suspect it is a firefox only thing because the browser tries to parse the .hbs file as xml or html and this is of course not a "valid" file. 

Can you actually not retrieve any records or are you just seeing the errors in the firefox console?

jonas-atmire:
@aschweer 
I've found the source of the first error (junk after document).
The datatype in the ajax call was text, but the actual mimetype defaulted to the XHR mimetype.
I've updated it to use plan text, so firefox doesn't try to parse a non-xml using xml parsers.

Also i've rebased on the latest master to include the .hbs files in the default_whitelist of file extensions allowed in xmlui.

I've you've got the time, could you verify that this first error is not present anymore and check if the second error is still popping up or not?

tdonohue:
Closing this PR in favor of #1462 (which includes some of the bug fixes here).

After analyzing this PR further, I do not like that this removes the `publication-lookup.cfg` file.  It doesn't follow our normal policies of have default settings visible someone in actual cfg files (even if they are commented out).  So, in #1462, I kept the `publication-lookup.cfg` and I renamed the setting there to be `publication-lookup.url` (and added some descriptive comments to that file explaining how to use it).

So, this PR is now replaced by #1462

PR #1378, opened 2016-04-29
[DS-3184] Adding a new bitstream format fails - JSPUI Test Plan Ref ADM18

mwoodiupui:
https://jira.duraspace.org/browse/DS-3184

The actual problem was unconditional dereference of the null list of extensions in a brand-new BitstreamFormat.

I got a bit carried away with tidying up the code, so this one may be easier to review one commit at a time.

### DIFF
--- a/dspace-jspui/src/main/webapp/dspace-admin/list-formats.jsp
+++ b/dspace-jspui/src/main/webapp/dspace-admin/list-formats.jsp
@@ -84,16 +84,19 @@
     for (BitstreamFormat format : formats)
     {
         List<String> extensions = format.getExtensions();
-        StringBuilder extValue = new StringBuilder(256);
-
-        for (String extension : extensions)
+        StringBuilder extValueBuilder = new StringBuilder(256);
+        if (null != extensions)
         {
-            if (extValue.length() > 0)
+            for (String extension : extensions)
             {
-                extValue.append(", ");
+                if (extValueBuilder.length() > 0)
+                {
+                    extValueBuilder.append(", ");
+                }
+                extValueBuilder.append(extension);
             }
-            extValue.append(extension);
         }
+        String extValue = extValueBuilder.toString();
 %>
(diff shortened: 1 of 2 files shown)


mwoodiupui:
The additional commit is tested as before:  create a new format and note that it does not fail.  Updating and deletion also tested.

dspace-jspui/src/main/webapp/dspace-admin/list-formats.jsp
```diff
@@ -30,14 +29,16 @@
...
 <%@ page import="java.util.List" %>
+<%@ page import="java.lang.StringBuilder" %>
 <%@ page import="org.dspace.content.service.BitstreamFormatService" %>
 <%@ page import="org.dspace.content.factory.ContentServiceFactory" %>
 
 
 <%
-    BitstreamFormatService bitstreamFormatService = ContentServiceFactory.getInstance().getBitstreamFormatService();
-    List<BitstreamFormat> formats =
-        (List<BitstreamFormat>) request.getAttribute("formats");
```
pnbecker: This should not stop us from merging, but the attribute "formats" is still set: https://github.com/DSpace/DSpace/blob/b6875cddaa42602bfa61c2521c57979b5c640159/dspace-jspui/src/main/java/org/dspace/app/webui/servlet/admin/BitstreamFormatRegistry.java#L171
Perhaps we should change the BitstreamFormatRegistry and also the other JSP (see https://github.com/DSpace/DSpace/blob/b6875cddaa42602bfa61c2521c57979b5c640159/dspace-jspui/src/main/java/org/dspace/app/webui/servlet/admin/BitstreamFormatRegistry.java#L136) as well?

pnbecker:
inspected and tested. We should remove one line in BitstreamFormatRegistry (see above). Beside that I'm +1.

hardyoyo:
+1 by inspection

PR #1387, opened 2016-05-04
Ds 2962 Robots.txt disallows site /discovery but not /handle/*/disover

PhilipVis:
Implemented a transformer that adds the meta tag described at http://www.robotstxt.org/meta.html to the pagemeta. Also updated themes Mirage and Mirage2 to add this meta tag to the page. 

The transformer is configured to replace the dynamic URLs from robots.txt.

jira ticket: https://jira.duraspace.org/browse/DS-2962?jql=text%20~%20%22robots.txt%22


hardyoyo:
+1 by inspection

terrywbrady:
+1, installed locally, confirmed the presence of <meta name="ROBOTS" content="NOINDEX, FOLLOW"> after navigating a facet

PR #1388, opened 2016-05-04
[DS-3003] Make the constructors of 2 services protected instead of private

KevinVdV:
Follow up for: https://jira.duraspace.org/browse/DS-3003

Make the constructors of ChoiceAuthorityServiceImpl & LegacyPluginServiceImpl protected instead of private because else we cannot override the behavior of these services.


pnbecker:
obvious code change. +1 by inspection (I don't think that is actually is testable).

In case someone else think about the other constructors changed by DS-3003, they are changed to be protected not private.

hardyoyo:
+1 by inspection, very tiny bug fix, let's merge

PR #1390, opened 2016-05-04
[DS-3171] Subscription list not shown - JSPUI Test Plan Ref EP4

mwoodiupui:
https://jira.duraspace.org/browse/DS-3171

Plus the usual folding, whitespace cleanup.


hardyoyo:
+1 by inspection

pnbecker:
inspected and tested. +1

PR #1392, opened 2016-05-09
DS-3206 Policy form merge field values when perform group search

oooriii:
This branch fixes the DS-3206 bug

https://jira.duraspace.org/browse/DS-3206


helix84:
+1, tested

Merging obvious bug fix.

PR #1407, opened 2016-05-18
DS-3150

PhilipVis:
request-form.jsp was expecting a string for bitstream_id while it received a UUID.

jira ticket:
https://jira.duraspace.org/browse/DS-3150


pnbecker:
My colleague run into this problem. I filed a patch and looked out for a ticket. Found DS-3150 and this PR which is exactly the same as what I wrote. => +1 :-)

pnbecker:
I'm going to merge this, as the bug is obvious and the solution simple. We need to move things forward for DSpace 6.0!

I tested it and together with #1393 it makes the request copy feature functional again.

PR #1414, opened 2016-05-20
DS-2766 /rest/collections/xxx?expand=parentCommunityList only returns 1 parent community

terrywbrady:
See https://jira.duraspace.org/browse/DS-2766 for the assumptions surrounding this change.

### DIFF
--- a/dspace-api/src/main/java/org/dspace/content/Collection.java
+++ b/dspace-api/src/main/java/org/dspace/content/Collection.java
@@ -259,7 +259,7 @@ public class Collection extends DSpaceObject implements DSpaceObjectLegacySuppor
      * Get the communities this collection appears in
      * 
      * See DS-2766 for some documented assumptions about collection parents.
-     * In DSpace6x, the assumption is that a collection exists in only one community/subcommunity.
+     * In DSpace6x, the assumption is that a collection exists in only one community/subcommunity. 
      *
      * @return array of <code>Community</code> objects
      * @throws SQLException if database error
--- a/dspace-api/src/main/java/org/dspace/content/Collection.java
+++ b/dspace-api/src/main/java/org/dspace/content/Collection.java
@@ -257,6 +257,9 @@ public class Collection extends DSpaceObject implements DSpaceObjectLegacySuppor
 
     /**
      * Get the communities this collection appears in
+     * 
+     * See DS-2766 for some documented assumptions about collection parents.
+     * In DSpace6x, the assumption is that a collection exists in only one community/subcommunity.
      *
      * @return array of <code>Community</code> objects
      * @throws SQLException if database error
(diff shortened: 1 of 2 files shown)


terrywbrady:
This repo has been activated with travis.  Need to force the build to start.

terrywbrady:
[12:27:19]	‹tdonohue›  not sure I understand this code change. It looks like a 'for' loop in a 'for' loop that might only find parent Communities that are two-levels deep? Or am I misunderstanding it
[12:27:38]	‹terry-b3›  It is going up the ancestor tree
[12:28:02]	‹terry-b3›  To return a full breadcrumb trail for a collection
[12:28:56]	‹tdonohue›  Oh, I see now
[12:29:30]	‹terry-b3›  The comment in the org.dspace.content file was to summarize our discussion about multiple parent communities for a collection
[12:31:02]	‹tdonohue›  yea, I think the comment seems out of place which confused me. I feel like it's more about the fact that "collection.getCommunities()" should only return *one* parent
[12:31:32]	‹hpottinger›  sorry to interrupt, but I'm pretty sure I just messed up the dspace-replicate release I was trying to "wing" just now
[12:31:39]	‹terry-b3›  I can take another pass at the comment block.
[12:32:07]	‹terry-b3›  Your question makes me want to re-verify that Community.getParentCommunities() returns a list.
[12:32:57]	‹tdonohue›  terry-b3: it's not a huge deal... I think it likely belongs more around the entire *if* regarding the "parentCommunityList" field. As DS-2766 (referred in the comment) is more about that, than about the inner 'for' loop
[12:32:58]	‹kompewter›  [ https://jira.duraspace.org/browse/DS-2766 ] - [DS-2766] /rest/collections/xxx?expand=parentCommunityList only returns 1 parent community - DuraSpace JIRA
[12:33:31]	‹tdonohue›  terry-b3: in any case, the code seems right. It's just the comment might be better moved up, as it's about the meaning of "parentCommunityList"

tdonohue:
+1 via code review. Needs a test.  Minor notes on comment location (as pasted above)

terrywbrady:
## Testing instructions on demo.dspace.org

* Create a community
* Create a subcommunity of the community
* Create a collection within the subcommunity

Note the following collection has 2 community parents: http://demo.dspace.org/xmlui/handle/10673/539

image

If you look at in the REST api, currently only one parent community is visible

https://demo.dspace.org/rest/collections/d0e8e9ae-db09-4df2-8967-d5af4d4d6271?expand=parentCommunityList

After merging this PR locally, you should be see a list of community parents.

terrywbrady:
Re-tested and noticed that this is not properly iterating up multiple levels of hierarchy.  I will add a fix.

terrywbrady:
Tested locally and confirmed that this code works with a chain of 1,2,3 or 4 community ancestors.

tmguimaraes:
+1
Tested with up to 5 community ancestos

tdonohue:
As noted in today's meeting, we should see if #1319 affects this bug, as it fixes a lower level issue with gathering hierarchies

terrywbrady:
Withdrawing this PR in favor of https://github.com/DSpace/DSpace/pull/1319

PR #1420, opened 2016-05-25
DS-3152 JSPUI -> Word breaking on long filenames/description

jonas-atmire:
JIRA ticket -> https://jira.duraspace.org/browse/DS-3152

Word-break on long filenames and file descriptions during submission/reviewstep + item display.
Re-usable css class if more similar bugs are found.

PR #1421, opened 2016-05-25
DS-3219 -> Word breaking on long filenames/description (review/upload)

jonas-atmire:
JIRA Ticket : https://jira.duraspace.org/browse/DS-3219
XMLUI counterpart for https://github.com/DSpace/DSpace/pull/1420
Long file name or description breaks file upload / review page (Mirage1+2)

### DIFF
--- a/dspace-xmlui-mirage2/src/main/webapp/styles/shared/_dspace-bootstrap-tweaks.scss
+++ b/dspace-xmlui-mirage2/src/main/webapp/styles/shared/_dspace-bootstrap-tweaks.scss
@@ -140,3 +140,6 @@ span.bold {
 p.pre {
     white-space:pre;
 }
+.break-all{
+  word-break: break-all;
+}
--- a/dspace-xmlui-mirage2/src/main/webapp/styles/shared/_dspace-bootstrap-tweaks.scss
+++ b/dspace-xmlui-mirage2/src/main/webapp/styles/shared/_dspace-bootstrap-tweaks.scss
@@ -142,6 +142,6 @@ p.pre {
 }
 .break-all{
   -ms-word-break:   break-all;
-  word-break:       break-word;
+  word-break:       break-all;
   word-wrap:        break-word;
 }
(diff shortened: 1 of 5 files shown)


hardyoyo:
+1 by review

terrywbrady:
+1, tested locally and confirmed fix in XMLUI

tdonohue:
Just noticed that the corresponding JSPUI ticket #1420 has had updates by @PhilipVis to the `break-all` styles to make them compatible to all browsers.

@jonas-atmire , while this sounds like it's mostly ready to go, I'm going to wait to merge until we get the same CSS level fixes in this PR.

Thanks!

jonas-atmire:
@tdonohue 
I was going to do the port but hadn't found the time yet.
This updates both Mirage1+2  to be concise with the JSPUI css fix

dspace-xmlui-mirage2/src/main/webapp/styles/shared/_dspace-bootstrap-tweaks.scss
```diff
@@ -139,4 +139,9 @@ span.bold {
 /* Preserve whitespace. Used to display error stack in a readable way. */
 p.pre {
     white-space:pre;
+}
+.break-all{
+  -ms-word-break:   break-all;
+  word-break:       break-word;
```
pnbecker: While testing #1420 I found that word-break must be set to break-all. (http://www.w3schools.com/cssref/css3_pr_word-break.asp)

terrywbrady:
Confirmed that the latest commit resolved the item submit/review issue in Firefox.

Note that there is an issue when editing an item (exists in Firefox and Chrome)

image

terrywbrady:
Tested in IE and Safari.  The first issue (item submit review) is resolved across browsers.  The edit item issue is present across browsers.

jonas-atmire:
Added the css class to both the overview (EditItemBitstreamForm) and specific edit of the bitstream (EditBitstreamForm) as this was broken as well.

Haven't had time to check for even more places where this could be broken, but it still would require only some minor changes if more pop up

terrywbrady:
Confirmed fix for DS-3223 in the latest commit

PR #1424, opened 2016-05-31
DS-3226: Avoid NPEs for wrong search URIs.

pnbecker:
https://jira.duraspace.org/browse/DS-3226

To test this open a community or collection in JSPUI and add `/simple-search?location=123456789%2F2&query=&rpp=10&sort_by=score&order=desc&filter_field_1=dateIssued&filter_type_1=equals&filter_value_2=2016` to the URL. Without this patch this results in an Null Pointer exception. With this patch this results in a search without any filter.


tmguimaraes:
+1 tested.

aschweer:
+1 tested and behaviour pre/post patch is as described by @pnbecker

PR #1425, opened 2016-05-31
DS-3227: Use cal.add(...) instead of cal.roll(...) in CheksumHistoryDAOImplTest.

pnbecker:
https://jira.duraspace.org/browse/DS-3227


DylanMeeus:
+1 by inspection. Fixes the travis builds and using `.add` makes more sense.
`cal.add(cal.DATE,+3)` will correctly return "june 3", whilst `.roll` would return '3th of may'.

KevinVdV:
+1, merging.

pnbecker:
@DylanMeeus @KevinVdV thanks for the immediate reaction!

PR #1431, opened 2016-06-06
DS-3231 fixed by using cc.license.jurisdiction

tmguimaraes:
Fixed [ DS-3231](https://jira.duraspace.org/projects/DS/issues/DS-3231) by changing webui.submit.cc-jurisdiction to cc.license.jurisdiction

Acording to the [docs](https://wiki.duraspace.org/display/DSDOC6x/Configuration+Reference#ConfigurationReference-ConfiguringCreativeCommonsLicense) cc.license.jurisdiction is a XMLUI property, but the current webui.submit.cc-jurisdiction does not exist. Instead of creating multiple properties for the same thing, i think cc.license.jurisdiction should be reused.


helix84:
+1, obvious fix

PR #1439, opened 2016-06-15
DS-3244: On error pages, links back to home page broken when xmlui deployed at /

rradillen:
check for existence in leading slash in context path in exception rendering


rradillen:
andrea's is better

PR #1446, opened 2016-06-28
[DS-3217] removed testGetCurrent

hardyoyo:
not worth the trouble, as written it will fail between 23UTC and 24 UTC, possibly through 1 UTC

### DIFF
--- a/dspace-api/src/test/java/org/dspace/content/DCDateTest.java
+++ b/dspace-api/src/test/java/org/dspace/content/DCDateTest.java
@@ -391,16 +391,6 @@ public class DCDateTest
                 equalTo("14-Apr-2010"));
     }
 
-    /**
-     * Test of getCurrent method, of class DCDate.
-     */
-    @Test
-    public void testGetCurrent()
-    {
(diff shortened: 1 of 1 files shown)


hardyoyo:
https://jira.duraspace.org/browse/DS-3217

LuizClaudioSantos:
Hi @hardyoyo 

It more a question than a suggestion, the code of the test should no be like the code below, I mean, you need to set the time to the object Date to compare?

Best
Luiz

`
@Test		
 public void testGetCurrent()		
  {		
        Calendar calendar = Calendar.getInstance();
         calendar.setTimeInMillis(System.currentTimeMillis());		
         assertTrue("testGetCurrent 0", DateUtils.isSameDay(DCDate.getCurrent().toDate(), calendar.getTime()));		
  }


`

hardyoyo:
Alternate PR: https://github.com/DSpace/DSpace/pull/1448

hardyoyo:
Closing this, since #1449 provides fixes for all the similarly-broken tests, no need to remove them

PR #1450, opened 2016-06-29
Solving the issue DS-3193

ghost:
As  I commented in the https://jira.duraspace.org/browse/DS-3193, I fixed the value of retrieveLink in the rest response instead of change the root of the rest application to the UI application.

### DIFF
--- a/dspace-rest/src/main/java/org/dspace/rest/common/Bitstream.java
+++ b/dspace-rest/src/main/java/org/dspace/rest/common/Bitstream.java
@@ -74,7 +74,7 @@ public class Bitstream extends DSpaceObject {
         description = bitstream.getDescription();
         format = bitstreamService.getFormatDescription(context, bitstream);
         sizeBytes = bitstream.getSize();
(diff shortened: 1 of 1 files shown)


KevinVdV:
@LuizClaudioSantos Not sure if the best way forward would be to hardcode the "rest" parameter. If anybody would like to deploy their rest under a different contextPath this would break the code again.

Perhaps try using request.getContextPath(): http://docs.oracle.com/javaee/7/api/javax/servlet/http/HttpServletRequest.html#getContextPath--

LuizClaudioSantos:
@KevinVdV I agree with you, but the HttpServletRequest is not available in Bitstream but in BitstreamResource instead, do you think that should I HttpServletRequest as a dependency of Bitstream?

LuizClaudioSantos:
I add the suggest of @KevinVdV in the [https://github.com/DSpace/DSpace/pull/1460](url), please use that one instead.

tdonohue:
Closing as requested. Replacement PR is at #1460

PR #1451, opened 2016-07-01
DS-3254 Community Admins don't get a "Delete" button on the EditCollection page in XMLUI

dimitrove:
https://jira.duraspace.org/browse/DS-3254

### DIFF
--- a/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/administrative/collection/EditCollectionMetadataForm.java
+++ b/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/administrative/collection/EditCollectionMetadataForm.java
@@ -29,6 +29,7 @@ import org.dspace.authorize.service.AuthorizeService;
 import org.dspace.content.Collection;
 import org.dspace.content.factory.ContentServiceFactory;
 import org.dspace.content.service.CollectionService;
+import org.dspace.core.Constants;
 
 
 /**
@@ -209,8 +210,8 @@ public class EditCollectionMetadataForm extends AbstractDSpaceTransformer
 	    
 		Para buttonList = main.addPara();
 	    buttonList.addButton("submit_save").setValue(T_submit_save);
-        //Only System Admins can Delete Collections
-	    if (authorizeService.isAdmin(context))
+
+	    if (AuthorizeManager.authorizeActionBoolean(context, thisCollection, Constants.DELETE))
         {
 	    	buttonList.addButton("submit_delete").setValue(T_submit_delete);
         }


dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/administrative/collection/EditCollectionMetadataForm.java
```diff
@@ -209,8 +210,8 @@ public void addBody(Body body)
 	    
 		Para buttonList = main.addPara();
 	    buttonList.addButton("submit_save").setValue(T_submit_save);
-        //Only System Admins can Delete Collections
-	    if (authorizeService.isAdmin(context))
+
+	    if (AuthorizeManager.authorizeActionBoolean(context, thisCollection, Constants.DELETE))
```
tdonohue: Hi @dimitrove , thanks for creating your first Pull Request (PR)!

One note here, it looks like Travis CI noticed an error in the code (hence the red x above).  In the DSpace 6 codebase, there's no longer a `AuthorizeManager` class.  It's been replaced by an `AuthorizeService`.

So, in this line of the code, I think you need it to read:
`authorizeService.authorizeActionBoolean(context, thisCollection, Constants.DELETE)`

Beyond that, it looks like you did everything right with this PR!  You should be able to simply add a new commit to your "DS-3254" fixing that line of code, and this PR should update automatically.

dimitrove: Thank you Tim,

I did this.

Further - to resolve the DELETE issue - there are two things:

- one is an obvious fix (like this one)
in
[dspace-src]\dspace-xmlui\src\main\resources\aspects\Administrative\administrative.js:

assertAdministrator(); in doEditCollectionMetadata
vs
assertAuthorized(Constants.COMMUNITY, communityID, Constants.DELETE); in
doEditCommunityMetadata

- the second has an obvious cause, but not obvious (to me) fix. what
happens now is:
--- the collection is detached from the community (where is the
authorization policy)
in
[dspace-src]\dspace-xmlui\dspace-api\src\main\java\org\dspace\content\Community.java:

            // Remove the parent/child mapping with this collection
            // We do this before deletion, so that the deletion doesn't
throw database integrity violations

--- then, during the deletion of the collection, removeTemplateItem() fails
- there is no authorization policy to justify it.

The above was about "delete collection" by Community Admin.
Very similar with my case of "delete subcommunity" by Community Admin.

We have CommunityA - SubcommunityB - CollectionC.
When we delete SubcommunityB, it is first detached from CommunityA (where
is the authorization policy).
Then, when trying to delete CollectionC - there is no authorization policy
to justify it.

May be I should put this in JIRA - to be clear what needs to be fixed?

Best regards
Evgeni

tdonohue:

PR #1457, opened 2016-07-11
Update logo (@mire -> Atmire)

cwilper:
This updates the Mirage/Mirage2 bylines to use the current Atmire logo instead of the old one.

Preview: http://imgur.com/a/5GBJL


helix84:
+1, obviously correct

PR #1460, opened 2016-07-13
[DS-3193] retrieveLink for bitstreams in REST API does not include /rest

ghost:
I think this is the best approach to fix https://jira.duraspace.org/browse/DS-3193 . I used a new instance of DSpace to get the httprequest and so get the path. So now I'm adding the path according to the request as suggested by @KevinVdV.

`String path = new DSpace().getRequestService().getCurrentRequest().getHttpServletRequest().getContextPath();
  retrieveLink = path + "/bitstreams/" + bitstream.getID() + "/retrieve";
`


tdonohue:
@KevinVdV : since you were actively reviewing the previous PR, do you (or someone else at Atmire) have a chance to give this a test. The code looks correct to me, just needs a test.

tdonohue:
+1 just tested this fix and it works.

mwoodiupui:
+1 by inspection.

tdonohue:
Merged by cherry-picking the single commit that was needed here: https://github.com/DSpace/DSpace/pull/1460/commits/89fa4ca0582ca2940f3f2f3207a955b24ebe7218  (See: https://github.com/DSpace/DSpace/commit/053b19b637fc86b611577d450fac7eb3fea26bcb)

This is now fixed on master & will be a part of 6.0.  Thanks @LuizClaudioSantos !

PR #1475, opened 2016-07-23
DS-3267 Add alt getCollectionDefaultRead method w/UUIDs

cwilper:
https://jira.duraspace.org/browse/DS-3267

This fixes the issue with editing collection default read policies by adding an alternative signature for getCollectionDefaultRead, which accepts and returns a UUID.


helix84:
+1, tested

PR #1477, opened 2016-07-24
DS-3276 Fix authority indexing NPE

cwilper:
https://jira.duraspace.org/browse/DS-3276

The problem was, AuthorityValue was never a Spring bean, but when refactored for DSpace 6, it was assumed to be one. Rather than making it and its subclasses' instantiation Spring-managed, this fix just removes the inoperable auto-wired fields, and gets the MetadataValueService on-demand, from ContentServiceFactory.


helix84:
+1, tested

PR #1488, opened 2016-08-11
DS-2948 : Set Solr captureAttr to true when indexing full text

ghost:
https://jira.duraspace.org/browse/DS-2948


abollini:
+1 it looks obvious to me

tdonohue:
+1 via code review. Looks good, may just need a quick verification test.

abollini:
tested. The metadata are now saved in a "meta" field of the solr document. The fulltext field start with lot of newlines but this is due to the SolrCell internal behaviour and as they are just "whitespace" don't affect the hightlighing snippet feature

PR #1502, opened 2016-08-23
[DS-3300] Citation page: add page format selection

isido:
This patch will allow user to select either LETTER (default) or A4 as the citation page size.

### DIFF
--- a/dspace-api/src/main/java/org/dspace/disseminate/CitationDocumentServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/disseminate/CitationDocumentServiceImpl.java
@@ -280,7 +280,24 @@ public class CitationDocumentServiceImpl implements CitationDocumentService, Ini
         try {
             Item item = (Item) bitstreamService.getParentObject(context, bitstream);
             sourceDocument = sourceDocument.load(bitstreamService.retrieve(context, bitstream));
-            PDPage coverPage = new PDPage(PDRectangle.LETTER); // TODO: needs to be configurable
+	    // Page format can be either LETTER (default) or A4
+	    String pageformat_cfg = configurationService.getProperty("citation-page.page_format");
+	    PDRectangle pageformat = PDRectangle.LETTER;
+	    if (pageformat_cfg == null) {
+		pageformat = PDRectangle.LETTER;
+	    } else {
+		String pf = pageformat_cfg.toUpperCase();
+		if (pf.equals("A4")) {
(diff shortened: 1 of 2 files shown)


dspace-api/src/main/java/org/dspace/disseminate/CitationDocumentServiceImpl.java
```diff
@@ -280,7 +280,24 @@ public File makeCitedDocument(Context context, Bitstream bitstream)
...
+	    } else {
+		String pf = pageformat_cfg.toUpperCase();
+		if (pf.equals("A4")) {
+		    pageformat = PDRectangle.A4;
+		} else if (pf.equals("LETTER")) {
+		    pageformat = PDRectangle.LETTER;
+		} else {
+		    log.info("Citation-page: Unknown page format ' " + pageformat_cfg + "', using LETTER.");
+		    pageformat = PDRectangle.LETTER;
+		}
```
kshepherd: Just some minor feedback on this: this if loop could probably be simplified a bit.. since pageformat is instantiated as PDRectangle.LETTER already, the test on pageformat_cfg could be for != null, and then an else block wouldn't be needed.. the final else on the inside test is probably fine as an explicit default and is a good opportunity to log that info message. Also, not an official style rule but DSpace variables tend to be in camelCase, eg pageformatConfig, rathter than pageformat_cfg.
Just feedback and not a blocker, I'm still +1 on the PR anyway :)

kshepherd:
looks good, thanks @isido ! i'll try to give this a test in the next day or two, too as A4 is my default paper size ;)
I notice this PR is against master but has a milestone of 6.3. it would be great to get some more tests/approvals and then port to dspace-6.x branch after merging

The new configuration options and behaviiour will also need documentation

isido:
Thanks for the feedback. I made this patch for a project I was working, before 6.0 was released, and there's a pr for 5-series as well. I'll have to wipe the dust from my development environment, and I'll freshen up the patch.

isido:
I've made a new pr, since it was the fastest way to get checkstyle etc. working. I probably can update this as well, depending on which target, 6 or 7, is better.

kshepherd:
Hi @isido, actually, this PR is fine for master - checkstyle is fine or Travis CI would have complained.
I got mixed up and submitted a master port for this just now, but I'll close that in favour of this PR.
You can see my 6.x port at #2033 (I'll keep that one open)

kshepherd:
Closing in favour of #2034 and #2033

PR #1510, opened 2016-09-07

helix84:


hardyoyo:
Compiles and deploys cleanly in Vagrant-DSpace, filter-media runs without complaint. +1

abollini:
tested as well, it works as expected

PR #1512, opened 2016-09-13

abollini:
Compiles and deploys cleanly. Filter-media runs and extract text from normal pdf files


lap82:
+1 verified

helix84:
+1, tested on 5.x

PR #1507, opened 2016-09-01
[DS-2604] Fix JSPUI Creative Commons license retrieval

aschweer:
The changes in this commit
- allow up to 5 redirects to be followed in retrieving the license
- ensure a user agent is sent when the license is retrieved
- clean up/modernise some parts of the code

PR #1518, opened 2016-09-16
[DS-2604] fix jspui retrieve CC rdf

lap82:
based on https://github.com/DSpace/DSpace/pull/1507


pnbecker:
@lap82 Is this the same as #1522 for 6.0? Why does #1522 contains so many more changed files than this one? Is this PR ready to be tested?

pnbecker:
In the meantime I found #1526. Is #1526 a PR additional to this one or does it replace it?

lap82:
replaced by #1526

PR #1522, opened 2016-09-22
DS-2604 port - DSpace 5.x - from XMLUI to JSPUI the approach to queries Creative Commons service (via REST API)

lap82:
https://jira.duraspace.org/browse/DS-2604

port - for DSpace 5.x - from XMLUI to JSPUI the approach to reach the Creative Commons service (via REST API), bug fixing XMLUI (change license lost the focus of the old selected license, error get property from configuration when no license selected, save the RDF file in the correct form - see https://jira.duraspace.org/browse/DS-3326), adjust JSPUI admin Creative Commons feature


dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/submission/submit/CCLicenseStep.java
```diff
@@ -70,8 +70,8 @@
     protected static final Message T_license = 
         message("xmlui.Submission.submit.CCLicenseStep.license");
         protected static final Message T_submit_remove = message("xmlui.Submission.submit.CCLicenseStep.submit_remove");
-        protected static final Message T_no_license    = message("xmlui.Submission.submit.CCLicenseStep.no_license");
-        protected static final Message T_select_change = message("xmlui.Submission.submit.CCLicenseStep.select_change");
+        protected static final Message T_no_license    = message("webui.Submission.submit.CCLicenseStep.no_license","xmlui.Submission.submit.CCLicenseStep.no_license");
+        protected static final Message T_select_change = message("webui.Submission.submit.CCLicenseStep.select_change","xmlui.Submission.submit.CCLicenseStep.select_change");
```
abollini: no need to changes the i18n key for the xmlui here

dspace/config/dspace.cfg
```diff
@@ -938,6 +938,9 @@ cc.submit.setname = true
 # Store license bitstream (RDF license text) during web submission
 cc.submit.addbitstream = true
 
+# ONLY JSPUI, enable Creative Commons admin 
+webui.submit.enable-cc = false
```
abollini: is it a new feature? or something that was previously missing in the configuration?

lap82: In previous version this configuration was missing

dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/submission/submit/CCLicenseStep.java
lap82: I changed the optionId not the i18n key, to avoid server side to double check for xmlui and jspui for this optionId. In any case in the PR for 6.x I follow your suggestion.

lap82:
@hardyoyo @pnbecker due to your interest into this feature in the last IRC, I ask if you can take a review, please. I have tested both JSPUI and XMLUI and it works!

lap82:
- added suggestion provided by @abollini;
- added selectOption when select_change or no_license dropdown option selected (XMLUI) - used to remove CC 
- added double check - servlet side - to manage no_license dropdown option selected

abollini:
tested on both JSPUI and XMLUI, +1

pnbecker:
I didn't have the for a proper review, but @lap82 is a trusted developer and also @abollini seems to have reviewed the code. Just one idea: This PRs deprecates some class variables. I think it would be great to add a short comment what to use instead than the deprecated stuff.

I tested it on JPSUI and to me it seems to be fine. I don't merge this now, as I would prefer if we could merge it together with the solution for DSpace 6. Nevertheless, I think this one is ready.

lap82:
I'm following suggestion provided by @pnbecker and so I added some comment (I hope without grammatical errors). I'm waiting for Travis green light and I will proceed to the merge.

PR #1528, opened 2016-09-25
[DS-3336] Properly sort collections in move item drop-down

aschweer:


aschweer:
https://jira.duraspace.org/browse/DS-3336

misilot:
I just tried this on my 5.5 install and it does not sort.

I have attached what it looks like after the patch.
capture

misilot:
Tested on 5.5 and it works here. Thanks @aschweer!

tdonohue (approved):
👍 This seems like a small, useful patch. To me it seems like one we should merge to 5.x and cherry-pick over to 6.x

tdonohue:
This was merged, and also cherry-picked (after testing minor updates to code) to 6.x: https://github.com/DSpace/DSpace/commit/97785d778f5c414a4e793009dee246e2cec16197

PR #1535, opened 2016-09-28
DS-3122 Set handle.canonical.prefix to the ${dspace.url}/handle.

pnbecker:
https://jira.duraspace.org/browse/DS-3112

DS-3112 suggest to change our default handle configuration. @bram-atmire created #1344 to resolve it by changing handle.canonical.prefix and handle.prefix. In the PR discussion it became clear that changing handle.prefix has some bigger implications inside and outside of DSpace (we should search code that looks out for 123456789 as handle prefix). Nevertheless the idea of changing handle.canonical.prefix is a good one.
This PR changes handle.canonical.prefix but leaves handle.prefix as it was in previous versions of DSpace. With this configuration valid URLs will be added in dc.identifier.uri and following in the OAI-PMH interface. Beside that it has much less implications as it does not change handle.prefix. I think we should merge this PR before DSpace 6.0 and may discuss later whether we also want to change handle.prefix.

### DIFF
--- a/dspace/config/dspace.cfg
+++ b/dspace/config/dspace.cfg
@@ -225,22 +225,21 @@ identifier.doi.namespaceseparator = dspace/
 
 # Canonical Handle URL prefix
 #
-# By default, DSpace is configured to use http://hdl.handle.net/
-# as the canonical URL prefix when generating dc.identifier.uri
-# during submission, and in the 'identifier' displayed in JSPUI
-# item record pages.
-#
-# If you do not subscribe to CNRI's handle service, you can change this
-# to match the persistent URL service you use, or you can force DSpace
-# to use your site's URL, eg.
-#handle.canonical.prefix = ${dspace.url}/handle/
+# Items in DSpace receive a unique URL, stored in dc.identifier.uri
+# after it is generated during the submission process.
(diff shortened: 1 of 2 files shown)


tdonohue (approved):
+1 This seems like a good change. Just changes the default config for `handle.canonical.prefix` to be `${dspace.url}/handle/`

dspace/config/dspace.cfg
```diff
@@ -225,22 +225,21 @@ identifier.doi.namespaceseparator = dspace/
...
-# By default, DSpace is configured to use http://hdl.handle.net/
-# as the canonical URL prefix when generating dc.identifier.uri
-# during submission, and in the 'identifier' displayed in JSPUI
-# item record pages.
-#
-# If you do not subscribe to CNRI's handle service, you can change this
-# to match the persistent URL service you use, or you can force DSpace
-# to use your site's URL, eg.
-#handle.canonical.prefix = ${dspace.url}/handle/
+# Items in DSpace receive an unique url, stored in dc.identifier.uri
```
mwoodiupui: "a unique URL"

and again in local.cfg.EXAMPLE

Otherwise, +1

PR #1540, opened 2016-09-29
DS-3311: Changing default configuration of healthcheck

pnbecker:
Healthcheck can check for passed Pre-3.0-embargos. These are disabled in DSpace out of the box in favor for Past-3.0-embargos. This PR just changes the configuration of the healthcheck to disable the pre-3.0-embargo check.


tdonohue (approved):
+1 This looks like an obvious change / improvement.

hardyoyo:
+1 by inspection

PR #1545, opened 2016-09-30
DS-3358 corrects display of author name in browsing while working with authority

olli-gold:
https://jira.duraspace.org/browse/DS-3358

In DSpace browsing authors shows the authority key instead of a human readable name, if this person has an authority key. This PR fixes this issue in JSPUI.

### DIFF
--- a/dspace-jspui/src/main/java/org/dspace/app/webui/jsptag/BrowseListTag.java
+++ b/dspace-jspui/src/main/java/org/dspace/app/webui/jsptag/BrowseListTag.java
@@ -532,24 +532,26 @@ public class BrowseListTag extends TagSupport
                             	{
                                     String argument;
                                     String value;
+                                    String authorityArgument = null;
+                                    String authorityValue = null;
                                     if (metadataArray[j].authority != null &&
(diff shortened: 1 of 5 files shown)


pnbecker:
Looks sensible to me. +1 by review, not tested (yet).

@olli-gold: Thanks for the PR! Can you please create an issue in https://jira.duarspace.org about this and link that issue here?

olli-gold:
I have found a bug in Jira, which is obviously exactly the problem I had: https://jira.duraspace.org/browse/DS-2401. This bug has been solved by configuration. I'll try to apply the configuration suggested there, this might make this PR obsolete.

olli-gold:
Obviously the ticket https://jira.duraspace.org/browse/DS-2401 only fixes the issue for xmlui - I have applied the configuration change to my test system and wasn't happy with it. So I'll leave this PR open, solving this issue also for JSPUI. I have created a new ticket on JIRA for this PR: https://jira.duraspace.org/browse/DS-3358

olli-gold:
Accidentally closed this PR. I'm not sure if this is still necessary as we switched to DSpace CRIS and that solves the issue for us. Should I reopen the PR anyway?

PR #1550, opened 2016-10-05
use BulkEditMetadataValue.getValue() instead of private field

pnbecker:
I just encountered a small problem that hinders the servlet container to compile dspace-admin/metadataimport-showchanges.jsp.


pnbecker:
Merged this as an obvious and very small bug/code task resulting from introduction of service API.

PR #1551, opened 2016-10-05
Quickfix for dspace-admin/batchimport.jsp

pnbecker:
Obvious fix for dspace-admin/batchimport.jsp


pnbecker:
Merged as obvious fix.

PR #1555, opened 2016-10-10
DS-3345 Add sql script from dspace 5.6

abollini:
https://jira.duraspace.org/browse/DS-3345

I don't remove the V6.0_2016.08.23__DS-3097.sql because
- it doesn't create any trouble to user that move from <= dspace 5 to DSpace 6 (it is a noop)
- it allows user that are using DSpace 6RC to easy migrate to the official DSpace 6 release (the 6.0 release has take so long that I guess many are using in their working progress project the RC version)

Please note that I have decided to not include the 4.7 sql script as
- it is unknown by flyway (not in DSpace 5.x)
- rerun the DSpace 5.6 and DSpace 6.0 sql script will cause no trouble (noop) to user that come from DSpace 4.7
- user that have installed a 4.7 could have missed to run (manually) the fix sql file so this give more guarantee

I have succesful tested the migration from 4.6, 4.7 and 5.6 previous installations


pnbecker:
+1 by review.

mwoodiupui (approved):
+1.  The logic seems right.  I tested upgrading 4.6 -> 5.6 -> 6.0-SNAPSHOT+this patch and it seems to work.

PR #1557, opened 2016-10-10
DS-3357 rethrow AuthorizationException in the SubmissionControl

abollini:
Small fix to allow the SubmissionController to rethrow authorization excpetion throws performing by submission steps

https://jira.duraspace.org/browse/DS-3357


pnbecker:
+1 by review and discussion in irc.

mwoodiupui (approved):
+1 by inspection.  That looks safe and simple.  Assuming that AuthorizationException is handled well further up the stack, that is an elegant solution.

PR #1572, opened 2016-11-14
DS-3363 CSV import error says "row", means "column" (backport to 6.x)

helix84:
https://jira.duraspace.org/browse/DS-3363


helix84:
Was already approved for master here: https://github.com/DSpace/DSpace/pull/1559

helix84:
Not needed anymore (experiments with branch protection).

PR #1573, opened 2016-11-15
DS-3387 ORCID/SolrAuthority Lookup button doesn't work in Mirage2

helix84:
https://jira.duraspace.org/browse/DS-3387

updating handlebars reportedly resolves the problem

### DIFF
--- a/dspace-xmlui-mirage2/src/main/webapp/bower.json
+++ b/dspace-xmlui-mirage2/src/main/webapp/bower.json
@@ -7,7 +7,7 @@
         "jquery": "~1.10.2",
         "jquery-ui": "1.10.3",
         "jqueryuibootstrap": "4f3772cd37b898f456c0126c4b44178ce8d4aad7",
-        "handlebars": "2.0.0",
+        "handlebars": "4.0.0",
         "respond": "1.4.2",
         "html5shiv": "3.7.2",
         "holderjs": "2.4.1",


alanorth:
Look at the commit again, I think you committed the wrong change.

helix84:
@alanorth Sorry, fixed.

seancarte (approved):
After changing the handlebars version from 2.0.0 to 4.0.0 in bower.json, the 'Lookup' button now launches the Person lookup dialogue. 

Unfortunately this change also seems to affect the advanced search filters.
selection_007

tobosi:
Dear guys

You said that it is claimed that updating handlebar fixes the ORCID lookup functionality, which I agree
with you but have you guys looked at the advanced search filters, because now they are blank and when
I revert back to the default version of handlebar that compiled with DSpace in our case 2.0.0 the advanced
search filters now show.

Is any of you guys facing the same issue?

tobosi (commented):
Does work but messes up your advanced search filters...

bram-atmire (changes requested):
Unfortunately, the current Mirage 2 templates won't work with handlebars 4.0.0. Would currently advise against merging this until Mirage 2 is thoroughly tested against handlebars 4.0.0

tobosi:
Thanks Bram... The silence got for me worried but it is good to see that
there are people looking into this.

tdonohue:
Closing this ticket, as it sounds like we cannot currently upgrade to Handlebars 4.0.0.  Once it is determined this is possible, we can reopen (as needed). Until then, a different solution would need to be found.

tobosi:
Thank you

PR #1576, opened 2016-11-23
DS-3385 add OAI excludeSet filter

helix84:
https://jira.duraspace.org/browse/DS-3385
fixes #6740 

not tested yet

Configuration example:
* Define the filter in the Filters section of xoai.xml:

```
 <Filter id="excludeSetFilter">
   <Class>org.dspace.xoai.filter.DSpaceExcludeSetFilter</Class>
   <Parameter key="excludeSet">
     <Value>col_123456789_371</Value>
   </Parameter>
 </Filter>
```
* Add the filter to your context:
```
 <Context baseurl="request">
   [...]
   <Filter refid="excludeSetFilter" />
 </Context>
```

### DIFF
--- a/dspace-oai/src/main/java/org/dspace/xoai/filter/DSpaceExcludeSetFilter.java
+++ b/dspace-oai/src/main/java/org/dspace/xoai/filter/DSpaceExcludeSetFilter.java
@@ -1,0 +1,80 @@
+/**
+ * The contents of this file are subject to the license and copyright
+ * detailed in the LICENSE and NOTICE files at the root of the source
+ * tree and available online at
+ *
+ * http://www.dspace.org/license/
+ */
+package org.dspace.xoai.filter;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.log4j.LogManager;
+import org.apache.log4j.Logger;
+import org.dspace.xoai.data.DSpaceItem;
+import org.dspace.xoai.filter.results.SolrFilterResult;
+
+import com.lyncode.xoai.dataprovider.xml.xoaiconfig.parameters.ParameterValue;
+import com.lyncode.xoai.dataprovider.xml.xoaiconfig.parameters.SimpleType;
+
+
+/**
+ * Excludes records from the set specified in the <code>excludeSet</code>
+ * parameter.
+ *
+ * @author Àlex Magaz Graça
+ */
+public class DSpaceExcludeSetFilter extends DSpaceFilter {
+    private static Logger log = LogManager.getLogger(DSpaceExcludeSetFilter.class);
+
+    private List<String> excludeSets;
+
+    private List<String> getExcludeSets() {
+        if (this.excludeSets == null) {
+            ParameterValue excludeSets = getConfiguration().get("excludeSet");
+            if (excludeSets == null) excludeSets = getConfiguration().get("excludeSet");
+
+            if (excludeSets instanceof SimpleType) {
+                this.excludeSets = new ArrayList<String>();
+                this.excludeSets.add(((SimpleType) excludeSets).asString());
+            } else {
+                this.excludeSets = new ArrayList<String>();
+                for (ParameterValue val : excludeSets.asParameterList().getValues())
+                    this.excludeSets.add(val.asSimpleType().asString());
+            }
+        }
+        return excludeSets;
+    }
+
+    @Override
+    public boolean isShown(DSpaceItem item) {
+        for (String excludeSet: this.getExcludeSets()) {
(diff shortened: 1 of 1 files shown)


rivaldi8 (changes requested):
String comparisons should be fixed. Apart from this, it works perfectly. I've tested excluding a single collection, multiple collections and a mix of collections and communities. The tests have been done in DSpace 5.6.

dspace-oai/src/main/java/org/dspace/xoai/filter/DSpaceExcludeSetFilter.java
```diff
@@ -0,0 +1,80 @@
...
+        return true;
+    }
+
+    @Override
+    public SolrFilterResult buildSolrQuery() {
+        StringBuilder cond = new StringBuilder("NOT (");
+        List<String> excludeSets = this.getExcludeSets();
+        for (int i = 0; i < excludeSets.size(); i++) {
+            // TODO: distinguish communities/collections
+            if (excludeSets.get(i).substring(0, 4) == "col_") {
```
rivaldi8: This should be:
```
    if (excludeSets.get(i).substring(0, 4).equals("col_")) {
```

dspace-oai/src/main/java/org/dspace/xoai/filter/DSpaceExcludeSetFilter.java
```diff
@@ -0,0 +1,80 @@
...
+                cond.append("item.collections:");
+            } else if (excludeSets.get(i).substring(0, 4) == "com_") {
```
rivaldi8: This should also use `equals`.

rivaldi8:
The configuration should be like this instead:
- In the Filters section of xoai.xml:
```
        <Filter id="excludeSetFilter">
            <Definition>
                <Custom ref="excludeSetFilterCondition"/>
            </Definition>
        </Filter>

        <CustomCondition id="excludeSetFilterCondition">
            <Class>org.dspace.xoai.filter.DSpaceExcludeSetFilter</Class>
            <Configuration>
                <string name="excludeSet">col_123456789_371</string>
            </Configuration>
        </CustomCondition>
```
- In the Context definition (`ref` instead of `refid`):
```
 <Context baseurl="request">
   [...]
   <Filter ref="excludeSetFilter" />
 </Context>
```

rivaldi8 (approved):
That's correct now.

tdonohue (approved):
👍 This code looks reasonable to me. However, if merged, I'd like to see some documentation/examples added on how it can be used.

rivaldi8:
Could someone comfirm that this is working?

After another round of tests in our way to upgrading to DSpace 5.6, I found that the filter is no longer working properly.

Listing all items works (`?verb=ListRecords&metadataPrefix=oai_dc`). I get all records minus the ones in the excluded collection.  parameter).

However, adding a set (`?verb=ListRecords&metadataPrefix=oai_dc&set=com_10459.1_2`) yields no matches.

In the Solr logs I see this query:
```
((NOT (item.collections:col_10459.1_371))) AND ((item.communities:com_10459.1_2))
```

Each condition seems to get two extra pair of parenthesis. If remove them like below, I get the correct list of records (tested from Solr admin):
```
NOT (item.collections:col_10459.1_371) AND ((item.communities:com_10459.1_2))
```

Alternatively, adding `*:*` at the beginning seems to make it work too:
```
((*:* NOT (item.collections:col_10459.1_371))) AND ((item.communities:com_10459.1_2))
```

I'm totally lost as to why this worked and now doesn't, so any hint will be highly appreciated.

rivaldi8:
I've found [this issue](https://jira.duraspace.org/browse/DS-2820) about a similar problem. I still don't understand why it worked in the first tests.

antoine-atmire:
> I'm totally lost as to why this worked and now doesn't, so any hint will be highly appreciated.
Solr doesn't support expressions where everything is a `NOT`. Except when that expression is the top expression... So every time a part within nested parentheses is like that, there aren't any matches.

When you add `*:*` in front of it, the query behaves as expected again. It works now because the expression isn't all-negative anymore. This is a fine solution, it's easy to implement and the logic is sound.

ppmdo:
I've worked this out with the fixed suggested by @antoine-atmire. Although it seems this PR has been abandoned for a while, so is it possible to push the fix to this PR or should I open another?

tdonohue:
Closing, as this PR seems abandoned & sounds like it needs updates [based on the discussion above](https://github.com/DSpace/DSpace/pull/1576#issuecomment-419305799).  If anyone wants to create a fresh PR, please feel free.

PR #1579, opened 2016-12-01
DS-3409 Handle of collections and communities are lost

arvoConsultores:
https://jira.duraspace.org/browse/DS-3409

During migration this sql dont update handles for communities and collections


helix84:
Thanks for the fix. But we can't change a migration that was already released because it won't run again. We need to make this a new migration.

arvoConsultores:
Extracted to external file, but is not tested. 
If other sql use the collection or community uuid from the V6.0_2015.03.07__DS-2701_Hibernate_migration.sql ant this sql will get a null value and can have unpredictable results

mwoodiupui (approved):
I tested this and it appears to work.  +1

tdonohue (approved):
👍 Looks good to me at a glance. This looks ready to merge, as it's been tested/reviewed.

cidus:
Hello,
I don't see these changes in either 6.1 or 6.2 tag, even though it's in the 6.1 milestone. Why is that? Still having issues migration Dspace installation with Oracle.
Sorry if this is the wrong place to ask...

PR #1580, opened 2016-12-01
Deleting collumns drop indexes

arvoConsultores:
Previously created indexes are lost


tdonohue (approved):
👍 Looks good, but needs at least one other person to test on Oracle before merging.

mwoodiupui (approved):
Tested by upgrading a fresh install of DSpace 5.0 to 7.0-SNAPSHOT+this patch, using Oracle Database 10g Express Edition Release 10.2.0.1.0.  The indexes are present after the upgrade, and I was able to create a community and manipulate its policies without incident.

PR #1596, opened 2017-01-02
DS-3367: Fix authorization error on claim by non-admin user

tomdesair:
https://jira.duraspace.org/browse/DS-3367

When a workflow task is removed from the pool in the configurable workflow, WRITE rights corresponding to the workflow group are also removed.

However the exception to this should be the user that claimed the task so that he can still update the item and change its state (archive or put back into workflow).


bram-atmire (approved):
+1 by inspection

tdonohue (approved):
👍 Assuming this has been tested, it looks good to me as well. This also should be cherry-picked to `dspace-6_x`

mwoodiupui (approved):
+1 tested as before (procedure detailed in the Jira issue) and it now behaves as expected.
I should note that I tested on dspace-6_x since master lacks a working GUI at the moment.

mwoodiupui:
Cherry-picked and pushed to dspace-6_x.

PR #1601, opened 2017-01-04
DS-3381 workspace item not saved when using versioning

tomdesair:
This PR fixes the versioning in XMLUI.

In DSpace 6, the context.commit() call was removed. However this call is required so that the item is saved to the database before the call to cocoon.redirectTo() in versioning.js (function startCreateNewVersionItem).


bram-atmire (approved):
tested on 3 different machines in our team

PR #1602, opened 2017-01-05
Fix for DS-3440

cjuergen:
replacing sword-server config reference with swordv2-server in swordv2


tdonohue (approved):
+1 Good catch, @cjuergen.  These look like obviously appropriate changes to me.

When merging this, we need to be sure to also apply it to the `dspace-6_x` branch.

PR #1606, opened 2017-01-06
DS-3441 READ permission on the Collection object not respected by the JSPUI (5_x)

abollini:
backporting of #1605 to 5_x


tdonohue (approved):
+1 Haven't tested this, but the code looks good and the changes look straightforward.

PR #1607, opened 2017-01-07
DS-3289 Removing double slashes in image paths

bram-atmire:


tdonohue (approved):
👍 Obvious fixes, other areas of the same file also do not place an extra slash between ${theme-path} and whatever follows it.

PR #1611, opened 2017-01-09
DS-3446: On bitstream delete, remove policies only after the bitstream has been updated

tomdesair:
Otherwise the current user who is removing the bitstream doesn't have the necessary WRITE rights

https://jira.duraspace.org/browse/DS-3446


tdonohue (approved):
👍 Looks correct to me. I don't see any reason why we'd want the old behavior of removing policies *before* removing the bitstream itself.

jacobcameron:
This fixed the problem I reported.  Thanks so much Tom!

hardyoyo:
+1 by review

kshepherd:
+1

kshepherd:
merged to 6.x and cherry-picked to master

PR #1613, opened 2017-01-11
DS-3436 Sharding SOLR cores corrupts multivalued fields

tomdesair:
https://jira.duraspace.org/browse/DS-3436

Based on the documentation found here https://cwiki.apache.org/confluence/display/solr/Uploading+Data+with+Index+Handlers and some help of my Atmire collegues, I modified the shard logic so that SOLR will spit the values of a multi-valued field when importing the CSV in the newly sharded core.

### DIFF
--- a/dspace-api/src/main/java/org/dspace/statistics/SolrLoggerServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/statistics/SolrLoggerServiceImpl.java
@@ -25,13 +25,16 @@ import org.apache.solr.client.solrj.impl.HttpSolrServer;
 import org.apache.solr.client.solrj.request.AbstractUpdateRequest;
 import org.apache.solr.client.solrj.request.ContentStreamUpdateRequest;
 import org.apache.solr.client.solrj.request.CoreAdminRequest;
+import org.apache.solr.client.solrj.request.LukeRequest;
 import org.apache.solr.client.solrj.response.FacetField;
+import org.apache.solr.client.solrj.response.LukeResponse;
 import org.apache.solr.client.solrj.response.QueryResponse;
 import org.apache.solr.client.solrj.response.RangeFacet;
 import org.apache.solr.client.solrj.util.ClientUtils;
 import org.apache.solr.common.SolrDocument;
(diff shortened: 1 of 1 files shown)


terrywbrady:
@tomdesair , I will test this out.  Do we have a reliable process to follow to force the repair of records?  I will also test to see if the stats repos can be re-indexed properly.

terrywbrady:
@tomdesair , this code looks good.  

I confirmed that this PR does handle the owningComm field properly in newly sharded records.

I recommend that this PR also include a mechanism to repair improperly sharded records.

I have not used the solr-import-statistics and solr-export-statistics commands before testing this, so I do not know how reliable they are.  They do not repair the corrupted fields if an improper shard was created.

Could this same logic be added into the solr-import-statistics code?  That could give repository managers a mechanism to fix their corrupted shards.

tomdesair:
Hi @terrywbrady,

Thanks for testing! I agree with you that we need a repair option. I'll take a look at your suggestion tomorrow.

terrywbrady:
@tomdesair , in my testing I have encountered an issue twice.  I do not know if this issue has anything to do with the updated code or if it is a result of the way that I am manufacturing records for sharding...

When I restart DSpace, my server startup has hung in the init phase.  The last item in the log is 

2017-01-13 13:54:39,809 INFO  org.dspace.statistics.SolrLoggerServiceImpl @ Loading core with name: statistics-2013

Looking at the statistics-2013/index directory, I see a lock file present.

I am going to do some more experimentation and I will report back with my results.

terrywbrady:
@tomdesair , I believe the repair option can be handled separately from this ticket.  Currently, the command solr-reindex-statistics replicates the problem that this PR addresses.  

While there are other pending issues with statistics shards, this PR appears to be valid.  It is just difficult to test due to DS-3457.

terrywbrady (approved):
Comments left in the PR conversation.

hardyoyo:
+1 by review

terrywbrady:
@tomdesair , in the DSpace meeting today I volunteered to port this to 5x and to master.

PR #1629, opened 2017-01-27
[DS-3469] added context and item.gethandle to vscan call to curate

hardyoyo:
This PR is as yet untested, but it does compile correctly. Will test today.

Jira issue: https://jira.duraspace.org/browse/DS-3469

### DIFF
--- a/dspace-api/src/main/java/org/dspace/submit/step/UploadStep.java
+++ b/dspace-api/src/main/java/org/dspace/submit/step/UploadStep.java
@@ -614,7 +614,7 @@ public class UploadStep extends AbstractProcessingStep
                 if (configurationService.getBooleanProperty("submission-curation.virus-scan"))
                 {
                     Curator curator = new Curator();
-                    curator.addTask("vscan").curate(item);
+                    curator.addTask("vscan").curate(context, item.getHandle());
                     int status = curator.getStatus("vscan");
                     if (status == Curator.CURATE_ERROR)
                     {


hardyoyo:
Alas, in testing, this proved to be not quite right:

`2017-01-27 13:32:13,255 WARN  org.dspace.app.webui.servlet.DSpaceServlet @ hpottinger@library.ucla.edu:session_id=92C4AE04243178D787EDC192DE2486DB:ip_addr=128.97.244.108:general_jspui_error:java.io.IOException\colon; Cannot perform curation task(s) on a null object identifier!
java.io.IOException: Cannot perform curation task(s) on a null object identifier!`

dspace-api/src/main/java/org/dspace/submit/step/UploadStep.java
```diff
@@ -614,7 +614,7 @@ public int processUploadFile(Context context, HttpServletRequest request,
                 if (configurationService.getBooleanProperty("submission-curation.virus-scan"))
                 {
                     Curator curator = new Curator();
-                    curator.addTask("vscan").curate(item);
+                    curator.addTask("vscan").curate(context, item.getHandle());
```
hardyoyo: instead of getHandle, I should probably use something else.

mwoodiupui: Doesn't work, eh?  Maybe the Handle has not been created yet.

curate(Context, String) just resolves the handle to a DSO again.  I think we need a new Curator.curate(Context, DSpaceObject) which sets the curation context (curationCtx) to that Context and then delegates to Curator.curate(DSpaceObject).  Then we don't have to resolve a handle that might not even exist yet, just to get back the DSO that we already have.  (There may also be concurrency issues with HandleService.resolveToObject(Context, DSpaceObject.getHandle()).)

hardyoyo: Indeed, I'm mystified how this ever worked correctly. I ran into trouble setting up the vscan step on DSpace 5.x and rather hopefully deferred further attempts until after an upgrade to 6.x. That hope seems to have been misplaced. I agree, we probably need to augment the Curator.curate... though it's used in a few places, so that counts as a refactor.

hardyoyo: hmm... this looks interesting: https://github.com/DSpace/DSpace/blob/269af71afb602808a14edf822ad658c3895e0a37/dspace-api/src/main/java/org/dspace/curate/service/WorkflowCuratorService.java

mwoodiupui: Indeed, see line 59.  Maybe the code in UploadStep just missed being refitted to use services.

mwoodiupui: I could argue that a non-breaking change needed for a bug fix is part of the bug fix, and that we should ignore the fact that it also introduces a new API feature that nobody asked for.

PR #1630, opened 2017-01-30
Add exception message in log when a bean cannot be loaded

cdelacroix:
Hi,

this is my first DSpace pull request, the beans exception was ignored whereas the root cause it contains is usually needed to fix the configuration issue (or head scratching ensues), so it's simply adding the exception message in existing log statements.

Cheers


hardyoyo:
+1 by review

mwoodiupui (approved):
+1 by inspection  Obviously correct.

tdonohue (approved):
Looks good, merging. Thanks!

PR #1631, opened 2017-01-30
annotate unit tests so they can be run directly from a Junit runner

cdelacroix:
When running unit tests directly from eclipse junit4 runner, abstract (base classes) tests are also run as unit tests and rightfully failing.
To prevent that, one solution is to @Ignore tests base classes (implemented here). Another one would be to rename base class names so they don't match any of maven-surefire-plugin patterns (here *Test).


tdonohue (approved):
👍 Seems like reasonable changes to me

PR #1632, opened 2017-01-30
[DS-3469] virus scan during submission attempts to read uploaded bitstream as anonymous user, which fails

mwoodiupui:
https://jira.duraspace.org/browse/DS-3469

This *untested* patch looks like it should supply the necessary identity to authorize the task (assuming that the current session *is* authorized).

### DIFF
--- a/dspace-api/src/main/java/org/dspace/curate/Curator.java
+++ b/dspace-api/src/main/java/org/dspace/curate/Curator.java
@@ -180,8 +180,12 @@ public class Curator
      * Performs all configured tasks upon object identified by id. If
      * the object can be resolved as a handle, the DSO will be the
      * target object.
-     * 
-     * @param c a Dpace context
+     *
+     * <p>
+     * Note:  this method has the side-effect of setting this instance's Context
+     * reference.  The setting is retained on return.
+     *
+     * @param c a DSpace context
      * @param id an object identifier
      * @throws IOException if IO error
      */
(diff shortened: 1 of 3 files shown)


mwoodiupui:
Noted in passing:  WorkflowCuratorServiceImpl probably needs a similar treatment.

dspace-api/src/main/java/org/dspace/curate/Curator.java
```diff
@@ -268,7 +268,22 @@ else if (type == Constants.SITE)
...
+     * (Community, Collection or Item).
+     *
+     * @param c session context in which curation takes place.
+     * @param dso the single object to be curated.
+     * @throws java.io.IOException passed through.
+     */
+    public void curate(Context c, DSpaceObject dso)
+            throws IOException
+    {
+        curationCtx.set(c);
```
mwoodiupui: Ugh, this has the side-effect of leaving the Curator set to the given Context.  Should we save the old reference and restore it before return?

Or should Curator have a setContext(Context) instead?

hardyoyo:
testing this PR locally, it seems more stable. I'm assuming that the UI you're using is XMLUI, I'm using JSPUI and it doesn't give any feedback to what the error might be, just a little yellow exclamation point (see https://jira.duraspace.org/browse/DS-3375). However, uploads of non-virus files seem to work fine, even with the vscan curation task configured, which is good enough for me. I am not sure what Travis is complaining about, but, assuming that is a transient error, +1

hardyoyo (approved):
This all looks good, and passes my testing.

tdonohue (approved):
👍 Looks good and obvious (mostly minor code cleanup).

dspace-api/src/main/java/org/dspace/curate/Curator.java
mwoodiupui: As discussed on IRC, I just added a note to the method doc comment.

PR #1625, opened 2017-01-24
[DS-3458]5x Allow Shard Process to Append to an existing repo

terrywbrady:
https://jira.duraspace.org/browse/DS-3458

This pull request contains the contents of https://github.com/DSpace/DSpace/pull/1623/files.  It is recommended that you install that PR first.

The context for this PR and a suggested test plan is documented here: https://wiki.duraspace.org/display/~terrywbrady/Statistics+Import+Export+Issues

PR #1633, opened 2017-01-30
[DS-3457] Address tomcat hang when multiple solr shards exist in DSpace 6

terrywbrady:
This PR addresses DS-3457 and DS-3458.

* master version: #1634 
* 5.x version: #1625 

The service initialization logic in DSpace 6 introduces a fatal bug when multiple statistics shards are present.

In order to address this, this code defers the initialization of the solr shards until tomcat is processing requests.  On the first request to the server that is requesting statistics data, any existing shards will be started. Example /handle/XXX/YYY/statistics.

Unfortunately, this leaves the shard repos uninitialized until such a request has been made.

A better approach would be to trigger this initialization at the moment that the /solr service has started.  I attempted to add this to a ServletContextListener, but I was not able to get that listener to fire at the correct time.  

I suspect that a better approach exists.  I am putting this PR forward to start a discussion.

This code may be easier to test if you also install https://github.com/DSpace/DSpace/pull/1624.

See the following page for an overview of the existing sharding issues. https://wiki.duraspace.org/display/~terrywbrady/Statistics+Import+Export+Issues

### DIFF
--- a/dspace-api/src/main/java/org/dspace/statistics/SolrLoggerServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/statistics/SolrLoggerServiceImpl.java
@@ -28,6 +28,7 @@ import org.apache.solr.client.solrj.request.CoreAdminRequest;
 import org.apache.solr.client.solrj.response.FacetField;
 import org.apache.solr.client.solrj.response.QueryResponse;
 import org.apache.solr.client.solrj.response.RangeFacet;
+import org.apache.solr.client.solrj.response.SolrPingResponse;
 import org.apache.solr.client.solrj.util.ClientUtils;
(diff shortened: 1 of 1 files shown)


mwoodiupui:
The moment we move to Solr 5+ we would lose the ability to trigger this at Solr startup, since it is packaged only to run with its integrated copy of Jetty and won't be running in Tomcat.  Any long-term approach to this problem should keep that in mind.

kshepherd (approved):
Tests successful. I can start tomcat up with existing shards, and also append stats from main core into existing shards with stats-util -s

hardyoyo:
+1 by review

tdonohue (approved):
👍 Code looks good to me

PR #1637, opened 2017-02-02
[DS-3477] fix altmetrics config lookups in item-view.xsl (6.x)

kshepherd:
Fixes issues noted in [DS-3477](https://jira.duraspace.org/browse/DS-3477) (Mirage item-view.xsl altmetrics config is not read due to deprecated scope reference)

To test, enable altmetrics by enabling it in `[dspace]/config/modules/altmetrics.cfg`:
`altmetric.enabled = true`

To test altmetric badges where URI is a handle set a test item's dc.identifier.uri to http://hdl.handle.net/2022/14471

To test altmetric badges where URI is a DOI, set a test item's dc.identifier.uri to http://doi.org/10.1038/nature.2012.9872 (or set dc.identifier.doi to this value and change identifier in the config like `altmetrics.field = dc.identifier.doi`)


tdonohue (approved):
👍 Code changes look correct to me. These were using old 5.x-type getProperty() references and seem to have been missed in adding the new config system in 6.x

terrywbrady (approved):
Reviewed code and tested locally.  Changes look good.

PR #1639, opened 2017-02-05
DS-3473: add guard code in case no dot is present in bitstream name

rradillen:
https://jira.duraspace.org/browse/DS-3473


bram-atmire (dismissed):
tnx roeland. Change looks good but there's something wrong with enters/linebreaks that makes it look like there are 83 changes in the commit.

mwoodiupui:
The only thing wrong is that our sources are littered with trailing whitespace.  This patch fixes that in passing.

tdonohue (approved):
👍 Looks good/correct to me. The spacing cleanup seems to just be removing extraneous trailing spaces, which is OK by me ;)

PR #1646, opened 2017-02-17
Add the spring-rest overlay to allow building of the new rest7

abollini:
temporary turn off mvn-enforcer dependency convergence
temporary turn off dspace-solr (to migrate to an external SOLR6)

see https://jira.duraspace.org/browse/DS-3482

### DIFF
--- a/dspace-spring-rest/src/main/resources/application.properties
+++ b/dspace-spring-rest/src/main/resources/application.properties
@@ -22,13 +22,14 @@
 # DSpace Settings
 #
 # DSpace installation directory
-dspace.dir=d:/install/dspace7
+dspace.dir=${dspace.dir}
+#dspace.dir=d:/install/dspace7
 
 ########################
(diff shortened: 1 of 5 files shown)


terrywbrady:
@abollini , This code has helped.  I am able to see the HAL browser.

I had to manually delete  slf4j-log4j12.jar from [dspace-install]/webapps/spring-rest/WEB-INF/lib in order to get spring to launch.

Do you envision deploying spring-rest to /spring-rest or to /?

abollini:
I have tested it just deploying as spring-rest but it should work on any context including ROOT.
I haven't had issue with slf4j-log4j12.jar... what do you see in the logs? which version of tomcat do you use? (I have tomcat 8.5.4 on my local environment)

tdonohue (approved):
👍 Approved in today's DSpace 7 UI mtg

PR #1649, opened 2017-02-21
[DS-3501] adjust XML validation

hardyoyo:
skip contents of any folder that includes the text node/node_modules

https://jira.duraspace.org/browse/DS-3501


tdonohue (approved):
👍

PR #1658, opened 2017-02-25
DS-3365 - Duplicated metadata in solr search index

arvoConsultores:
https://jira.duraspace.org/browse/DS-3365

### DIFF
--- a/dspace-api/src/main/java/org/dspace/discovery/SolrServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/discovery/SolrServiceImpl.java
@@ -112,6 +112,7 @@ import org.springframework.stereotype.Service;
  * @author Kevin Van de Velde (kevin at atmire dot com)
  * @author Mark Diggory (markd at atmire dot com)
  * @author Ben Bosman (ben at atmire dot com)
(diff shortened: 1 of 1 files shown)


tdonohue (approved):
👍 Code looks obvious, but I'd like a quick test to ensure this is a common issue, etc.  @terrywbrady has volunteered to test this before merging

terrywbrady:
@arvoConsultores , I am attempting to reproduce the problem that you have reported in DSpace 5x.

Here is the testing process that I am following (without this patch included)

1. Add a "site" and "default" configuration to discovery.xml

                <entry key="site" value-ref="homeConfiguration"/>
                <entry key="default" value-ref="defaultConfiguration"/>

2. Ensure that both configurations share the same "searchFilterAuthor"
3. Re-index the repository index-discovery -f
4. Query SOLR (in SolrAdmin) listing the "author" field for every record
5. Look for duplicate authors in every record.

Unfortunately, I am not seeing the problem that you are attempting to resolve.  Do you have additional testing advice for me to follow?

terrywbrady (changes requested):
Additional instructions are needed on how to reproduce the original problem.

PR #1662, opened 2017-03-01
[DS-3463] Fix IP authentication for anonymous users

minurmin:
Suggested fix for https://jira.duraspace.org/browse/DS-3463. Adds special groups membership check even if no eperson is in the context (as is the case with IP-based authentication). Reported in dspace-tech 23 Feb 2017, tested with Dspace 6.x.


cjuergen (approved):
+1 Tested it and worked fine.

tdonohue (approved):
👍 Code looks good & is tested (see above), merging this into `master` and `dspace-6_x`

PR #1670, opened 2017-03-07
[DS-3521] Bugfix browsing embargoed thumbnail

pbroman:
https://jira.duraspace.org/browse/DS-3521

If `webui.browse.thumbnail.show` is set `true` but a user misses read permission to a bitstream in JSPUI an exception is thrown and the whole browse request results in an internal server error.


pnbecker (approved):
Obvious bug fix. +1

I should mention, that pbroman is my colleague, so I would appreciate another complete review by another DSpace committer.

tdonohue (approved):
👍 Code looks good to me. I haven't given it a test, but it looks rather obvious.

PR #1671, opened 2017-03-08
[DS-3505] Bad redirection from logout action

mwoodiupui:
https://jira.duraspace.org/browse/DS-3505

In root context, HttpRequest.getContextPath() is specified to return the relative "" rather than the absolute "/", which leads to redirect looping.  Instead, on logout redirect to the value of dspace.url, which should do what seems to have been the intent.


mwoodiupui:
Tested on the 6_x branch.  A variant for 5.2 has been in production for several weeks.

tomdesair (approved):
I've tested this PR and it solves the issue.

I think the code is fine.

tdonohue (approved):
👍 Looks good. This is essentially a one-liner with some code cleanup around it.

PR #1673, opened 2017-03-09
[DS-3523] Bugfix for search with embargoed thumbnails

pbroman:
https://jira.duraspace.org/browse/DS-3523

If webui.browse.thumbnail.show is set true but a user lacks read permission to a bitstream shown in the JSPUI search page, an exception is thrown and the whole request results in an internal server error.


pnbecker (approved):
The bug fix is simple and obvious. Comparable to #1670 which is already merged.

full disclosure: pbroman is my colleague.

PR #1674, opened 2017-03-09
[DS-3042] Convert media filters to curation tasks

mwoodiupui:
https://jira.duraspace.org/browse/DS-3042

Fixes: https://github.com/DSpace/DSpace/issues/6398

A copy of @richardrodgers [code](https://github.com/richardrodgers/ctask/tree/master/mediafilter), ported to DSpace 6+, and taking a few suggestions from my IDE.  UNTESTED by me.

### DIFF
--- a/dspace-api/pom.xml
+++ b/dspace-api/pom.xml
@@ -735,6 +735,12 @@
             <artifactId>jackson-annotations</artifactId>
             <version>2.7.0</version>
         </dependency>
+
+        <dependency>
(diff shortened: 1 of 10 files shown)


mwoodiupui:
See https://jira.duraspace.org/browse/DS-4309.  We should ensure that this task can skip over bitstreams that cause errors, and continue.  Errors should be well-reported to facilitate repair of the bitstream.

PR #1678, opened 2017-03-16
DS-3513 Provide an entry endpoint that exposes the known endpoints via the HAL browser

abollini:
Hi @terrywbrady 
here my initial implementation of an endpoint that exposes the known endpoints via the HAL browser.
Start from
https://github.com/4Science/DSpace/blob/DS-3513/dspace-spring-rest/src/main/java/org/dspace/app/rest/RootRestResourceController.java

There are several area that needs improvements
https://github.com/4Science/DSpace/blob/DS-3513/dspace-spring-rest/src/main/java/org/dspace/app/rest/DiscoverableEndpointsService.java#L81
https://github.com/4Science/DSpace/blob/DS-3513/dspace-spring-rest/src/main/java/org/dspace/app/rest/DiscoverableEndpointsService.java#L68
https://github.com/4Science/DSpace/blob/DS-3513/dspace-spring-rest/src/main/java/org/dspace/app/rest/RestResourceController.java#L67

### DIFF
--- a/dspace-spring-rest/src/main/java/org/dspace/app/rest/DiscoverableEndpointsService.java
+++ b/dspace-spring-rest/src/main/java/org/dspace/app/rest/DiscoverableEndpointsService.java
@@ -1,0 +1,87 @@
+/**
+ * The contents of this file are subject to the license and copyright
+ * detailed in the LICENSE and NOTICE files at the root of the source
+ * tree and available online at
+ *
+ * http://www.dspace.org/license/
+ */
+package org.dspace.app.rest;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
+
+import org.springframework.hateoas.Link;
+import org.springframework.stereotype.Service;
+
+/**
+ * A Service able to list all the discoverable endpoints in our REST
+ * application. Endpoints need to register their managed endpoints. The service
+ * is responsible to check conflict and priorities
+ *
+ * @author Andrea Bollini (andrea.bollini at 4science.it)
+ * @author Tim Donohue
+ */
+@Service
+public class DiscoverableEndpointsService {
+	/**
+	 * Contains all the registeredEndpoints as received by the single controller
+	 */
+	private Map<Object, List<Link>> registeredEndpoints = new HashMap<Object, List<Link>>();
+
+	/**
+	 * Is the computed list of discoverableEndpoints valid?
+	 */
+	private boolean initialized = false;
+
+	private List<Link> discoverableEndpoints = new ArrayList<Link>();
+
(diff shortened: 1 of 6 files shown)


dspace-spring-rest/src/main/java/org/dspace/app/rest/DiscoverableEndpointsService.java
```diff
@@ -0,0 +1,87 @@
...
+	/**
+	 * Is the computed list of discoverableEndpoints valid?
+	 */
+	private boolean initialized = false;
+
+	private List<Link> discoverableEndpoints = new ArrayList<Link>();
+
+	public void register(Object controller, List<Link> links) {
+		synchronized (this) {
+			initialized = false;
```
terrywbrady: Should this be set to true?

terrywbrady (approved):
Tested locally.  This works very nicely, and it is good to see the approach you took.

A couple additional thoughts. 
- Should all of the other endpoints have a link back to this endpoint for convenience?
- Would it be simpler to rebuild the list of links on every request rather than having the synchronized methods?  How often will this endpoint actually be invoked?

dspace-spring-rest/src/main/java/org/dspace/app/rest/DiscoverableEndpointsService.java
```diff
@@ -0,0 +1,87 @@
...
+						// sanity check
+						// FIXME improve logging for debugging
+						if (rels.contains(link.getRel())) {
+							throw new IllegalStateException("The rel " + link.getRel() + " is defined multiple times!");
+						}
+						rels.add(link.getRel());
+					}
+				}
+			}
+
```
abollini: now we can set

`initialized = true;`

dspace-spring-rest/src/main/java/org/dspace/app/rest/DiscoverableEndpointsService.java
abollini: no, it should be done after the creation of the result lists in the getDiscoverableEndpoints method (see comment below)

abollini:
> Should all of the other endpoints have a link back to this endpoint for convenience?
This is typically not done. Spring Data Rest doesn't do that, it is arguable but a client should start to use the API from the ROOT so this is an external know information, there are no value to add this information in all the response making them larger than necessary

> Would it be simpler to rebuild the list of links on every request rather than having the synchronized methods? How often will this endpoint actually be invoked?
Following a HAL approach, each client should make this initial request to register which endpoints are needed for each specific functionality. The client should not assume that /api/core/items is the endpoint to use to deal with items, it should only know that our API expose a items resource, the actual link is what is expressed in the ROOT document. This should allow easy replacement of an endpoint with a local customized version.

Compute the links could be intensive once we deal with overriding and customization, this is way I have introduced a cache / syncronization

PR #1679, opened 2017-03-16
Merge the Rest7 branch back to the master

abollini:
This PR allows us to merge back the rest7 branch in the master. We can use it to easily monitor merge conflict if we decide to still wait some days to have dspace-solr restored (it is disabled now in the pom.xml - we have commented out the dspace-solr profile)


abollini:
with the commit https://github.com/DSpace/DSpace/pull/1679/commits/b66066ffb60ad0f8b38f144e2f792221dc198445
we have restored the dspace-solr profile. I'm able to succesful build and deploy solr locally so we should proceed with the merge now

tdonohue (approved):
👍 Approved in today's DSpace 7 UI Meeting

PR #1680, opened 2017-03-20
Add endpoints for the Site object

terrywbrady:
Adding support to obtain the Site object from the repository since this will likely be needed to generate the collection hierarchy.

### DIFF
--- a/dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/SiteRestRepository.java
+++ b/dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/SiteRestRepository.java
@@ -40,7 +40,6 @@ public class SiteRestRepository extends DSpaceRestRepository<SiteRest, UUID> {
 	
 	
 	public SiteRestRepository() {
-		System.out.println("Repository initialized by Spring");
 	}
 
 	@Override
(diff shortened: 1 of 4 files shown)


terrywbrady:
Note that AIP import/export functions require a user to have access to the site object. If we intend to pass those operations through REST, we will need to expose the site object.

terrywbrady:
Updated base to point to master

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/SiteRestRepository.java
```diff
@@ -0,0 +1,83 @@
...
+
+@Component(SiteRest.NAME)
+public class SiteRestRepository extends DSpaceRestRepository<SiteRest, UUID> {
+	SiteService sitesv = ContentServiceFactory.getInstance().getSiteService();
+	@Autowired
+	SiteConverter converter;
+	
+	
+	public SiteRestRepository() {
+		System.out.println("Repository initialized by Spring");
```
mwoodiupui: Shouldn't this go to the log, at DEBUG level?

abollini (changes requested):

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/SiteRestRepository.java
abollini: this is a sysout that I have introduced in early stage of the development and we have copied and paste over all the repository now. We should just remove this statement.

terrywbrady:
@abollini , I made the changes requested and have retested this change

abollini (approved):

PR #1681, opened 2017-03-21
Endpoint to return all top level communities

terrywbrady:
@abollini , I am curious to get your feedback on this PR.

I created a new controller to provide access to the list of top communities via /api/core/communities/top.

I exposed this method from the site object.  Does this approach make sense?

### DIFF
--- a/dspace-spring-rest/src/main/java/org/dspace/app/rest/CommunityResourceController.java
+++ b/dspace-spring-rest/src/main/java/org/dspace/app/rest/CommunityResourceController.java
@@ -1,0 +1,74 @@
+/**
+ * The contents of this file are subject to the license and copyright
+ * detailed in the LICENSE and NOTICE files at the root of the source
+ * tree and available online at
+ *
+ * http://www.dspace.org/license/
+ */
+package org.dspace.app.rest;
+
+import static org.springframework.hateoas.mvc.ControllerLinkBuilder.linkTo;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+
+import org.dspace.app.rest.exception.PaginationException;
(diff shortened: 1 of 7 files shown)


terrywbrady:
Assuming that this implementation looks good, how can this endpoint method become exposed to the list of documented methods /api?

abollini:
Hi Terry,
the implementation looks technically good. Nevertheless I don't feel that it is the direction that we should follow.
I'm not sure that we need a site resource right now. It could become useful especially if we want to work on a multi-tenant dspace but it is not required if we only need to expose top communities. Maybe we want to introduce the resource anyway to hold other information like configuration settings, statistics, etc.

Spring Data REST define a search endpoint for all the managed resource. This endpoints in turn expose all the query methods available for the specific resource. In our case this could mean
/communities/search
will return a HAL document with a least one "top" link
/communities/search/top
that should be implemented as in your controller.

If we like to follow the same approach than spring data rest I expect that some annotation can help us to provide a generic autodiscovery implementation for the /search endpoints and the subsequent enpoint.
We should add methods to the CommunityRestRepository annotated as @searchRestMethod

terrywbrady:
@abollini , I can resubmit the PR without the site stuff.  I agree that it is not incredibly valuable to most client apps.

Note that AIP import/export functions require a user to have access to the site object.  If we intend to pass those operations through REST, we will need to expose the site object.

terrywbrady:
Closing per comments above.  Will recreate

mwoodiupui:
Aw, I had just begun working on Site support, so as to have a place to hang a replacement for DS-552.

terrywbrady:
@mwoodiupui , I kept a PR open that only the Site code: https://github.com/DSpace/DSpace/pull/1680

Perhaps you can weigh in on the merits of the endpoint.  I will add you as a reviewer.

PR #1686, opened 2017-03-23
[DS-3544] Provide endpoint to return top level communities

terrywbrady:
https://jira.duraspace.org/browse/DS-3544

See discussion on https://github.com/DSpace/DSpace/pull/1681 for some background.

@abollini , I created the annotation that you recommended, but I did not implement any code to discover and expose search methods with this annotation.

### DIFF
--- a/dspace-spring-rest/src/main/java/org/dspace/app/rest/CommunityResourceController.java
+++ b/dspace-spring-rest/src/main/java/org/dspace/app/rest/CommunityResourceController.java
@@ -1,0 +1,68 @@
+/**
+ * The contents of this file are subject to the license and copyright
+ * detailed in the LICENSE and NOTICE files at the root of the source
+ * tree and available online at
+ *
+ * http://www.dspace.org/license/
+ */
+package org.dspace.app.rest;
+
+import static org.springframework.hateoas.mvc.ControllerLinkBuilder.linkTo;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+
+import org.dspace.app.rest.exception.PaginationException;
+import org.dspace.app.rest.model.CommunityRest;
+import org.dspace.app.rest.model.hateoas.CommunityResource;
+import org.dspace.app.rest.model.hateoas.DSpaceResource;
+import org.dspace.app.rest.repository.CommunityRestRepository;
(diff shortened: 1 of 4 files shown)


terrywbrady:
See https://github.com/DSpace/DSpace/pull/1688

terrywbrady:
In today's REST meeting, @abollini indicated "my suggestion is to implement the search endpoint in a generic way so that we can fastly add new query for any objects".  Assigning to him for additional direction.

abollini (commented):
The Spring Data Rest project provide a controller to implement the search endpoint across all the spring data repository
https://github.com/spring-projects/spring-data-rest/blob/master/spring-data-rest-webmvc/src/main/java/org/springframework/data/rest/webmvc/RepositorySearchController.java
we should try to mimic this behavior

artlowel:
As discussed in today's rest meeting:

Currently in the Angular UI it is assumed that you can pass a parameter called `scopeID` to the `/communities` endpoint.
(actually the `/collections` endpoint, because we decided to remove the concept of communities from the Rest API and UI in dspace 7, but to prevent confusion I'll keep calling them communities for now)

A GET of `/communities?scopeID=65b194f9-100c-42fd-b93f-cd82d99b70e4` would return the direct children of the community with id `65b1…70e4`

Leaving out that `scopeID` param would return the top-level communities.

Perhaps that is an approach to consider?

abollini:
@artlowel I prefer to follow the usual structure of REST apps, each endpoints should return the list of all the resources of a specific type (i.e. /communities should return all the communities not only the top level communities).
To work on a collection of resources I propose to use the /search sub-endpoint so that looking to /communities/search the client can discover all the available options to get "subset" of the communities list for instance
/communities/search/top will return the ones without parent

about the merge of communities and collections we have agree to make an effort to have these two concepts merged in the new DSpace 7. My understanding was that the new UI will have for sure a single "page" for both entities and *if possible* the REST API will expose the two objects on a single endpoint. So I haven't yet created a task for that and anyway we are proceeding now with two separate endpoints. After DSpace 7 we will work to have the two concepts merged also at the database/low level api
Accomplish that as a later refactor make sense or is it better to plan and work on that immediately also on the REST API?

abollini:
I'm closing this PR in favor of #1726

PR #1687, opened 2017-03-23
[DS-3487] Implement Simple Hierarchy Traversal

terrywbrady:
In today's REST meeting, I raised a question about how/when to initialize community and collection relationships.   I created this PR to draw attention to this question.

What is the proper way to trigger the initialization of the subcommunity and collection relationships for a parent community?

Question
1. What request will trigger a community to initialize its child community/collection relationships?
2. When examining a CommunityRest object, how can you distinguish an empty child list from an uninitialized child list?


terrywbrady:
Looking at https://jira.duraspace.org/browse/DS-3533, the suggestion appears to be to pass the projection to the converter.  I will create an alternative that makes use of that field

terrywbrady:
Closing this PR in favor of an approach that uses projection.

PR #1690, opened 2017-03-28
Dspace 6 ds 3545 mirage2: custom sitemap.xmap is ignored

samuelcambien:
Fixes #6895

There are two resources plugins:

* one for the local customizations (outside the maven-iterator-plugin)
* one for the dspace source files (inside the maven-iterator-plugin)

The first one is ran first, afterwards the other one runs.
The property 'overwrite=false' should avoid that local customizations would be overwritten by dspace source files.
But the property 'filtering=true' causes maven to ignore the overwrite property.
So the sitemap (the only file that will be filtered) will always be overwritten by the dspace original!

To fix this, I also moved the local customizations resources-plugin to inside the maven-iterator-plugin.
This is necessary anyway, because to filter a sitemap, the name of the theme directory is needed for the ${theme-path} and ${theme-name} variables.
I put it after the resources-plugin for the dspace source, and gave it 'overwrite=true'.
So the order is now reversed: first the source files are processed, and then overwritten by any local customizations.

I also made the split between files to filter and files not to filter consistent accross the two resources plugins: only filter the sitemap.
(instead of filtering everything, except a list of image extensions)

### DIFF
--- a/dspace/modules/xmlui-mirage2/pom.xml
+++ b/dspace/modules/xmlui-mirage2/pom.xml
@@ -66,57 +66,6 @@
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-resources-plugin</artifactId>
-                <executions>
-                    <execution>
-                        <goals>
-                            <goal>copy-resources</goal>
-                        </goals>
-                        <phase>process-resources</phase>
-                        <configuration>
-                            <outputDirectory>${project.build.directory}/themes</outputDirectory>
-                            <resources>
-                                <resource>
-                                    <directory>${basedir}/src/main/webapp/themes</directory>
-                                    <excludes>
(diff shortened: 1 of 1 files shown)


samuelcambien:
also committed this to

dspace-5_x
https://github.com/DSpace/DSpace/pull/1691

master
https://github.com/DSpace/DSpace/pull/1689

hardyoyo:
I'm not sure why Travis-CI testing of this PR has stalled, I can't find it anywhere in the Travis-CI logs for pull requests. I think our only option is to temporarily close then re-open this PR, which I'm willing to do, but I figured I'd mention it here before I try. Another option is to rebase this PR on upstream dspace-6_x, which isn't required, but *would* force Travis-CI to re-run tests.

hardyoyo:
This PR looks good, though, and should be easy to test, since it's just a change to the build.

bram-atmire (approved):

PR #1692, opened 2017-03-31
DS-3549: Encode bitstreams with UTF-8 in XMLUI

PhilipVis:
When viewing a bitstream in the browser the file is encoded as ISO-8859 instead of UTF-8. This change forces the UTF-8 encoding. 

https://jira.duraspace.org/browse/DS-3549

See also #1693

### DIFF
--- a/dspace-xmlui/src/main/webapp/WEB-INF/web.xml
+++ b/dspace-xmlui/src/main/webapp/WEB-INF/web.xml
@@ -94,19 +94,22 @@
   <!--
     - Declare a filter to force UTF-8 encoding for all servlet requests
     -->
-  <filter>
-    <filter-name>SetCharacterEncoding</filter-name>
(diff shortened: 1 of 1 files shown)


AlexanderS:
This obviously breaks all ISO-8859 content. Some browsers (Chrome) will get the right encoding from the data if no encoding is transmitted in the response header:

* Without this patch both UTF-8 and ISO-8859 are displayed correct in Chrome, but only ISO-8859 in Firefox.
* With this patch only UTF-8 works in Chrome and Firefox.

Additionally I do not think, that `application/pdf;charset=UTF-8` is a valid `Content-Type`.

kshepherd:
This PR seems to need further discussion, and the travis report isn't a pass yet. Shifting to 6.4 until we can resolve those issues (if we get some action in the next week I'll move back to 6.3)

PR #1695, opened 2017-04-04
DS-3553: added context.complete to VersionManager.processCreateNewVersion

samuelcambien:
https://jira.duraspace.org/browse/DS-3553

when creating a new item version, I get the following error:
TypeError: Cannot call method "getSubmitter" of null (resource://aspects/Submission/submission.js#229)
this is because in submission.js:
getWorkspaceItemService().find(getDSContext(), workspaceID);
returns null.

this is because VersionManager.processCreateNewVersion() creates a new work space item, and passes the workspace item id to the request, but doesn't commit these changes.


tdonohue (changes requested):
This PR needs to be updated (or recreated) to point at our `dspace-6_x` maintenance branch. The XMLUI no longer exists on `master`, and any further bug fixes to it should be applied to `dspace-6_x` instead.

samuelcambien:
done
https://github.com/DSpace/DSpace/pull/1703

PR #1696, opened 2017-04-05
DS-2748: Do not throw an exception in the PageNotFoundTransformer

tomdesair:
But do return a 404 error code.

This will keep the Cocoon logs clean and prevent Cocoon from leaking memory.


tdonohue (approved):
👍 Code looks correct. We should simply return a 404 response instead of throwing an error.  I'd prefer a sanity check test here though to ensure Cocoon passes this 404 response up to the browser (as we'd expect it to).  I know we've had past issues with Cocoon + 404s, so testing this may be warranted.

AlexanderS:
Checked this and it seems to work correctly:

404-page-not-found

mwoodiupui (approved):
+1 by inspection.  The bulk of the patch simply removes cacheability, which seems sensible for "not found".  The meat of the change is replacing a throw with setting the response code.  A tiny, obvious change.

PR #1697, opened 2017-04-05
DS-2748: Do not throw an exception in the PageNotFoundTransformer

tomdesair:
But do return a 404 error code.

This will keep the Cocoon logs clean and prevent Cocoon from leaking memory.

See also #1696


tdonohue (approved):
👍 Code looks correct. We should simply return a 404 response instead of throwing an error.  I'd prefer a sanity check test here though to ensure Cocoon passes this 404 response up to the browser (as we'd expect it to).  I know we've had past issues with Cocoon + 404s, so testing this may be warranted.

mwoodiupui (approved):
+1 by inspection.  Same as #1696.

PR #1701, opened 2017-04-08
DS-2852 - Discovery label fix for authority display value

tomdesair:
Fixes #6208
This pull request will make sure the author's name is displayed when clicking a discovery facet and the author is under authority control.

### DIFF
--- a/dspace-api/src/main/java/org/dspace/app/itemimport/ItemImport.java
+++ b/dspace-api/src/main/java/org/dspace/app/itemimport/ItemImport.java
@@ -284,7 +284,7 @@ public class ItemImport
             if (line.hasOption('z'))
             {
                 zip = true;
-                zipfilename = sourcedir + System.getProperty("file.separator") + line.getOptionValue('z');
+                zipfilename = line.getOptionValue('z');
             }
 
             //By default assume collections will be given on the command line
@@ -733,7 +733,7 @@ public class ItemImport
                 {
                     clist = mycollections;
                 }
-                addItem(c, mycollections, sourceDir, dircontents[i], mapOut, template);
+                addItem(c, clist, sourceDir, dircontents[i], mapOut, template);
                 System.out.println(i + " " + dircontents[i]);
                 c.clearCache();
             }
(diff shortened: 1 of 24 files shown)


tomdesair:
This has been rebased and can be considered final now.

hardyoyo:
I've skimmed this code and it looks OK to me. I will note that I asked as a comment on DS-2852 if this applies to DSpace 6x as well, and @tomdesair said that it did. Do we need a PR for that? Or is forward-porting enough?

AlexanderS:
Forward-porting is not enough `AuthorityValueFinder` is not available in the `dspace-6_x` branch anymore. I prepare a second pull request for DSpace 6x.

AlexanderS:
I just tested this with the mirage theme and it did not work... It breaks the UI and fail if the filter is applied again:

f4fe2e15-abf7-4d61-b577-6fc900041fd9

If you press "Apply" the value (f.e. "Haak, Laurel") is submitted as the ID and that obviously does not work.

LotteHofstede:
About the UI problem: 
This issue was not limited to the Mirage theme, but also caused problems for Mirage2. I changed the name of the display value field and hid it.

AlexanderS:
So the patch currently does not change the display in the Mirage theme?! :confused:

LotteHofstede:
DS-2852 only relates to Mirage2. 
The specific issue is that the "tag" beneath the discovery search input field is showing an identifier instead of something human readable (Eg. for authors --> full name). 

This "tag" html element does not exist in Mirage(1), so there's nothing to "fix" in Mirage(1). 

The advanced filters still show the identifier (in both Mirage2 and Mirage), this seems logical, as it is explicitly searching for an identifier. (The dropdown values say "Author" and "ID"; it would be strange if you would expect a name to appear here instead of an ID)

kshepherd:
hi @tomdesair , this looks good... if you'd like this fix to get into 6.3 could you please rebase to dspace-6_x or cherrypick and submit a new PR to dspace-6_x? we can give it a quick re-test then should be good to go!

J4bbi (approved):
Same code as #1800, I have test it and confirmed it works.

PR #1708, opened 2017-04-12
DS-3568. UTF-8 characters are now supported in configuration files

christian-scheible:
https://jira.duraspace.org/browse/DS-3568

To reproduce the error add a UTF-8 character to dspace.cfg and look at the result before and after the patch.

1. Change configuration like this: `dspace.name = DSpace at My Univörsity`
2. View the outcome at http://YOUR_TEST_INSTALLATION/oai/driver?verb=Identify

This sets the encoding for the configuration files dspace.cfg and local.cfg to UTF-8.
This works on all included configuration files to.

### DIFF
--- a/dspace/config/config-definition.xml
+++ b/dspace/config/config-definition.xml
@@ -31,14 +31,14 @@
 
         <!-- Allow user to override any configs in a local.cfg -->
         <!-- Any properties in this config will override defaults in dspace.cfg (or any included *.cfg file) -->
-        <properties fileName="local.cfg" throwExceptionOnMissing="false" config-name="local" config-optional="true">
+        <properties fileName="local.cfg" throwExceptionOnMissing="false" config-name="local" config-optional="true" encoding="UTF-8">
             <!-- Reload this file if it has changed, and at least 5 seconds (5,000 ms) have passed -->
             <reloadingStrategy refreshDelay="5000"
                 config-class="org.apache.commons.configuration.reloading.FileChangedReloadingStrategy"/>
         </properties>
 
         <!-- Load our dspace.cfg (which in turn loads all module configs via "include=" statements) -->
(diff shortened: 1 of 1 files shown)


tdonohue (approved):
:+1: This looks like an obvious fix to me.

Once accepted & merged, this would also need to be cherry-picked to the `dspace-6_x` maintenance branch.

hardyoyo:
I've been so used to working on the dspace-6_x branch, it threw me for a loop to work with the master branch... I will test this today, and if it works, I will merge it and cherry-pick it into the dspace-6_x branch.

tomdesair:
For me this doesn't work and I don't know why:

1. I've started Tomcat 8 with `-Dfile.encoding=UTF-8`
2. In local.cfg, I have `dspace.name = DSpacé at 即 الذين My Univörsity`
3. That file is UTF-8 encoded in the install directory:
```
$ file -I config/local.cfg
config/local.cfg: text/plain; charset=utf-8
```
4. But the `oai/request?verb=Identify` page returns "DSpac???? at ?????? ???????????????????? My Univ????rsity"

@christian-scheible Any idea what's going wrong?

christian-scheible:
@tomdesair. I am not sure about the problem. Your setup seems to be right. The only difference I can see is that I am currently using tomcat7.
auswahl_004

christian-scheible:
@tomdesair, I just checked again (removing encoding="UTF-8" and adding it again).
It took several restarts of tomcat and oai clean-cache until the change was visible.

tomdesair:
Ok, I'll try again

AlexanderS:
:+1: Works for me after `dspace oai clean-cache`.

tomdesair:
👍  Now it indeed works without problems.

mwoodiupui (approved):
Tested:  works.  Obvious fix.

PR #1709, opened 2017-04-13
DS-3516 ImageMagick PDF Thumbnail class should only process PDFs

alanorth:
Input formats for ImageMagick mediafilter plugins should be defined in dspace.cfg. Currently they implement `SelfRegisteredInputFormats`, which causes the PDF filter to attempt to process other formats such as JPEGs that might be in the `ORIGINAL` bitstream bundle.

### DIFF
--- a/dspace-api/src/main/java/org/dspace/app/mediafilter/ImageMagickPdfThumbnailFilter.java
+++ b/dspace-api/src/main/java/org/dspace/app/mediafilter/ImageMagickPdfThumbnailFilter.java
@@ -43,10 +43,4 @@ public class ImageMagickPdfThumbnailFilter extends ImageMagickThumbnailFilter {
 	    }
     }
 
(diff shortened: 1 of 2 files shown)


alanorth:
This one is ready for you @tdonohue and @terrywbrady.

tdonohue (approved):
👍 Looks correct to me. For more explanation on these changes, see this [JIRA comment](https://jira.duraspace.org/browse/DS-3516?focusedCommentId=55072&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-55072) 

NOTE: we should also cherry-pick this to `dspace-6_x` and `master`, as the same minor changes are necessary on both those branches as well.

terrywbrady (approved):

terrywbrady:
+1, tested and verified fix

PR #1717, opened 2017-04-19
[DS-3564] Limit maximum idle database connections by default

mwoodiupui:
https://jira.duraspace.org/browse/DS-3564

If there are "too many" idle database connections, have the pool close some.  By default keep no more than 10.

At this writing, untested.

This should be ported to master as well.


pnbecker (approved):

pnbecker:
+1 by review

tdonohue (approved):
👍 Looks good to me.

PR #1718, opened 2017-04-19
[DS-3505] Bad redirection from logout action (ported to dspace-5_x)

mwoodiupui:
Like the patch for DSpace 6, but for DSpace 5.  Should also work with DSpace 4.


rivaldi8 (approved):
It works for me. Tested with DSpace 5.6.

PR #1726, opened 2017-04-20
DS-3544 add support to search methods for repositories

abollini:
initial draft with sample implementation of the find top level communities.
When a repository have a method annotated with @SearchRestMethods the resource endpoint (i.e. /api/core/communities) return a link to the /search (/api/core/communities/search).
The repository search endpoint is implemented listing all the available search methods i.e. for communities 
/api/core/communities/search/top

The search method endpoint works right now but we need to improve the invocation strategy discovering via reflection which parameters need to be used (now it is assumed that only a pagination is required but of course we need for other use cases to add query parameters). See
https://github.com/DSpace/DSpace/compare/master...4Science:DS-3544?expand=1#diff-091f26fa1a213c77e1ff0275f1da97d4R195

### DIFF
--- a/dspace-spring-rest/src/main/java/org/dspace/app/rest/RestResourceController.java
+++ b/dspace-spring-rest/src/main/java/org/dspace/app/rest/RestResourceController.java
@@ -293,15 +293,16 @@ public class RestResourceController implements InitializingBean {
 	
 	@RequestMapping(method = RequestMethod.GET, value="/search/{searchMethod}")
 	@SuppressWarnings("unchecked")
-	<T extends RestModel> PagedResources<DSpaceResource<T>> executeSearchMethods(@PathVariable String apiCategory, 
+	<T extends RestModel> ResourceSupport executeSearchMethods(@PathVariable String apiCategory, 
 			@PathVariable String model, @PathVariable String searchMethod, Pageable pageable, Sort sort, PagedResourcesAssembler assembler, 
 			@RequestParam MultiValueMap<String, Object> parameters) throws IllegalAccessException, IllegalArgumentException, InvocationTargetException {
 			
 		Link link = linkTo(this.getClass(), apiCategory, model).slash("search").slash(searchMethod).withSelfRel();
 		DSpaceRestRepository repository = utils.getResourceRepository(apiCategory, model);
-		Page<DSpaceResource<T>> resources = null;
+		Object searchResult = null;
 		boolean searchEnabled = false;
 		boolean searchMethodFound = false;
+		boolean returnPage = false;
 		for (Method method : repository.getClass().getMethods()) {
 			SearchRestMethod ann = method.getAnnotation(SearchRestMethod.class);
 			if (ann != null) {
@@ -311,8 +312,9 @@ public class RestResourceController implements InitializingBean {
 					name = method.getName();
 				}
 				if (StringUtils.equals(name, searchMethod)) {
+					returnPage = method.getReturnType().isAssignableFrom(List.class);
 					searchMethodFound = true;
-					resources = ((Page<T>) executeQueryMethod(repository, parameters, method, pageable, sort, assembler)).map(repository::wrapResource);
+					searchResult = executeQueryMethod(repository, parameters, method, pageable, sort, assembler);
 					break;
 				}
 			}
@@ -323,7 +325,15 @@ public class RestResourceController implements InitializingBean {
 		if (!searchEnabled) {
 			throw new RepositorySearchNotFoundException(model);
 		}
-		PagedResources<DSpaceResource<T>> result = assembler.toResource(resources, link);
+		
+		ResourceSupport result = null;
+		if (returnPage) {
+			Page<DSpaceResource<T>> resources = ((Page<T>) searchResult).map(repository::wrapResource);
+			result = assembler.toResource(resources, link);
+		}
+		else {
+			result = repository.wrapResource((T) searchResult);
+		}
 		return result;
 	}
 	
--- a/dspace-spring-rest/src/main/java/org/dspace/app/rest/RestResourceController.java
+++ b/dspace-spring-rest/src/main/java/org/dspace/app/rest/RestResourceController.java
@@ -7,6 +7,7 @@
  */
 package org.dspace.app.rest;
 
+import static org.springframework.data.rest.webmvc.ControllerUtils.EMPTY_RESOURCE_LIST;
 import static org.springframework.hateoas.mvc.ControllerLinkBuilder.linkTo;
 import static org.springframework.hateoas.mvc.ControllerLinkBuilder.methodOn;
 
@@ -16,6 +17,7 @@ import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.UUID;
+import java.util.Map.Entry;
 
 import javax.servlet.http.HttpServletRequest;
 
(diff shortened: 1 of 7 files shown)


abollini:
Rebased on the master. Mimic the spring data rest implementation I have introduced the support for search methods with custom parameter. The method signature is analyzed using reflection an any request parameter is automatically bind using the spring convert framework.
See 
https://github.com/4Science/DSpace/blob/DS-3544/dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/CommunityRestRepository.java#L102
for an example. The param annotation is optional but help to avoid accidental breaking the rest contract due to rename of the method parameters

dspace-spring-rest/src/main/java/org/dspace/app/rest/RestResourceController.java
```diff
@@ -226,4 +264,180 @@ private void checkModelPluralForm(String apiCategory, String model) {
 			throw new RepositoryNotFoundException(apiCategory, model);
 		}
 	}
+	
+	@RequestMapping(method = RequestMethod.GET, value="/search")
+	ResourceSupport listSearchMethods(@PathVariable String apiCategory, @PathVariable String model) {
+		ResourceSupport root = new ResourceSupport();
+		DSpaceRestRepository repository = utils.getResourceRepository(apiCategory, model);
+		boolean searchEnabled = false;
+		for (Method method : repository.getClass().getMethods()) {
```
tomdesair: Because the `RestResourceController` is getting quite big, would it be possible to move all "reflection" related logic to a `DSpaceRestReflectionHelper` class (or something similar)? For example this for-loop could be moved to a method `List<String> getSearchMethodNames(DSpaceRestRepository repository)` after the names can be used to construct links.

dspace-spring-rest/src/main/java/org/dspace/app/rest/RestResourceController.java
```diff
@@ -226,4 +264,180 @@ private void checkModelPluralForm(String apiCategory, String model) {
...
+	<T extends RestModel> PagedResources<DSpaceResource<T>> executeSearchMethods(@PathVariable String apiCategory, 
+			@PathVariable String model, @PathVariable String searchMethod, Pageable pageable, Sort sort, PagedResourcesAssembler assembler, 
+			@RequestParam MultiValueMap<String, Object> parameters) throws IllegalAccessException, IllegalArgumentException, InvocationTargetException {
+			
+		Link link = linkTo(this.getClass(), apiCategory, model).slash("search").slash(searchMethod).withSelfRel();
+		DSpaceRestRepository repository = utils.getResourceRepository(apiCategory, model);
+		Page<DSpaceResource<T>> resources = null;
+		boolean searchEnabled = false;
+		boolean searchMethodFound = false;
+		for (Method method : repository.getClass().getMethods()) {
```
tomdesair: This for-loop could be moved to a `Method findSearchMethod(DSpaceRestRepository repository, String searchMethod)` in the reflection helper class.

dspace-spring-rest/src/main/java/org/dspace/app/rest/RestResourceController.java
```diff
@@ -226,4 +264,180 @@ private void checkModelPluralForm(String apiCategory, String model) {
...
+			result.put(parameter.getParameterName(), entry.getValue());
+		}
+
+		return invokeQueryMethod(repository, method, result, pageable, sort);
+	}
+	
+	/*
+	 * Adapted from org.springframework.data.repository.support.ReflectionRepositoryInvoker.invokeQueryMethod(Method, MultiValueMap<String, ? extends Object>, Pageable, Sort)
+	 */
+	public Object invokeQueryMethod(DSpaceRestRepository repository, Method method, MultiValueMap<String, ? extends Object> parameters, Pageable pageable,
```
tomdesair: I would also move these out of the `RestResourceController` class into a reflection / invoker utility class.

dspace-spring-rest/src/main/java/org/dspace/app/rest/RestResourceController.java
```diff
@@ -226,4 +264,180 @@ private void checkModelPluralForm(String apiCategory, String model) {
...
+
+			if (Pageable.class.isAssignableFrom(targetType)) {
+				result[i] = pageable;
+			} else if (Sort.class.isAssignableFrom(targetType)) {
+				result[i] = sortToUse;
+			} else {
+
+				String parameterName = param.getParameterName();
+
+				if (!StringUtils.isNotBlank(parameterName)) {
```
tomdesair: I think this be replaced by `StringUtils.isBlank`.

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/CommunityRestRepository.java
```diff
@@ -74,12 +76,50 @@ public CommunityRest findOne(Context context, UUID id) {
...
+	// communities
+	@SearchRestMethod(name="top")
+	public Page<CommunityRest> findAllTop(Pageable pageable) {
+		List<Community> topCommunities = new ArrayList<Community>();
+		int total = 0;
+		try {
+			List<Community> it = cs.findAllTop(obtainContext());
+			total = it.size();
+			for (Community c : it) {
+				topCommunities.add(c);
```
tomdesair: Can't you use `topCommunities.addAll(it)` here?

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/CommunityRestRepository.java
```diff
@@ -74,12 +76,50 @@ public CommunityRest findOne(Context context, UUID id) {
...
+		} catch (SQLException e) {
+			throw new RuntimeException(e.getMessage(), e);
+		}
+		Page<CommunityRest> page = utils.getPage(topCommunities, pageable).map(converter);
+		return page;
+	}
+
+	// TODO: add method in dspace api to support direct query for subcommunities
+	// with pagination and authorization check
+	@SearchRestMethod(name="subCommunities")
```
tomdesair: Why didn't we add subcommunities to the `org.dspace.app.rest.model.CommunityRest` and embedded them in the `org.dspace.app.rest.model.hateoas.CommunityResource` class (limited to a depth of 1)? I think that in most cases you need the subcommunity information if you fetch the Community information.

tomdesair (changes requested):
Overall it looks good. I've added some minor comments and I think we should move the reflection logic to a separate utility class/service.

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/CommunityRestRepository.java
tomdesair: See also https://jira.duraspace.org/browse/DS-3650

dspace-spring-rest/src/main/java/org/dspace/app/rest/RestResourceController.java
abollini: Good suggestion thanks. I'm going to rebase the PR on the current master and refactor the code

dspace-spring-rest/src/main/java/org/dspace/app/rest/RestResourceController.java
abollini: ok

dspace-spring-rest/src/main/java/org/dspace/app/rest/RestResourceController.java
abollini: ok

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/CommunityRestRepository.java
abollini: We need to find a good way to manage lazy load optimized collections, please check the discussion here: https://wiki.duraspace.org/display/DSPACE/DSpace+7+REST%3A+Coding+DSpace+Objects#DSpace7REST:CodingDSpaceObjects-LinkedRepositoryObject(TODO-proposal)

I suggest to keep the search method subCommunities in place for now as a simple example of searchMethod we can decide later if we want to remove it and provide only a /api/communities/:id/subCommunities endpoint instead (btw, a similar discussion could happen also for the top communities that could be retrieved from the site endpoint)

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/CommunityRestRepository.java
abollini: indeed it can be simplified. We don't need two lists at all

abollini:
@tomdesair @tdonohue I have rebased the PR and applied the proposed refactor. IMHO, it should be ready to merge now (waiting for Travis to confirm)

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/CommunityRestRepository.java
tomdesair: Ok agreed, we can work on this in the lazy loading/projections discussions.

tdonohue (approved):
👍 This looks good to me.  No objections. 

I do see that @tomdesair still has an outstanding change request (an "X") next to his code review, but it looks like his concerns have all been addressed.

abollini:
@tomdesair I will merge this tomorrow if you raise no objection

PR #1730, opened 2017-04-24
DS-2675: Bugfixing: Jump to value general errors with order (for 6.x)

jonas-atmire:
https://jira.duraspace.org/browse/DS-2675

Contains the following:

Fixed "starts_with" implementation to only retrieve items that ACTUALLY start with the string.
Ascending/Descending ordering now properly reset the offset (instead of remaining somewhere halfway)

5.x version is #1710

### DIFF
--- a/dspace-api/src/main/java/org/dspace/browse/BrowseDAO.java
+++ b/dspace-api/src/main/java/org/dspace/browse/BrowseDAO.java
@@ -393,4 +393,8 @@ public interface BrowseDAO
     public boolean isEnableBrowseFrequencies();
 
 	public void setEnableBrowseFrequencies(boolean enableBrowseFrequencies);
+
+    public void setStartsWith(String startsWith);
+
+    public String getStartsWith();
 }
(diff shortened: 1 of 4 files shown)


AlexanderS:
I just tested this with mirage theme and it somehow works. But jumping to a letter while browsing (f.e.) by author it only returns the authors starting the the letter, now. You cannot select the previous/next page anymore. Is this intended?

jonas-atmire:
@AlexanderS 
That is correct, this is the intended behaviour -> If the user selected the "starts_with" for a certain letter, only those results are returned. 
The previous functionality on this page could have some weird side effects that it wasn't actually properly filtered. If a user clicks "starts with p" and there are no results starting with p, a whole slew of non-related results are returned. (This also has the side effect that updating the ordering is quite strange)

On the first page, the "previous page" button would then be disabled of course, because there are no previous results. 
IF there are more than on the currently displayed page (this can, for example, be checked by setting the results per page to 5 or some other low number), then the "next page" button should still be visible.

AlexanderS:
So it is actually not "jump to" anymore, but "filter".

terrywbrady (approved):
+1, reviewed and tested locally

kshepherd (approved):
+1 tested locally, thanks @jonas-atmire 
the non-UI components of this PR need porting to master, I think

PR #1740, opened 2017-04-27
DS-3141 Allow metadata field internationalization in XMLUI

edusperoni:
This is a recreation of #1365 in the dspace-6_x branch with a few additions.

Fixed some whitespace inconsistency

Mirage2 support is also added.

Here's how it looks like on Mirage2:

image


edusperoni:
Oh boy, here we go.

So, testing this I came to the conclusion that changing:

```
<language value-pairs-name="common_iso_languages">true</language>
```
to:
```
<language value-pairs-name="anything">true</language>
```

had absolutely no effect.

## Problem 1: It uses the hardcoded `LanguageName = "common_iso_languages".`

I found this:

https://github.com/DSpace/DSpace/blob/master/dspace-api/src/main/java/org/dspace/app/util/DCInput.java#L36

So I went to change it and then I saw that `value-pairs-name` is also hardcoded into the fieldMap:

```
if ("dropdown".equals(inputType) || "qualdrop_value".equals(inputType)
                || "list".equals(inputType))
        {
            valueListName = fieldMap.get("value-pairs-name");
            valueList = listMap.get(valueListName);
        }
```

But then, how do we get the `value-pairs-name` for the language?

## Problem 2: `value-pairs-name` is used by two field simultaneously

Here the PAIR_TYPE_NAME is defined:

https://github.com/DSpace/DSpace/blob/master/dspace-api/src/main/java/org/dspace/app/util/DCInputsReader.java#L53-L54

But here it gets ugly:

https://github.com/DSpace/DSpace/blob/master/dspace-api/src/main/java/org/dspace/app/util/DCInputsReader.java#L376-L423

As you can see `field.put(PAIR_TYPE_NAME, pairTypeName);` is called for both language and types dropdown, qualdrop_value, and list, which means one will overwrite the other. And worse, the language is not even being used.

I have two possible solutions to this:

1. Use the same value-pairs-name
2. Create a new language-value-pairs-name to avoid conflict.

The final solution is simple. After we decide between what to use above, we jsut need to rewrite a single line in the DCInput. Here's the ugly version:
```
valueLanguageList = listMap.get(fieldMap.get("language-value-pairs-name") == null? LanguageName : fieldMap.get("language-value-pairs-name"));
```

@tdonohue @hardyoyo I'd like some guidance on this. I can update this PR, but this change will also need to made in the master. Should I open a new jira issue or continue working in this one?

I already have it working in my test enviroment

dspace-xmlui-mirage2/src/main/webapp/xsl/core/forms.xsl
```diff
@@ -549,7 +571,43 @@
...
+                            <xsl:variable name="name">
+                                <xsl:choose>
+                                    <xsl:when test="dri:params[@operations='add delete']">
+                                        <xsl:value-of select="concat(@n, '_', $nro)"/>
+                                    </xsl:when>
+                                    <xsl:otherwise>
+                                        <xsl:value-of select="@n"/>
+                                    </xsl:otherwise>
+                                </xsl:choose>
+                            </xsl:variable>
```
AlexanderS: This `xsl:variable` block is indented one level too far. (Sorry for the nitpicking, but it looks like it's a child of the variable `nro` above.)

dspace-xmlui-mirage2/src/main/webapp/xsl/core/forms.xsl
```diff
@@ -265,6 +265,28 @@
...
 
 
+    <xsl:template name="languageSelectBuilder">
+        <xsl:param name="name"/>
+        <xsl:param name="selected"/>
+        <select>
+            <xsl:call-template name="fieldAttributes"/>
+            <xsl:apply-templates/>
+            <xsl:attribute name="name"><xsl:value-of select="concat($name, '[lang]')"/></xsl:attribute>
+            <xsl:for-each select="../dri:params/dri:param[@name='langs']/dri:option | ./dri:params/dri:param[@name='langs']/dri:option">
```
AlexanderS: The `../dri:params` part is never used? As far as I can see, the template is only called with a `test="dri:params/dri:param[@name = 'langs']"`around. So the `dri:params` element should always be a child of the current element.

dspace-xmlui-mirage2/src/main/webapp/xsl/core/forms.xsl
```diff
@@ -265,6 +265,28 @@
     </xsl:template>
 
 
+    <xsl:template name="languageSelectBuilder">
+        <xsl:param name="name"/>
+        <xsl:param name="selected"/>
+        <select>
+            <xsl:call-template name="fieldAttributes"/>
+            <xsl:apply-templates/>
```
AlexanderS: Are you sure this should be here?

dspace-xmlui/src/main/webapp/themes/dri2xhtml-alt/core/forms.xsl
```diff
@@ -770,6 +777,25 @@
         </xsl:choose>
     </xsl:template>
 
+    <!-- build the select field with the language options -->
+    <xsl:template name="languageSelectBuilder">
+        <xsl:param name="name"/>
+        <xsl:param name="selected"/>
+        <select>
+            <xsl:attribute name="name"><xsl:value-of select="concat($name, '[lang]')"/></xsl:attribute>
+               <xsl:for-each select="../dri:params/dri:param[@name='langs']/dri:option | ./dri:params/dri:param[@name='langs']/dri:option">
```
AlexanderS: Here applies the same comment like in Mirage2: I cannot find a place where this template is called, so that the `../dir:params/` part is used.

dspace-xmlui-mirage2/src/main/webapp/xsl/core/forms.xsl
edusperoni: The fieldAttributes are required, but the apply-templates might not be. I'll try removing it and testing.

dspace-xmlui-mirage2/src/main/webapp/xsl/core/forms.xsl
edusperoni: I'll confess this and the dri2xhtml-alt code were directly ported from https://github.com/DSpace/DSpace/pull/1365, I'll try removing the `../dri:params` and will update the PR as needed.

MW3000:
I just wanted to test the functionality but did not have time to look at the code yet:
I applied the pull request to 6.3 and added some `<language value-pairs-name="common_iso_639-2_languages">true</language>` to `onebox` fileds in my `input-forms.xml`.
It seems to have no effect at all: There does not appear an additional dropdown menu with language options in my form. Have you successfully tested this @edusperoni and @AlexanderS ?

edusperoni:
We're currently using this on 6.2 with no issu (we're using Mirage2). Does `common_iso_639-2_languages` exist in your form? We use it in almost every field, so it should work in onebox

MW3000:
@edusperoni Your code does work in 6.3. There was an interference with some changes we added in modules overlays.
@tdonohue You can bring back the _quick win_ tag :)

PR #1741, opened 2017-05-03
DS-2291 Autocomplete not working on Mirage2

edusperoni:
Fixes DS-2291.

The autocomplete was not working due to it not being included in the scripts.xml.

Also, the collection wasn't between single quotes, which threw another javascript error.

End result:

image

### DIFF
--- a/dspace-xmlui-mirage2/src/main/webapp/scripts.xml
+++ b/dspace-xmlui-mirage2/src/main/webapp/scripts.xml
@@ -13,6 +13,10 @@
     <script src="vendor/jquery/jquery.min.js">&#160;</script>
     <script src="vendor/jquery-ui/ui/jquery.ui.core.js">&#160;</script>
     <script src="vendor/jquery-ui/ui/jquery.ui.datepicker.js">&#160;</script>
+    <script src="vendor/jquery-ui/ui/jquery.ui.widget.js">&#160;</script>
+    <script src="vendor/jquery-ui/ui/jquery.ui.position.js">&#160;</script>
+    <script src="vendor/jquery-ui/ui/jquery.ui.menu.js">&#160;</script>
+    <script src="vendor/jquery-ui/ui/jquery.ui.autocomplete.js">&#160;</script>
     <script src="vendor/handlebars/handlebars.js">&#160;</script>
     <script src="vendor/holderjs/holder.js">&#160;</script>
     <!--<script src="vendor/bootstrap-sass-official/assets/javascripts/bootstrap/affix.js">&#160;</script>-->
--- a/dspace-xmlui-mirage2/src/main/webapp/styles/classic_mirage_color_scheme/_jquery_ui.scss
+++ b/dspace-xmlui-mirage2/src/main/webapp/styles/classic_mirage_color_scheme/_jquery_ui.scss
@@ -16,7 +16,7 @@
 }
 
 input[type='text'].ui-autocomplete-loading{
-    background-image: url("../../images/authority_control/lookup-indicator.gif");
+    background-image: url("../images/authority_control/lookup-indicator.gif");
     background-repeat: no-repeat;
     background-position: 99% 50%;
     background-size: 25px 25px;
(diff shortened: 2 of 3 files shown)


hardyoyo:
I'm still trying to test this, I tried to cherry-pick it to my branch where I'm working on implementing authority control for my fork of DSpace, as it would have been easier to test in that way... however, I never could get the suggest/autocomplete to work. So... I'll go back to testing in isolation instead, probably in Vagrant-DSpace.

edusperoni:
This is the config I'm using:

```
plugin.selfnamed.org.dspace.content.authority.ChoiceAuthority = \
    org.dspace.content.authority.DSpaceControlledVocabulary, \
    org.dspace.content.authority.DCInputAuthority


choices.plugin.dc.subject.cnpq = common_cnpq
choices.presentation.dc.subject.cnpq = suggest
choices.closed.dc.subject.cnpq = true
authority.controlled.dc.subject.cnpq = true
authority.required.dc.subject.cnpq = true
```

I define common_cnpq in the inputform. I never managed to get a controlled vocabulary to work properly for the field. It worked only with the DCInputAuthority

hardyoyo:
Thanks, I think I found it... and a good thing, too... our Mirage2 theme uses the Maven overlay... and the overlay has a copy of the *bad* choice-authority-control.xsl file in it... and... you can figure out the rest. Hilarity ensues. I'll give it another try after this meeting I have to attend.

hardyoyo:
This does in fact fix both DS-2291 as well as DS-3611. I have tested this in my fork of DSpace on our staging server. As noted above, you have to be careful about replacing choice-authority-control.xsl in any Maven Overlay you might be using. :-\ I will approve this change, but here's my +1 while I'm typing.

hardyoyo (approved):

edusperoni:
Ok, we might have to hold it for a second.

I'm doing my own testing and found another error.

This lookup button doesn't work:

image


It's directly tied with the other commit which dealt with the `addLookupButton` template.

You can see the `javascript:DSpaceChoiceLookup` is missing quotes around the collection:

```
<input class="pull-right ds-button-field btn btn-default ds-add-button" name="lookup_value_34" type="button" value="Lookup" onclick="javascript:DSpaceChoiceLookup('/admin/lookup', 'dc_publisher_program', 'aspect_administrative_item_EditItemMetadataForm_div_edit-item-status', 'value_34', 'value_34_authority', 'aspect_administrative_item_EditItemMetadataForm_field_value_34_confidence_indicator', 11664701-bd0c-4053-819c-77fb39dc97d7, false, false);">
```

The discrepancy lies in the following code. In addLookupButtonAuthor:

```
          <!-- Collection ID for context -->
                <xsl:choose>
                    <xsl:when test="/dri:document/dri:meta/dri:pageMeta/dri:metadata[@element='choice'][@qualifier='collection']">
                        <xsl:text>'</xsl:text>
                        <xsl:value-of select="/dri:document/dri:meta/dri:pageMeta/dri:metadata[@element='choice'][@qualifier='collection']"/>
                        <xsl:text>'</xsl:text>
                    </xsl:when>
                    <xsl:otherwise>
                        <xsl:text>-1</xsl:text>
                    </xsl:otherwise>
                </xsl:choose>
```

and in addLookupButton:

```
          <!-- Collection ID for context -->
          <xsl:choose>
            <xsl:when test="/dri:document/dri:meta/dri:pageMeta/dri:metadata[@element='choice'][@qualifier='collection']">
              <xsl:value-of select="/dri:document/dri:meta/dri:pageMeta/dri:metadata[@element='choice'][@qualifier='collection']"/>
            </xsl:when>
            <xsl:otherwise>
              <xsl:text>-1</xsl:text>
            </xsl:otherwise>
          </xsl:choose>
```

I believe both this and #1764 are needed. I can update my PR to include this fix.

edusperoni:
I went ahead and updated the PR with the fix. Makes no sense to implement this change without fixing the lookup bug

PR #1742, opened 2017-05-05
DS-3558 Port from dspace 6 to dspace 7

Frederic-Atmire:
Ported this pull request: https://github.com/DSpace/DSpace/pull/1707 to master, all the tests still work so I'm assuming this works correctly, but review is needed.
Also tested this with the rest-api by doing some requests with different user-agents and solr still logged the right statistics.


hardyoyo (approved):

tdonohue (approved):
👍 Looks good. Merging as this passes tests and looks to be the same code as #1707

PR #1747, opened 2017-05-18
DS-3598: Allow localization of input-forms.xml with XMLUI

AlexanderS:
https://jira.duraspace.org/projects/DS/issues/DS-3598

### DIFF
--- a/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/submission/submit/DescribeStep.java
+++ b/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/submission/submit/DescribeStep.java
@@ -11,10 +11,14 @@ import java.io.IOException;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Locale;
+import java.util.Map;
 import java.util.UUID;
 
 import javax.servlet.ServletException;
 
+import org.apache.avalon.framework.parameters.Parameters;
+import org.apache.cocoon.ProcessingException;
+import org.apache.cocoon.environment.SourceResolver;
 import org.dspace.app.util.DCInput;
 import org.dspace.app.util.DCInputSet;
 import org.dspace.app.util.DCInputsReader;
@@ -48,6 +52,7 @@ import org.dspace.content.authority.service.ChoiceAuthorityService;
 import org.dspace.content.authority.service.MetadataAuthorityService;
 import org.dspace.content.factory.ContentServiceFactory;
 import org.dspace.content.service.ItemService;
+import org.dspace.core.I18nUtil;
 import org.dspace.services.factory.DSpaceServicesFactory;
 import org.xml.sax.SAXException;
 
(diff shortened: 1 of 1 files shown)


dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/submission/submit/DescribeStep.java
```diff
@@ -64,238 +69,224 @@
...
+        this.requireStep = true;
+    }
+
+    @Override
+    public void setup(SourceResolver resolver, Map objectModel, String src, Parameters parameters) throws ProcessingException, SAXException, IOException
     {
-        if (INPUTS_READER == null)
+        super.setup(resolver, objectModel, src, parameters);
+
+        String formFileName = I18nUtil.getInputFormsFileName(context.getCurrentLocale());
```
tomdesair: I think this is the only necessary change to fix DS-3598. So I'm not sure why we need all these other changes.

dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/submission/submit/DescribeStep.java
```diff
@@ -416,893 +407,900 @@ else if (inputType.equals("qualdrop_value"))
...
+                DCPersonName dpn = new DCPersonName(dcValue.getValue());
+
+                lastName.addInstance().setValue(dpn.getLastName());
+                firstName.addInstance().setValue(dpn.getFirstNames());
+                Instance fi = fullName.addInstance();
+                fi.setValue(dcValue.getValue());
+                if (isAuthorityControlled)
                 {
-                    if (dcInput.getWarning() != null && dcInput.getWarning().length() > 0)
+                    if (dcValue.getAuthority() == null || dcValue.getAuthority().equals(""))
```
tomdesair: It's not clear to me here why you changed "getWarning" to "getAuthority" and how this is related to DS-3598.

tomdesair (changes requested):
Hi @AlexanderS,

Could you limit the code changes to the changes required to solve DS-3598? That would make it a lot easier to review.

This implementation indeed aligns XMLUI with the JSPUI approach, but personally I don't like the JSPUI approach. If you have a repository that supports 6 different languanges, you will have to maintain 6 different input-forms files. A better approach would be to support message-keys in the input-forms.xml file, e.g.:

```
<field>
          <dc-schema>dc</dc-schema>
          <dc-element>title</dc-element>
          <dc-qualifier></dc-qualifier>
          <repeatable>false</repeatable>
          <label>Submission.submit.input.dc.title.label</label>
          <input-type>onebox</input-type>
          <hint>Submission.submit.input.dc.title.hint</hint>
         <required>Submission.submit.input.dc.title.required</required>
</field>
```

But that would be an backwards-incompatible change. So I think this PR is OK for DSpace 6.x, but I think we would need a better approach for master.

dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/submission/submit/DescribeStep.java
AlexanderS: Did you see, that this PR contains two commits? The first one is regarding DS-3598 and is comparatively small. The second one only contains whitespace changes.

If you review the commits in the github UI you may add a `?w=1` to the URL to ignore all whitespace changes.

dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/submission/submit/DescribeStep.java
AlexanderS: This is actually not a change. The diff algorithm is not able to detect correct changes, due to the massive change of whitespaces. Before the change the file used a mixture of spaces and tabs and the indentation was distorted. But if you want, I can remove the second commit fixing the whitespace issues.

For review you want to look at the single commits and/or add `?w=1` to the URL.

AlexanderS:
Regarding the "better approach for master": This is not usable on master anyway, because it is part of XMLUI. I do not know, if the i18n system of dspace-angular will be compatible with the current message-key, but this is future for now.

hardyoyo:
here's a link to the whitespace-change-ignored diff: https://github.com/DSpace/DSpace/pull/1747/files?w=1

tomdesair:
👍 

I've tested the current approach and it works. The code also looks OK when using `?w=1`.

But as mentioned above, I think we need a better approach for DSpace 7 so that a repository administrator doesn't have to maintain an `input-forms.xml` file per locale.

terrywbrady (approved):
+1, tested locally by enabling webui.supported.locales = en, de, fr and creating input-forms_fr.xml with some simple changes.

PR #1749, opened 2017-05-18
DS-3588 configuration of CORS headers for the API

abollini:
It allows the use of the application.properties to set the allowed origin for cross origin call.
It solves CORS issues that can be observed using the HAL browser in a domain and the REST API on a different one. It supports pre-flight requests as by spring documentation and tested using custom header on the HAL browse that force the client to send a preflight request also for GET methods (that are the only one implemented right now in the DSpace API).


tdonohue (approved):
👍 Looks good

PR #1750, opened 2017-05-18
DS-3593 allow aggregation of endpoints by category / Include UUID in serialization

abollini:
Add a category to each REST Repository so to structure the endpoints by categories

/api/core/communities
/api/core/collections
...

/api/eperson/groups
/api/eperson/epersons
...

etc.

It is easy to change the categories just editing the CATEGORY constant in each REST domain object

### DIFF
--- a/dspace-spring-rest/src/main/java/org/dspace/app/rest/RestResourceController.java
+++ b/dspace-spring-rest/src/main/java/org/dspace/app/rest/RestResourceController.java
@@ -49,7 +49,7 @@ import org.springframework.web.bind.annotation.RestController;
  *
  */
 @RestController
-@RequestMapping("/api/core/{model}")
+@RequestMapping("/api/{apiCategory}/{model}")
 @SuppressWarnings("rawtypes")
 public class RestResourceController implements InitializingBean {
 	@Autowired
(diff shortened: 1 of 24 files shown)


tdonohue (approved):
👍 Looks good to me. I like the ease of changing `{apiCategory}` added in here.

terrywbrady:
@abollini , I was just typing a review as the merge occurred.  One change to consider: consider using an enum for the list of available categories rather than string constants.  This might provide some flexibility in the future.  One caveat... it would be important to reserve a category for local instance specific functions.

abollini:
@terrywbrady sorry for that. I agree, I'm not happy with the current solution to have all the categories listed as constant in the RestModel Interface. It was just an initial choice but we could find a better place/solution later. Ideally it should be not necessary to reserve categories for local instance specific functions. New endpoint can be added defining new category at any time as it is just a string not an enum. Should we open an issue to move forward/track this discussion? do you want to add some note on the code practice documentation on that?

terrywbrady:
@abollini , I will create an issue.  I would recommend changing the method signatures to take the category enum (or an interface implemented by the category enum).

terrywbrady:
See https://jira.duraspace.org/browse/DS-3606

PR #1752, opened 2017-05-22
DS-3601: Fix NPE when accessing feedback page without "Referer" header

AlexanderS:
https://jira.duraspace.org/browse/DS-3601


tdonohue (approved):
👍 Code looks correct to me. Thanks!

This could use a quick test from a volunteer, and then this looks good to merge.

tomdesair:
Tested and works as advertised. The user is presented with the error: 

> The feedback mail page may only be invoked from another DSpace page

+1

terrywbrady (approved):
+1, tested through postman.  Verified change from NPE to AuthorizeException

PR #1755, opened 2017-05-31
DS-3404: JSPUI: Fix authority lookup

AlexanderS:
https://jira.duraspace.org/browse/DS-3404

This is #1578 against dspace-6_x slightly reworked and with an additional fix for `edit-item-form.jsp` like mentioned in the original pull request.


tdonohue (approved):
👍 Code looks correct. Seems obvious that some single quotes are missing here. That said, I haven't tested it yet.

kshepherd (approved):
+1, yep, collection UUIDs will have introduced the need to quote the result of getID()

PR #1757, opened 2017-05-31
DS-3245: CSV linebreaks not supported by Bulkedit

rivaldi8:
I haven't found why the change to the `isNotBlank` check was introduced, so I've just reverted it. Not sure if it may introduce some regression.

I've tested it on DSpace 5.6. On master, only through unit tests.

Fixes https://jira.duraspace.org/browse/DS-3245

### DIFF
--- a/dspace-api/src/test/java/org/dspace/app/bulkedit/DSpaceCSVTest.java
+++ b/dspace-api/src/test/java/org/dspace/app/bulkedit/DSpaceCSVTest.java
@@ -43,7 +43,8 @@ public class DSpaceCSVTest extends AbstractUnitTest
                             "+,56599ad5-c7d2-4ac3-8354-a1f277d5a31f,Easy line,\"Lewis, Stuart\",A nice short abstract",
                             "+,56599ad5-c7d2-4ac3-8354-a1f277d5a31f,Two authors,\"Lewis, Stuart||Bloggs, Joe\",Two people wrote this item",
                             "+,56599ad5-c7d2-4ac3-8354-a1f277d5a31f,Three authors,\"Lewis, Stuart||Bloggs, Joe||Loaf, Meat\",Three people wrote this item",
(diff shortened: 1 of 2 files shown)


tomdesair:
I've tested this and it works!

We should also port this to DSpace 6.

tomdesair:
@rivaldi8 Can you create a PR for DSpace 6 of this fix?

rivaldi8:
@tomdesair Sure, I'll work on it.

terrywbrady (approved):
+1, Tested in 6x

dspace-api/src/test/java/org/dspace/app/bulkedit/DSpaceCSVTest.java
```diff
@@ -43,7 +43,8 @@ public  void testDSpaceCSV()
                             "+,56599ad5-c7d2-4ac3-8354-a1f277d5a31f,Easy line,\"Lewis, Stuart\",A nice short abstract",
                             "+,56599ad5-c7d2-4ac3-8354-a1f277d5a31f,Two authors,\"Lewis, Stuart||Bloggs, Joe\",Two people wrote this item",
                             "+,56599ad5-c7d2-4ac3-8354-a1f277d5a31f,Three authors,\"Lewis, Stuart||Bloggs, Joe||Loaf, Meat\",Three people wrote this item",
-                            "+,56599ad5-c7d2-4ac3-8354-a1f277d5a31f,\"Two line\ntitle\",\"Lewis, Stuart\",abstract",
+                            "+,56599ad5-c7d2-4ac3-8354-a1f277d5a31f,\"Two line\n\ntitle\",\"Lewis, Stuart\",abstract",
```
helix84: The change on this line seems unintentional; the two newlines test is added below.

helix84:
A late +1 apart from my comment above.

PR #1759, opened 2017-05-31
DS-3281: Start workflow for REST submissions

AlexanderS:

This is #1743 but against dspace-6_x.

Note: The log message at the end of the function `"Item successfully created in collection ..."` may be wrong now, but #1743 also did not touch that.


tdonohue (approved):
👍 Code looks good. I also gave this a test, and verified that the submission now gets put into Workflow.

PR #1760, opened 2017-06-01
DS-3582

pnbecker:
https://jira.duraspace.org/browse/DS-3582


hardyoyo (approved):
+1 obviously correct

pnbecker:
We just encountered that we'll need more JSPs fixed. => Marking this work in progress.

pnbecker:
In the same commit that remove the `context.abort()` call, three more JSPs were changed. This PR no fixes all of them.

pnbecker:
We discussed exactly this approach yesterday in slack. The resulting code change is obvious. I'll merge it without another review.

PR #1762, opened 2017-06-02
[DS-3563] Port PR to 6x

terrywbrady:
https://jira.duraspace.org/browse/DS-3563

https://github.com/DSpace/DSpace/pull/1706


terrywbrady:
@tomdesair , @kshepherd , I think I encountered a dependency on this while testing, so I cherry-picked this PR

kshepherd (approved):
+1 by inspection, if the dspace 6/master db schema still uses metadatavalue table in a similar way then we'll want this index to be created. Thanks @terrywbrady !

PR #1773, opened 2017-06-13
[DS-3615] Allows changing of value-pairs-name for language tag.

edusperoni:
https://jira.duraspace.org/browse/DS-3615
If this gets approved I'll port it to master.

### DIFF
--- a/dspace-api/src/main/java/org/dspace/app/util/DCInput.java
+++ b/dspace-api/src/main/java/org/dspace/app/util/DCInput.java
@@ -33,7 +33,7 @@ public class DCInput
     private boolean language = false;
     
     /** the language code use for the input */
-    private String languageName = "common_iso_languages";
+    private String languageName = null;
 
     /** the language list and their value */
     private List<String> valueLanguageList = null;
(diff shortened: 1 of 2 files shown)


dspace-api/src/main/java/org/dspace/app/util/DCInput.java
```diff
@@ -114,7 +114,8 @@ public DCInput(Map<String, String> fieldMap, Map<String, List<String>> listMap)
         valueLanguageList = new ArrayList();
         if (language)
         {
-            valueLanguageList = listMap.get(LanguageName);
+            languageName = fieldMap.get("language-value-pairs-name");
```
AlexanderS: Can we use the default value here, if no language-value-pairs-name is given?

AlexanderS: Ah scratch this. I just saw that `DCInputsReader` is enforcing the language attribute.

dspace-api/src/main/java/org/dspace/app/util/DCInput.java
```diff
@@ -33,7 +33,7 @@
     private boolean language = false;
     
     /** the language code use for the input */
-    private static final String LanguageName = "common_iso_languages";
+    private String languageName = "common_iso_languages";
```
AlexanderS: This can be `null`. No need for a value, that is never used.

edusperoni: Good call, added it to the PR.

AlexanderS (approved):
Tested it and it works.

PR #1774, opened 2017-06-14
[DS-3602] Incremental Update of Legacy Id fields in Solr Statistics (w/ Solr Schema Change)

terrywbrady:
Based on the discussion of https://github.com/DSpace/DSpace/pull/1766, here is an alternative approach that updates the SOLR schema.

@tomdesair , @wwelling , @kshepherd , FYI

### DIFF
--- a/dspace-api/src/main/java/org/dspace/util/SolrUpgradeStatistics6.java
+++ b/dspace-api/src/main/java/org/dspace/util/SolrUpgradeStatistics6.java
@@ -75,6 +75,8 @@ public class SolrUpgradeStatistics6
         //Maximum number of solr ids to pull in a facet query... when processing infrequently accessed ids, the
         //facet size may grow.
         private static final int    FACET_LIMIT = 1000;
+        //Once the number of records returned from a facet is smaller than this number, process all items for a query
+        private static final int    FACET_CUTOFF = 10;
         
         private static final String INDEX_DEFAULT = "statistics";
         private static final String MIGQUERY = "NOT(dspaceMig:*)";
@@ -402,9 +404,7 @@ public class SolrUpgradeStatistics6
                         while(run(FIELD.scopeId, Constants.COMMUNITY) > 0){};
 
                         //Handle irregular type/scopetype values
-                        for(FIELD field: FIELD.values()){                                
-                                while(run(field, -1) > 0){};                                
-                        }
+                        while(run(FIELD.epersonid, -1) > 0){};
                 }
                 
                 if (numProcessed > 0) {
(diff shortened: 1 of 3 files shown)


terrywbrady:
Latest commit improves facet handling.  Looking at the process output over 3.3M records, ~700 records are processed more than once.

terrywbrady:
@tomdesair , @wwelling , @kshepherd , I know the priority on this issue has gone down, but I found an additional optimization for the code.  The prior logic was using facets to group similar items resulting in many small solr queries.  I realized I could approximate that logic by using only sort fields.  

While overall processing time is unchanged, there are 2 benefits to this approach: (1)the code is simpler and (2)there is no accidental re-processing of records.  I hope to now leave this code until there is feedback.

terrywbrady:
Closing.   If we decide this PR is needed sooner than 7.0, this PR would be appropriate to re-open.

-------

RE 3602: I plan to resolve the outstanding issue with this PR for 6.2: https://github.com/DSpace/DSpace/pull/1782  The comprehensive fix for this issue is addressed in https://github.com/DSpace/DSpace/pull/1774  Since 1774 requires a Solr schema change, do we want to defer it to DSpace 7 or target it for 6.2?  I could be convinced of either option.  I just want to label the milestone properly.


Tim Donohue [9:20 AM] 
@terrywbrady: I think the Solr Schema change question is entirely a question of priority.  If we have a definitive reason why Solr *must* have the Schema change to function properly in 6.x, then it should be in 6.2.  If there's no problem with behavior in 6.x, then it at least needs to be in 7.0 (and there may be no rush to put it into 6.2)


[9:22] 
When scheduling is concerned (with any upcoming PR), the understanding also needs to be that 7.0 isn't likely until mid-to-late 2018 (and that's a very rough estimate).  Anything needing more immediate fixing should go into a 6.2 (which is as yet unscheduled, and needs a coordinator, but likely could happen later this year)


[9:23] 
with the Solr Schema change, it sounds like we feel it likely isn't _required_ in 6.2...but it would likely be good to verify there's no adverse behavior to leaving this change until 7.0


Terry Brady [9:30 AM] 
Thanks @tdonohue.  There is some risk to maintaining the current inconsistency of legacy stat records. PR 1782 missed the 6.1 release because of an unexpected side effect of relying on both the legacy id and the uuid for statistics capture.  Given that we did not opt for the complete fix (with schema change) for 6.1, it probably makes sense to defer 1774 to 7.0.  If we change our minds, we can reassign it.


Tim Donohue [9:31 AM] 
Sounds fine to me. I'm OK with it either way...just wanted to note what the decision points should be.  :slightly_smiling_face:


Terry Brady [9:50 AM] 
I created a new PR for 7.0: https://github.com/DSpace/DSpace/pull/1810.  I will close 1774 along with a snapshot of this discussion.  I suspect that I will wish I had advocated more strongly to get these PR's merged.

PR #1776, opened 2017-06-16
Missed later improvements

ghost:
Lato fonts
Fixed askus glyphicon
Scoped css


tdonohue:
Hi @wwelling : This looks like an accidental PR. I'm assuming you likely wanted to create this elsewhere, so I'm going to close it.  Please, let me know if I'm mistaken

wwelling:
Hi @tdonohue, absolutely was unintended. Thought pressing back fast enough would cancel it. Guess not. When making a pull request from a fork of a repo it defaults to upstream to merge into.

wwelling:
@tdonohue I think there is this bug though, https://github.com/DSpace/DSpace/blob/dspace-6_x/dspace-api/src/main/java/org/dspace/eperson/service/GroupService.java#L196

The call here, https://github.com/DSpace/DSpace/blob/dspace-6_x/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/versioning/Navigation.java#L120, can not find method due the interface method signature not having public member access keyword.

tdonohue:
@wwelling : If you could log a bug, or submit a quick PR, I'd appreciate it. I'm not going to have a chance to look at this until after OR2017. But, perhaps someone else can do so.

PR #1780, opened 2017-06-21
DS-3595

PhilipVis:
This change prevents errors when switching the language on the submission or admin pages in theme Mirage2. 

jira ticket
https://jira.duraspace.org/browse/DS-3595

This change is based on the fix that is already added to DSpace for theme Mirage:
https://jira.duraspace.org/browse/DS-1810

PR #1781, opened 2017-06-21
DS-3595: Language change causes "page not found" error in several forms on XMLUI

PhilipVis:
This change prevents errors when switching the language on the submission or admin pages in theme Mirage2.

jira ticket
https://jira.duraspace.org/browse/DS-3595

This change is based on the fix that is already added to DSpace for theme Mirage:
https://jira.duraspace.org/browse/DS-1810


tdonohue (approved):
👍 Looks good to me. This is an identical port of #1780 (which was already tested/merged into 6.x)

PR #1782, opened 2017-06-21
[DS-3602] Ensure Consistent Use of Legacy Id in Usage Queries

terrywbrady:
@kshepherd identified that the initial issue reported in the following ticket was that statistics records with a legacyId in owningItem were not being included in the Usage reports.

It appears that the legacy id was also not used in queries for scopeId, owningColl, and owningComm.

There is an another PR (https://github.com/DSpace/DSpace/pull/1774) for this ticket to migrate all legacy ids in statistics records.  This PR is recommended in addition to that PR.  This PR can be merged without merging 1774.

This PR will ensure the accessibility of legacy statistics records until that upgrade is performed.

### DIFF
--- a/dspace-api/src/main/java/org/dspace/statistics/content/StatisticsDataSearches.java
+++ b/dspace-api/src/main/java/org/dspace/statistics/content/StatisticsDataSearches.java
@@ -12,6 +12,7 @@ import org.apache.commons.lang.StringUtils;
 import org.apache.solr.client.solrj.SolrServerException;
 import org.apache.solr.client.solrj.util.ClientUtils;
 import org.dspace.content.DSpaceObject;
+import org.dspace.content.DSpaceObjectLegacySupport;
 import org.dspace.core.Context;
 import org.dspace.services.factory.DSpaceServicesFactory;
 import org.dspace.statistics.Dataset;
@@ -201,8 +202,12 @@ public class StatisticsDataSearches extends StatisticsData {
     protected String getQuery() {
         String query;
         if(currentDso != null){
-            query = "scopeType: " + currentDso.getType() + " AND scopeId: " + currentDso.getID();
-
+            query = "scopeType: " + currentDso.getType() + " AND ";
+            if(currentDso instanceof DSpaceObjectLegacySupport){
+                query += " (scopeId:" + currentDso.getID() + " OR " + ((DSpaceObjectLegacySupport) currentDso).getLegacyId() + ")";
+            }else{
+                query += "scopeId:" + currentDso.getID();
+            }
         }else{
             query = "*:*";
         }
--- a/dspace-api/src/main/java/org/dspace/statistics/content/StatisticsDataSearches.java
+++ b/dspace-api/src/main/java/org/dspace/statistics/content/StatisticsDataSearches.java
@@ -204,7 +204,7 @@ public class StatisticsDataSearches extends StatisticsData {
         if(currentDso != null){
             query = "scopeType: " + currentDso.getType() + " AND ";
             if(currentDso instanceof DSpaceObjectLegacySupport){
-                query += " (scopeId:" + currentDso.getID() + " OR " + ((DSpaceObjectLegacySupport) currentDso).getLegacyId() + ")";
+                query += " (scopeId:" + currentDso.getID() + " OR scopeId:" + ((DSpaceObjectLegacySupport) currentDso).getLegacyId() + ")";
             }else{
                 query += "scopeId:" + currentDso.getID();
             }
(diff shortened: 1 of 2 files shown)


terrywbrady:
Updated based on bug reported on earlier version.  I am puzzled that the following line does not appear to be a problem.  I would expect to need to repeat "id:"

- 
https://github.com/Georgetown-University-Libraries/DSpace/blob/3c50fc3a0d722b0aac2a96a60047112b9193e420/dspace-api/src/main/java/org/dspace/statistics/content/StatisticsDataVisits.java#L809

kshepherd:
@terrywbrady that line is OK as far as I can tell... (in a slight rush this morning so hopefully I am not missing anything!)

If we take a look at L809:
`query += " (id:" + dso.getID() + " OR " + ((DSpaceObjectLegacySupport) dso).getLegacyId() + ")";`

For id 1234 and legacyId 999 And this would result in:
`(id:123 OR 999)`

Which is fine, that's an OK way to do an "OR" search over a field (id). You could test this out in your local solr, just try eg. http://localhost:8080/solr/statistics/select?q=(id%3A1+OR+2)&wt=json&indent=true

Where '1' and '2' are two valid item ids and/or legacy ids.

Cheers

Kim

terrywbrady:
@kshepherd ,
Note the following query results

    q=+(owningItem:552886)&_=1498255388951&wt=json&rows=0} hits=558 
    q=+(owningItem:92eb9ef0-2591-48df-92f6-8603d1f8716b+)&_=1498255342731&wt=json&rows=0} hits=0 

If owningItem: is provided on both sides of the OR

    q=+(owningItem:92eb9ef0-2591-48df-92f6-8603d1f8716b+OR+owningItem:552886)&_=1498255357603&wt=json&rows=0} hits=558 

If owningItem is only provided on the left hand side

    q=+(owningItem:92eb9ef0-2591-48df-92f6-8603d1f8716b+OR+552886)&_=1498255333321&wt=json&rows=0} hits=34 

Note in the next example that id:552886 returns 34 records.  So, I suspect that a bare search term searches the "id" field.

## Note that the behavior is different for id 

    q=id:552886&_=1498255949598&wt=json&rows=0} hits=34 
    q=+(id:92eb9ef0-2591-48df-92f6-8603d1f8716b)&_=1498255290597&wt=json&rows=0} hits=2 

The resulting count is the same with or without id: on the right hand side of the OR

    q=+(id:92eb9ef0-2591-48df-92f6-8603d1f8716b+OR+552886)&_=1498255298004&wt=json&rows=0} hits=36 

    q=+(id:92eb9ef0-2591-48df-92f6-8603d1f8716b+OR+id:552886)&_=1498255313987&wt=json&rows=0} hits=36 

I will update my PR and provide an explanation of these findings.

tdonohue:
@kshepherd : Any chance you'd have time to give this another quick review/test?  There's still a bit of time to squeeze this into 6.1 if we can get some quick testers/approvers.

tomdesair:
+1 I've tested this PR and I can confirm that bitstream downloads with a legacy owningItem ID are visible in the UI.

terrywbrady:
I have verified the issue reported by Adan Roman: https://jira.duraspace.org/browse/DS-3602?focusedCommentId=56248&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-56248

This will require some additional development to resolve.

terrywbrady:
Removing work in progress based on https://jira.duraspace.org/browse/DS-3602?focusedCommentId=56488&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-56488

hardyoyo:
@kshepherd this is ready for testing, and since you have had an interest in this PR in the past, I wonder if you can re-test it again? Thansk!

hardyoyo:
Possible testing procedure: Vagrant-DSpace up a 5x-based instance, poke poke poke, verify there are a few stats in there, screencapture some proof for later, check out this PR, upgrade your 5x-based instance to 6x. Do the old stats you just made still "look the same?--use the screencaps you made to keep you honest.

tdonohue (approved):
👍 Code changes look good here. I haven't had a chance to test it though.

hardyoyo:
before:
dspr1782testing-before-2017-08-16 17-34-08

after:
dspr1782-after-2017-08-16 19-12-05

hardyoyo:
a better after (specific to the first handle, apples to apples comparison)
dspr1782after

hardyoyo:
better after, using the same handle for comparison: 
dspr1782-after

hardyoyo:
OK, so, stats "look the same" for my test, looks good to me. +1

hardyoyo (approved):

tdonohue:
Thanks for the testing, @hardyoyo!  Merging this for inclusion in 6.2

tdonohue:
Merged into dspace-6_x (as a squashed commit) and cherry-picked to `master`

PR #1783, opened 2017-06-21
Add Bitstream /content endpoint

abollini:
Add a /retrieve endpoint under the bitstreams to provide the capability to access to the actual content of a bitstream.
It is a very rough implementation, we need to add documentation and better support for authorization exception that now are simply re-thrown as runtime exception.

Support for HEAD operation to compute the checksum and if-none-match or if-modified-since request header should be also added.

Nevertheless, the code has been already applied to the demo rest 7 server as it was needed for the demonstration at OR2017

### DIFF
--- a/dspace-spring-rest/src/main/java/org/dspace/app/rest/BitstreamContentRestController.java
+++ b/dspace-spring-rest/src/main/java/org/dspace/app/rest/BitstreamContentRestController.java
@@ -1,0 +1,71 @@
+/**
+ * The contents of this file are subject to the license and copyright
+ * detailed in the LICENSE and NOTICE files at the root of the source
+ * tree and available online at
+ *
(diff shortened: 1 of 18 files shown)


abollini:
@tdonohue I have rebased the pr and changed the endpoint to get the bitstream content in ... /content as discussed in slack. The new code is running on the rest7 demo site.
The content endpoint is still named retrieve in the HAL document, I will open a pull request to change that but right now it is keep unchanged to avoid issue with the angular demo. It is also a proof that the HAL specification provide a nice way to abstract the client implementation from the real URL used this could be a strategy to support in future enpoint renaming keeping backward compatibility

tdonohue (approved):
👍 This looks good to me. Thanks, @abollini

PR #1793, opened 2017-07-04
DS-3579: Context mode and cache management for CLI commands

tomdesair:
This is the port of https://github.com/DSpace/DSpace/pull/1727 to the master branch.

I've also update the use of the Hibernate API in some places to prevent test failures.


tdonohue (approved):
👍 Looks good. Also compared first commit with #1727 and it looks to be the same changes.

PR #1794, opened 2017-07-04
DS-3563: Fix Oracle Flyway migration error

tomdesair:
Today we noticed that there is a bug in the Oracle script of https://github.com/DSpace/DSpace/pull/1706 😞  

The script worked when executed directly on a Oracle database, but fails through Flyway. I've now triple checked that it also works through Flyway.

This commit should also be cherry-picked to master and dspace-6_x.

Sorry for the inconvenience!


tdonohue (approved):
👍 Minor, obvious fix.

PR #1795, opened 2017-07-04
DS-3637 Fix links in Discovery and change order form method to GET.

### LINKED ISSUE #6986: [DS-3637] 404 response encountered by bots in Discovery.
<i>Imported from JIRA <a href="https://jira.lyrasis.org/browse/DS-3637">[DS-3637]</a> created by pablo.cda</i>
<p>Order links (href) in /discovery reference to non exisisting resources.</p>

<p>Example:</p>

<p><a class="" href="sort_by=dc.title_sort&order=asc"><br/>
 <span class="glyphicon glyphicon-ok btn-xs invisible"></span>Título Asc<br/>
</a></p>

<p>Despite this is the desired behavior (as javaScript use href's contents to build a form and submit it by POST), it causes issues with bots that follows these links and encounter a 404 response.<br/>
This issue affects both mirage and mirage2.</p>

<p>We propose to correct the links and make the query by GET, the same as faceted and filtered search links.</p>

<p> </p>

<p> </p>

PabloDeAlbu:
Description at [DS-3637](https://jira.duraspace.org/browse/DS-3637)
Fixes #6986

### DIFF
--- a/dspace-xmlui-mirage2/src/main/webapp/scripts/search-controls.js
+++ b/dspace-xmlui-mirage2/src/main/webapp/scripts/search-controls.js
@@ -252,7 +252,8 @@
                 $this = $this.find('a');
             }
             //Retrieve the params we are to fill in in our main form
-            params = $this.attr('href').split('&');
+            var separators = ['\\\&', '\\\?'];
+            params = $link.attr('href').split(new RegExp(separators.join('|'), 'g'));
 
             mainForm = $('#aspect_discovery_SimpleSearch_div_main-form');
             //Split them & fill in in the main form, when done submit the main form !
--- a/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/discovery/AbstractSearch.java
+++ b/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/discovery/AbstractSearch.java
@@ -239,7 +239,7 @@ public abstract class AbstractSearch extends AbstractDSpaceTransformer implement
         DSpaceObject dso = HandleUtil.obtainHandle(objectModel);
 
         //We set our action to context path, since the eventual action will depend on which url we click on
-        Division mainForm = searchDiv.addInteractiveDivision("main-form", getBasicUrl(), Division.METHOD_POST, "");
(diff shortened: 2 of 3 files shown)


J4bbi:
This is not related to #2045 as those are internal Solr queries.

It changes the main search form method to get from post as stated.

My problem is understanding why a method of using post has been chosen over get, generally.

It is generally desirable to be able to link search results, even though, as new items are added/deleted the results for a specific set of parameters will change accordingly.

I tend to focus on XMLUI but why it's standard I don't get.
https://github.com/DSpace/DSpace/blob/b0470eb59da0b43534de24021b72ea9970ab4d5a/dspace-xmlui-mirage2/src/main/webapp/xsl/core/navigation.xsl#L51

J4bbi:
Closing and opening to instigate checks. @kshepherd , @alanorth , @nwoodward  any thoughts?

alanorth:
Regarding the "shareability" of Discovery search result URLs, this seems to _only affect queries that use sort options_. I did a Discovery search for "Livestock" from the XMLUI homepage on my local DSpace 6.3 test instance and then used the gear icon to sort by "Title Asc".

Before:

> http://localhost:8080/discover

After:

> http://localhost:8080/discover?rpp=10&etal=0&query=Livestock&group_by=none&sort_by=dc.title_sort&page=1&order=asc

That matches the behavior seen in other Discovery result pages, for example facets and queries with search filters. Seems to be a positive change to match existing behavior. :+1:

Regarding the fixing of sort link targets, I confirm that this fixes the link targets of the sort options in the gear icon drop down. Here is the anchor tag for the "Title Asc" button before and after.

Before:

> `<a class="" href="sort_by=dc.title_sort&order=asc"><span class="glyphicon glyphicon-ok btn-xs invisible"></span>Title Asc</a>`

After:

> `<a class="" href="discover?rpp=10&etal=0&query=Livestock+farms&scope=/&group_by=none&sort_by=dc.title_sort&page=1&order=asc&filtertype_0=type&filter_relational_operator_0=contains&filter_0=Report"><span class="glyphicon glyphicon-ok btn-xs invisible"></span>Title Asc</a>`

In my brief testing I don't see anything broken. :+1:

PR #1804, opened 2017-07-11
DS-3889 Change Content-Type in OAI-PMH Response

saiful-semantic:
As per OAI 2.0 spec (3.1.2.1) the response content type *must* be text/xml.
http://www.openarchives.org/OAI/openarchivesprotocol.html#HTTPResponseFormat

Our OAI client is rejecting the response from DSpace OAI server due to wrong header in OAI response.

### DIFF
--- a/dspace-oai/src/main/java/org/dspace/xoai/controller/DSpaceOAIDataProvider.java
+++ b/dspace-oai/src/main/java/org/dspace/xoai/controller/DSpaceOAIDataProvider.java
@@ -94,7 +94,7 @@ public class DSpaceOAIDataProvider
             OutputStream out = response.getOutputStream();
             OAIRequestParameters parameters = new OAIRequestParameters(buildParametersMap(request));
 
-            response.setContentType("application/xml");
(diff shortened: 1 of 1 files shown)


hardyoyo (approved):
+1 by inspection, obviously correct

jcreel (approved):
The Content-Type returned for all OAI-PMH requests is now text/xml in the OAI webapp.

terrywbrady:
I did a scan of the OAI code base for "application/xml".  Should the following change as well?  https://github.com/DSpace/DSpace/blob/dspace-6_x/dspace/config/crosswalks/oai/metadataFormats/didl.xsl#L37-L42

kshepherd (approved):
+1, matches spec and a nice quick win.  Thanks @saiful-semantic !
@terrywbrady I think the DIDL crosswalk mime type does also need checking, but as per Slack discussion, let's keep this scoped to HTTP response for OAI PMH and check the crosswalks out separately

Note to self: this needs porting to 6.3

PR #1809, opened 2017-07-14
HAL link to the bitstream content renamed

abollini:
This PR remove the link "retrieve" from the bitstream resource in favor of "content" to keep the HAL link name aligned with the endpoint path.
It should be not merged before than the corresponding angular issue (https://github.com/DSpace/dspace-angular/issues/125 ) has been solved and the new code deployed on the demo angular ui site.

Please note that the angular ui site can be updated before to apply this PR to the rest demo site as the current code support access to the bitstream content using both link names (retrieve and content)


abollini:
This PR can be merged as the corresponding angular PR has been already approved & merged see https://github.com/DSpace/dspace-angular/pull/131

PR #1811, opened 2017-07-20
  [DS-3656]  DOIOrganiser CLI does not change the database anymore

ssolim:
Hibernate db connection add flush in uncache entity - master


ssolim:
I will close this as https://github.com/DSpace/DSpace/pull/1823/files is preferred solution

PR #1814, opened 2017-07-25
XMLUI/SwordClient: Fix typo in i18n key

AlexanderS:
The previous key was twice in the `messages.xml` and `xmlui.swordclient.DepositAction.package_format_error` one was not there but is used in [DepositAction.java](https://github.com/DSpace/DSpace/blob/dspace-6_x/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/swordclient/DepositAction.java#L30).


tdonohue (approved):
👍 Obvious change. Looks good.

PR #1817, opened 2017-08-02
[DS-3661] ImageMagick PDF Processing Degraded with Changes in 5.7 release

terrywbrady:
https://jira.duraspace.org/browse/DS-3661

### DIFF
--- a/dspace-api/src/main/java/org/dspace/app/mediafilter/ImageMagickThumbnailFilter.java
+++ b/dspace-api/src/main/java/org/dspace/app/mediafilter/ImageMagickThumbnailFilter.java
@@ -35,164 +35,162 @@ import org.dspace.core.ConfigurationManager;
  * thumbnail.maxwidth, thumbnail.maxheight, the size we want our thumbnail to be
  * no bigger than. Creates only JPEGs.
  */
-public abstract class ImageMagickThumbnailFilter extends MediaFilter
-{
-	private static int width = 180;
-	private static int height = 120;
+public abstract class ImageMagickThumbnailFilter extends MediaFilter {
+        private static int width = 180;
+        private static int height = 120;
         private static boolean flatten = true;
-	static String bitstreamDescription = "IM Thumbnail";
-	static final String defaultPattern = "Generated Thumbnail";
-	static Pattern replaceRegex = Pattern.compile(defaultPattern);
-	static String cmyk_profile;
-	static String srgb_profile;
-	
-	static {
-		String pre = ImageMagickThumbnailFilter.class.getName();
-		String s = ConfigurationManager.getProperty(pre + ".ProcessStarter");
-		ProcessStarter.setGlobalSearchPath(s);
-		width = ConfigurationManager.getIntProperty("thumbnail.maxwidth", width);
-		height = ConfigurationManager.getIntProperty("thumbnail.maxheight", height);
+        static String bitstreamDescription = "IM Thumbnail";
+        static final String defaultPattern = "Generated Thumbnail";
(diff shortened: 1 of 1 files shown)


alanorth:
Hard to see the code changes with all the whitespace noise...

terrywbrady:
@alanorth , try https://github.com/DSpace/DSpace/pull/1817/files?w=1 to simplify the review.  The whitespace was inconsistent so I let Eclipse normalize it.

alanorth:
Regarding the changes, you're correct to only get the image info if the cmyk and srgb profiles are defined 👍. Good catch. I take it that at GeorgeTown you've not defined those, so it is unnecessarily checking the image class every time?

It took me longer than it should have to read this, due to the non-functional changes to style and formatting. In my opinion commits should be concise and atomic—apply the functional changes first and then clean up the formatting in a separate commit. You should also describe your changes in the commit message(s). 😄

terrywbrady:
@alanorth , thanks for offering to test this.  This issue led to a significant performance issue in my production environment, so my creation of this PR was done in a bit of a scramble.  I appreciate your patience with the format changes and I agree that your suggestions would be the ideal way to create the PR.

We have the color profile option disabled in dspace.cfg.  We added several hundred large PDF's into our repository.  When the conversion was invoked on these files, the process appeared to serialize each page of the PDF as a file in /tmp even though we only needed the first page.  It took about 30 seconds per page to process.  When I bypassed the calls to org.im4java.core.Info, the thumbnail generation process ran significantly faster (50-100x faster),

alanorth:
@terrywbrady no problem, your solution looks logical. In fact, it seems obvious in hindsight that we'd only even _attempt_ to get the image info if we had the profile options enabled! We don't use very large PDFs so didn't notice this issue. I suspect we need to pass a page number to `Info()` like we do for the actual image thumbnailing.

Case in point, ImageMagick on the command line takes nearly one minute to return the results of this simple command:

```
$ identify -format '%r\n' ~/Downloads/DSpace-Manual.pdf | wc -l
833
```

And notice it returns 833 lines (the PDF is 833 pages!). Versus this, which returns instantly:

```
$ identify -format '%r\n' ~/Downloads/DSpace-Manual.pdf\[0\]
DirectClass sRGB Alpha
```

So your observation is correct! Maybe try to see if you can find a way to pass a page number to the `Info()` call.

mwoodiupui:
Tangential comment on formatting:  I tend to fix the formatting *first* so that I can read the code without getting a headache.  But I agree that it's best to split formatting and logic into separate commits.

terrywbrady:
I am still running ghostscript 8.7.  I cannot find any icc files in my ghostscript install directory, so this leaves me unable to fully test this process.

It would be good for someone with a newer version of ghostscript to resolve this issue.

Once this is resolved, the color profile change still needs to be ported to 6x and to master.  That will require additional testing.

terrywbrady:
I created a new ticket to address the underlying cause of this issue.  https://jira.duraspace.org/browse/DS-3664.  This PR can provide an immediate workaround for the performance issue.

alanorth:
I just tested this change on our DSpace 5.x instance and it works. To clarify, we're using the CMYK detection code and this commit does not affect that functionality, but _does_ make it more streamlined in that it avoids some unnecessary setup/tests for users who _aren't_ using the CMYK detection functionality. Looks good to me. 👍

@terrywbrady can you force push a new commit that has a nice description, though? Future DSpace developers will thank you when they see a nice description in the `git log`. 😄

tdonohue (approved):
👍 Looks good to me.

terrywbrady:
I see 2 +1's on this.  Merged...

PR #1818, opened 2017-08-03
XMLUI: Remove doubled translation key

AlexanderS:
This is just another quick cleanup of the messages.xml.

The key "xmlui.ChoiceLookupTransformer.lookup" is already in line 2368 of the
same file: https://github.com/DSpace/DSpace/blob/dspace-6_x/dspace-xmlui/src/main/webapp/i18n/messages.xml#L2368


helix84:
+1

helix84 (approved):

PR #1827, opened 2017-08-15
[DS-3674] copied over input-forms.xml to the test config folder

hardyoyo:
https://jira.duraspace.org/browse/DS-3674


mwoodiupui (approved):
+1 looks well, harmless.

PR #1831, opened 2017-08-25
DS-3682: Fix reusing of the same vocabulary dialog

AlexanderS:
https://jira.duraspace.org/browse/DS-3682

If the same vocabulary is used multiple times on the same input page, the
dialog is only fetched the first time from the server and reused afterwards.
The problem is, that the target input field is contained in the dialog. When
reusing the dialog the target field should be updated.

### DIFF
--- a/dspace-xmlui-mirage2/src/main/webapp/scripts/vocabulary-support.js
+++ b/dspace-xmlui-mirage2/src/main/webapp/scripts/vocabulary-support.js
@@ -36,6 +36,7 @@
             var vocabularyDialog = $('div#'+id);
             if(0 < vocabularyDialog.length){
                 //Open the modal
+                vocabularyDialog.find('input[type="hidden"][name="metadataFieldName"]').val(inputFieldName);
                 vocabularyDialog.modal('show')
             }else{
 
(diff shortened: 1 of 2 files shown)


terrywbrady:
@AlexanderS , I do not use controlled vocabularies, so I am looking for some advice on how to configure a test for this PR.  Could you add some set up instructions to the PR description?  The code looks good, but I would like to test this before approving the change.

AlexanderS:
@terrywbrady You simply need to add a vocabulary to two input fields in your `input-forms.xml`. Something like this from the default `input-forms.xml` file:

```xml
<field>
    <dc-schema>dc</dc-schema>
    <dc-element>subject</dc-element>
    <dc-qualifier/>
    <!-- An input-type of twobox MUST be marked as repeatable -->
    <repeatable>true</repeatable>
    <label>Subject Keywords</label>
    <input-type>twobox</input-type>
    <hint>Enter appropriate subject keywords or phrases. </hint>
    <required/>
    <vocabulary>srsc</vocabulary>
</field>
```

You will get a "Subject Categories" (or something like this) link below the input field. Just add a second field using the same vocabulary on the same page. You will notice, that (without the patch) all selections (from both links) will get into the field, you opened the popup first (do not click the "Add" button between using the two popups). The popup is fetched from the server only the first time and the target field is saved inside the popup.

Hope this helps to reproduce the issue.

terrywbrady (approved):
Reviewed code.  Tested in Mirage and Mirage2 themes

hardyoyo (approved):
+1 by inspection, tested in vagrant-dspace, works as expected

PR #1832, opened 2017-08-28
DS-3683 Check if workflow group is empty or not

jonas-atmire:
PR for 
https://jira.duraspace.org/browse/DS-3683

The following fixes that problem that a workflowgroup might be present, but might not actually contain anything, resulting in the advancing of a step to fail on authorization

### DIFF
--- a/dspace-api/src/main/java/org/dspace/workflow/WorkflowManager.java
+++ b/dspace-api/src/main/java/org/dspace/workflow/WorkflowManager.java
@@ -417,8 +417,7 @@ public class WorkflowManager
             // advance(...) will call itself if no workflow step group exists
             // so we need to check permissions only if a workflow step group is
             // in place.
-                if (wi.getCollection().getWorkflowGroup(1) != null && !wi.getCollection().getWorkflowGroup(1).isEmpty())
-            {
+            if (wi.getCollection().getWorkflowGroup(1) != null && !wi.getCollection().getWorkflowGroup(1).isEmpty()) {
                 // FIXME note:  authorizeAction ASSUMES that c.getCurrentUser() == e!
                 AuthorizeManager.authorizeAction(c, wi.getCollection(), Constants.WORKFLOW_STEP_1, true);
             }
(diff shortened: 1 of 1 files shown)


dspace-api/src/main/java/org/dspace/workflow/WorkflowManager.java
```diff
@@ -301,22 +301,27 @@ public static void claim(Context c, WorkflowItem wi, EPerson e)
         {
         case WFSTATE_STEP1POOL:
 
-            // FIXME note:  authorizeAction ASSUMES that c.getCurrentUser() == e!
-            AuthorizeManager.authorizeAction(c, wi.getCollection(), Constants.WORKFLOW_STEP_1, true);
-            doState(c, wi, WFSTATE_STEP1, e);
+                if (wi.getCollection().getWorkflowGroup(1) != null && !wi.getCollection().getWorkflowGroup(1).isEmpty())
+                {
+                    AuthorizeManager.authorizeAction(c, wi.getCollection(), Constants.WORKFLOW_STEP_1, true);
+                }                doState(c, wi, WFSTATE_STEP1, e);
```
AlexanderS: This formatting is strange. The `doState(...)` should be on a new line.

dspace-api/src/main/java/org/dspace/workflow/WorkflowManager.java
```diff
@@ -412,7 +417,7 @@ public static boolean advance(Context c, WorkflowItem wi, EPerson e,
             // advance(...) will call itself if no workflow step group exists
             // so we need to check permissions only if a workflow step group is
             // in place.
-            if (wi.getCollection().getWorkflowGroup(1) != null)
+                if (wi.getCollection().getWorkflowGroup(1) != null && !wi.getCollection().getWorkflowGroup(1).isEmpty())
```
AlexanderS: Why did you change the indentation?

jonas-atmire: @AlexanderS Good catch, I hadn't noticed this since I had my IDE configured to ignore whitespace/new line changes (And I'd copied this code from a test instance, incorrectly checking the indentation)

PR #1834, opened 2017-08-31
Vsim 8 carousel

hardyoyo:
theme tweaks for VSim, including a new tour carousel, image buttons on the home page, and the beginning of the featured collection blurb on the home page


hardyoyo:
whoops, wrong fork :-\

PR #1836, opened 2017-09-05
DS-3687

Generalelektrix:
Changed hard coded string for reference to a new field in language bundle.


terrywbrady (changes requested):
@Generalelektrix , thank you for submitting this fix.  This text also appears in the XMLUI.  Is there a way to name the I18n field so that it could be referenced by either XMLUI or JSPUI?  I will be happy to test your solution in XMLUI.

Generalelektrix:
I'm not using xmlui here, in which file is the resource bundle for xmlui located?

terrywbrady:
@Generalelektrix , the XMLUI file is here: https://github.com/DSpace/DSpace/blob/dspace-6_x/dspace-xmlui/src/main/webapp/i18n/messages.xml

Unfortunately, this would create a separate key for the same value.  I wonder if there is another resource bundle for dspace-api.  I will post a question to Slack to see if someone can comment here.

Generalelektrix:
Maybe we could just give the value a more generic key name.

terrywbrady:
@Generalelektrix , that makes sense to me.  See the following conversation with @tdonohue.

Terry Brady [1:22 PM] 
See https://github.com/DSpace/DSpace/pull/1836 .  What is the proper way to manage the display text referenced in this PR?  This value will display in both XMLUI and JSPUI.


Tim Donohue [1:33 PM] 
@terrywbrady : unfortunately, it has to be done separately for JSPUI vs XMLUI, as they use different i18n keys/settings.  JSPUI uses Messages.properties, while XMLUI uses messages.xml, and there's no relationship between these two "messages" files


[1:35] 
Although, hmm...in this case you are using shared API-level code.


[1:36] 
So, it might just use Messages.properties only


[1:36] 
Have you tested this PR? It might "just work" in XMLUI...it could be that the key should just be named something more generic (i.e. not start with "jspui.") (edited)


Terry Brady [1:38 PM] 
I have not tested it.  I did not realize that all of the keys in the resource bundle were specific to a single UI which tells me that my initial solution may have been a bit messy.


Tim Donohue [1:40 PM] 
Unfortunately, my answer may have been misleading cause it is confusing.  JSPUI always only uses Messages.properties.... however XMLUI will use messages.xml (for i18n in the dspace-xmlui) and sometimes Messages.properties (if the i18n occurs at the dspace-api layer)


[1:40] 
So, it's a bit of a mess


[1:41] 
My "guess" for this 1836 PR is that the PR might be OK mostly as-is, but the key should be named something like `org.dspace.statistics.content.StatisticsDataVisits.legacy = (legacy)`


Terry Brady [1:41 PM] 
Thanks. I will add this conversation to the PR.


Tim Donohue [1:42 PM] 
(As this key likely is NOT JSPUI specific...and you'll see other keys named like this in the Messages.properties near the bottom)

AlexanderS:
@Generalelektrix You may want to remote this `.metadata` stuff from eclipse.

Generalelektrix:
Yeah, I commited those files by mistake and I'm trying to remove them, but I'm new to git (mainly used SVN and TFS here) and I'm working on it.

Generalelektrix:
I think I've corrected everything, should be ready for review now.

terrywbrady (approved):
Changes look good.  I will run a test with XMLUI to verify.

terrywbrady:
+1, verified that the label is visible in XMLUI.

hardyoyo (approved):
+1 by review

PR #1841, opened 2017-09-20
DS-3696: Absolute links in root resource

tomdesair:
This is a fix for https://jira.duraspace.org/browse/DS-3696


abollini (changes requested):
This looks to me unnecessary complicate, it should be possible to get the base URL in a simpler way:
https://stackoverflow.com/questions/6271075/how-to-get-the-base-url-from-jsp-request-object
Anyway, the REST base URL should be calculated only one time outside the loop

tomdesair:
Hi @abollini, you are right. The solution you proposed is better. I've adjusted the code.

Thanks for reviewing!

abollini (approved):

PR #1842, opened 2017-09-20
DS-3698 Correct HAL browser context path

tomdesair:
This is a fix for https://jira.duraspace.org/browse/DS-3698

### DIFF
--- a/dspace-spring-rest/src/main/webapp/index.html
+++ b/dspace-spring-rest/src/main/webapp/index.html
@@ -291,7 +291,7 @@ Content-Type: application/json
 <script>
     var browser = new HAL.Browser({
         container: $('#browser'),
-        entryPoint: '/dspace-spring-rest/api'
+        entryPoint: window.location.pathname + 'api'
     });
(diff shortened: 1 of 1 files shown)


abollini (approved):
I'm fine with the proposed code. I will make a try late this week but if other can make a fast run to double check and it works feel free to merge. 
Please note that this is related to an issue in the Spring HAL library that was reported to the spring community and could be unnecessary in the long run: https://jira.spring.io/browse/DATAREST-971 the latest comment here also suggest to investigate a bit more in our configuration

tomdesair:
Indeed, it seems to be related to the issue you mention. Another user mentions the `HalBrowserMvcEndpoint` class in that ticket, but I'm not sure how to use it. So for now, I think the JavaScript workaround is the easiest solution.

mwoodiupui:
+1  I just now tested this, after spending too much time trying to figure out why I always got 404 in the HAL Browser but 'curl' worked as expected, and then finding that someone else had already fixed it.  Thanks!

PR #1848, opened 2017-09-22
DS-3700: MediaFilterServiceImpl forgot to close an input stream.

pnbecker:
https://jira.duraspace.org/browse/DS-3700
This needs to be cherry-picked to master as well.

The fastest method to reproduce this is to lower the hard limit of allowed file descriptors. This can be done under linux by using the shell built-in command ulimit: `ulimit -Hn 220`. The exact limit my differ depending on your shell, your configuration and your system (I was able to reproduce the problem with a hard limit of 220).

Ulimit knows a soft and a hard limit. The soft limit can be changed by every process, the hard limit cannot be changed. I needed to change the hard limit and then run the `[dspace]/bin/dspace filter-media -v` on a repository containing several bitstreams: `ulimit -Hn 220 ; ~/dspace/bin/dspace filter-media -f -v'.

The only simple way I found to enhance a previously lowered hard limit was by closing the current shell and reopening a new one (I didn't looked for any other method). So start from a higher value and reduce it until you found a number, that fits your system. And don't forget to start a new shell session after testing, as ulimit limits the file descriptors for each process started by this shell session.

### DIFF
--- a/dspace-api/src/main/java/org/dspace/app/mediafilter/MediaFilterServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/app/mediafilter/MediaFilterServiceImpl.java
@@ -21,6 +21,7 @@ import org.dspace.services.ConfigurationService;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.annotation.Autowired;
 
+import java.io.IOException;
 import java.io.InputStream;
 import java.util.*;
 
@@ -314,12 +315,10 @@ public class MediaFilterServiceImpl implements MediaFilterService, InitializingB
         // get bitstream filename, calculate destination filename
         String newName = formatFilter.getFilteredName(source.getName());
 
-        Bitstream existingBitstream = null; // is there an existing rendition?
-        Bundle targetBundle = null; // bundle we're modifying
-
-        List<Bundle> bundles = itemService.getBundles(item, formatFilter.getBundleName());
-
         // check if destination bitstream exists
+        Bundle existingBundle = null;
+        Bitstream existingBitstream = null;
+        List<Bundle> bundles = itemService.getBundles(item, formatFilter.getBundleName());
         if (bundles.size() > 0)
         {
             // only finds the last match (FIXME?)
@@ -328,13 +327,12 @@ public class MediaFilterServiceImpl implements MediaFilterService, InitializingB
 
                 for (Bitstream bitstream : bitstreams) {
                     if (bitstream.getName().equals(newName)) {
-                        targetBundle = bundle;
+                        existingBundle = bundle;
                         existingBitstream = bitstream;
                     }
                 }
             }
         }
-
         // if exists and overwrite = false, exit
         if (!overWrite && (existingBitstream != null))
         {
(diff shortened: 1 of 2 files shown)


pnbecker:
I'm still testing it myself. But the problem looked so obvious to me, that I already created the PR.

dspace-api/src/main/java/org/dspace/app/mediafilter/MediaFilterServiceImpl.java
```diff
@@ -352,10 +353,20 @@ public boolean processBitstream(Context context, Item item, Bitstream source, Fo
...
+            destStream = formatFilter.getDestinationStream(item, srcStream, isVerbose);
+            if (srcStream != null)
+            {
+                try {
+                    srcStream.close();
+                } catch (IOException ex)
+                {
+                    // nothing to be done here.
+                }
+            }
```
tomdesair: The above if-block can be replaced by org.apache.commons.io.IOUtils#closeQuietly(java.io.InputStream)

dspace-api/src/main/java/org/dspace/app/mediafilter/MediaFilterServiceImpl.java
```diff
@@ -366,6 +377,11 @@ public boolean processBitstream(Context context, Item item, Bitstream source, Fo
             }
         } catch (OutOfMemoryError oome) {
             System.out.println("!!! OutOfMemoryError !!!");
+            try {
+                if (destStream != null) {
+                    destStream.close();
+                }
+            } catch (IOException ex) {}
```
tomdesair: The above try-block can be replaced by org.apache.commons.io.IOUtils#closeQuietly(java.io.InputStream)

dspace-api/src/main/java/org/dspace/app/mediafilter/MediaFilterServiceImpl.java
```diff
@@ -382,10 +398,21 @@ public boolean processBitstream(Context context, Item item, Bitstream source, Fo
...
+        // The input stream was stored in the appropriate asset store, close it to avoid socket leaks.
+        if (destStream != null)
+        {
+            try {
+                destStream.close();
+            } catch (IOException ex)
+            {
+                // we cannot recover from not being able to close the input stream, so ignore the exception.
+            }
+        }
```
tomdesair: The above if-block can be replaced by org.apache.commons.io.IOUtils#closeQuietly(java.io.InputStream)

tomdesair (changes requested):
Looks good, I only added some suggestions for code cleanup.

dspace-api/src/main/java/org/dspace/app/mediafilter/MediaFilterServiceImpl.java
mwoodiupui: Try-with-resources might also work, and is even less noisy.

tomdesair: That's correct since we can assume Java 7 now.

pnbecker:
While try with resources is a cleaner approach I had to change more code to apply it here. In the meantime I found another socket leak in DSBitStoreService which I fixed while being on it.

This PR ensures that we close three streams we haven't closed before. While I don't see anymore unclosed streams in this code area, I can still bring our filter-media script to an IOException: too many open files. Nevertheless it is a big progress and we should merge this PR.

If someone else has a chance to take another look unclosed sockets, I would appreciate that.

mwoodiupui:
'mvn findbugs:findbugs' on dspace-api turned up 28 possible unclosed streams.  We should perhaps have a separate issue for a general unclosed-streams bug hunt.

In total, findbugs noted over a thousand constructions that it disliked.

pnbecker:
@mwoodiupui pointing towards findbugs is a good idea. I would prefer to do as you suggested and try to fix this in a separate issue. What do you think of this PR as it is for now?

hardyoyo (approved):
+1 by inspection

hardyoyo:
@tomdesair I think @pnbecker has made the changes you requested, can you confirm and resolve your review? Thanks!

kshepherd:
+1
i tested using @pnbecker 's instructions, I had to use a hard limit of 215 on my particular system and number of bitstreams, but managed to reproduce this error on dspace-6_x:
ERROR filtering, skipping bitstream #7caffaf3-7845-4098-99b8-184539337e99 java.io.IOException: java.io.FileNotFoundException: /home/vagrant/dspace/assetstore/37/71/23/37712310757983322376345532200733929853 (Too many open files)

And ran successfully with same limit using this PR.

Thanks @pnbecker !

tomdesair (approved):

tomdesair:
I've merged this PR and created a cherry-pick PR for the master branch here: https://github.com/DSpace/DSpace/pull/1963

PR #1854, opened 2017-10-03
[DS-3704] Expand DSpace REST Reports to include bitstream fields in item listing

terrywbrady:
See https://jira.duraspace.org/browse/DS-3704 for screenshots illustrating this new feature.


kshepherd (approved):
Tested using instructions in https://github.com/terrywbrady/restReportTutorial/blob/master/demo5/README.md and was able to see additional bitstream columns as advertised.
+1, works well, thanks @terrywbrady !

tdonohue (approved):
👍 Looks good to me!

PR #1857, opened 2017-10-04
[DS-2670] Restrospectively create DOIs for all items in archive

### LINKED ISSUE #6030: [DS-2670] Restrospectively create DOI's for alle items in archive
<i>Imported from JIRA <a href="https://jira.lyrasis.org/browse/DS-2670">[DS-2670]</a> created by cjuergen</i>
<p>It should be possible to retrospectively add DOI's to all the in_archive items in a DSpace instance. At the moment only new Items (after enabling the identifier service) get a DOI.</p>

<p>It is possible to create a doi for a single item with the doi-organiser, but this seems rather a bug see <a href="https://jira.lyrasis.org/browse/DS-2669" title="Inconsistent behaviour of doi-organiser" class="issue-link" data-issue-key="DS-2669"><del>DS-2669</del></a></p>

mwoodiupui:
This curation task will create *any* identifiers which are configured on but not present in a given Item.
Briefly tested a version for DSpace 5.6.  Please inspect, test, critique.

Fixes #6030


hardyoyo (approved):
+1 by inspection, I love this idea, and it'll come in handy for anyone standing up a data repository

pnbecker:
Just did a rough code inspection. I love it.
We need to test it with:

- HandleIdentifierProvider
- VersionedHandleIdentifierProvider
- VersionedHandleIdentifierProviderWithCannonicalHandles
- DOIIdentifierProvider and
- VersionedDOIIdentifierProvider

mwoodiupui:

mwoodiupui:
Close/reopen to kick stuck tests.

pnbecker (commented):
The code relies that `IdentifierProvider` do not register a new identifier if an item already has an identifier (of the kind the provider supports). All `IdentifierProvider` that I know in DSpace behave like this and the Javadoc tells people to do that: https://github.com/DSpace/DSpace/blob/main/dspace-api/src/main/java/org/dspace/identifier/IdentifierProvider.java#L57. I would prefer if we could check if an Item has an Identifier before registering it, but I doubt that this is possible with reasonable effort. The curation task creates all kind of identifiers, not only DOIs. The name of the curation task is reflecting this correctly.

I'm +1 on the code. I would like to give it a test run, and hope to be able to do this next week. If anyone is quicker with a test, I would appreciate it.

pnbecker:
@kshepherd Can you please give this a test run and merge it then?

kshepherd:
This is a great simple script. I've tested it with handles and DOIs in different conditions to understand the results, and the code being 'tested' is really the identifier service / provider register() behaviour, but it's good to know before running it over a large set of items in prod. As noted below, I did find some issues with the versioned handle provider with canonical handles (where the handle is rewritten to point to the newest version, and the old ones get the x.1 and x.2 handles etc) that means things get a bit messed up when the task is run over old versions of items that have handles.

**Handle testing**

Worked well for `HandleIdentifierProvider` and `VersionedHandleIdentifierProvider` - if already present, metadata deleted and re-added which is nice for prefix updates too?
Deleted handle from metadata but not DB, and was added again.
Using the standard `VersionedHandleIdentifierProvider`, the versioned handle for the newest item was correctly recreated after deletion, but not for older item versions...
Running the script over these items in typical circumstances where they already had handles / versioned handles didn't result in any problems or duplicate identifiers.

`VersionedHandleIdentifierProviderWithCanonicalHandles` did show some problems. I made a new version of an item and ran the script without any other changes. When it hit the 23.1 handle (the oldest version, since canonical for new version was updated 23 at the time of versioning), it acted almost as though I'd made a new version, assigning a new 23.2 handle to the *newest* item and updating 23 to the oldest, which now has 23.1 and 23 pointing to it:

```
28 | 123456789/23.2 | 2 | | 942c31dd-6b6d-4b86-a987-1ef98d12397a
29 | 123456789/23.1 | 2 | | c16970f3-d28a-40d6-bacd-f23f3eb884a2
27 | 123456789/23   | 2 | | c16970f3-d28a-40d6-bacd-f23f3eb884a2
```

I don't know if this needs to be a blocker for this script, or just a note not to run it on items with versions under the `VersionedHandleIdentifierProviderWithCanonicalHandles` provider. Perhaps in the task itself, the handle identifier provider service could be looked up, and if the class matches the Canonical provider, exit the curation task with a failure message? 

It seems to me that the `VersionedHandleIdentifierProviderWithCanonicalHandles` provider needs to change some assumptions about what to do when register() is called, because it won't always be in the context of a new version. But that's probably out of scope for this PR.

**DOI testing**

Worked for DOIs and versioned DOIs and also respected item filter. If DOI already present, metadata deleted and re-added (again, nice if eg dx.doi.org -> doi.org)
If metadata deleted but in DB, the correct DOI URI is added back to metadata.

If DOI is missing from the DB but a DOI URI kept in metadata, a new DOI was queued for registration.
This is to be expected, and isn't really a problem, but I'm just noting it here.

(perhaps in the future the DOI providers could look in metadata and if it finds a valid DOI URI of the prefix managed by DSpace, put it in the DB with "already reg'd" status?)

In the particular case I tested, it was a versioned DOI and the newly minted DOI was ../10 rather than ending in a .2 or .3 as with other versioned identifiers, but that's not a big deal either as far as i can tell, just aesthetic.

dspace-api/src/main/java/org/dspace/ctask/general/CreateMissingIdentifiers.java
```diff
@@ -0,0 +1,78 @@
...
+            context = Curator.curationContext();
+        } catch (SQLException ex) {
+            report("Could not get the curation Context:  " + ex.getMessage());
+            return Curator.CURATE_ERROR;
+        }
+
+        // Find the IdentifierService implementation
+        IdentifierService identifierService = new DSpace()
+                .getServiceManager()
+                .getServiceByName(null, IdentifierService.class);
```
kshepherd, line 60: This could use the newer pattern
`IdentifierService identifierService = IdentifierServiceFactory.getInstance().getIdentifierService()`
But it's functionally the same so not really an issue.

mwoodiupui:
So, is this ready to go?

kshepherd:
> So, is this ready to go?

I'm +1 as long as we really clearly explain that this will be destructive if run over repositories using `VersionedHandleIdentifierProviderWithCanonicalHandles` - it will reassign versioned handles incorrectly and cause some items to have multiple handles. I think ideally we'd check for that particular provider somehow and exit the script, but perhaps it can be handled in documentation? I'm not trying to be a blocker, just a bit worried about that scenario.

kshepherd (approved):
Escape looks good! +1 and merging

PR #1860, opened 2017-10-05
DS-3710 Fix ehcache config conflict

aprilherron:
This fix is intended to solve the issue with conflicting users accessing "DSpaceHibernateCache" in a shared temp directory.


hardyoyo (approved):
tested in our staging environment, with multiple DSpace instances, for each instance, I can now see multiple ehcache_auto_created_????????????? folders under each /usr/local/dspace-instance?/temp folder, which is perfect. +1, everything works as expected for each instance

kshepherd:
+1, quick win

PR #1863, opened 2017-10-05
[DS-3713] REST Query/Collection Report - Bug Filtering for Bitstream Permissions

terrywbrady:
https://jira.duraspace.org/browse/DS-3713

The following tutorial page will provide instructions on how to replicate this error.
https://github.com/terrywbrady/restReportTutorial/blob/master/demo4/README.md


pnbecker (approved):
Obvious fix.

kshepherd:
+1, tested this and could reproduce the issue where restricted ORIGINAL bitstreams did not show up, after applying this PR the filter displays them correctly

PR #1864, opened 2017-10-06
Ds 3629   listing of all groups misses pagination - XMLUI

ssolim:
This small fix enables again the pagination in admin/groups


hardyoyo (approved):
+1 obviously correct

kshepherd (approved):
+1 by inspection, good quick bugfix
this needs partial port to master (GroupServiceImpl, not the xmlui stuff)

terrywbrady:
the non-xmlui change was already on master

PR #1870, opened 2017-10-23
DS-3724: Missing readonly in input-forms.dtd

AlexanderS:
https://jira.duraspace.org/browse/DS-3724

The `input-forms.dtd` is missing the `readonly` element, that is already handled by JSPUI and XMLUI.

### DIFF
--- a/dspace/config/input-forms.dtd
+++ b/dspace/config/input-forms.dtd
@@ -14,7 +14,7 @@
  <!ATTLIST form name NMTOKEN #REQUIRED>
  <!ELEMENT page (field)+ >
  <!ATTLIST page number NMTOKEN #REQUIRED>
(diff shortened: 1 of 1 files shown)


terrywbrady (approved):
+1, code review

tdonohue (approved):
Looks correct to me, but I haven't tested to ensure the DTD validation works now.

mwoodiupui (approved):
+1 briefly tested, seems to work.

mwoodiupui:
Should be ported to DSpace 7, but note that these configuration files are being refactored there.

AlexanderS:
I have created 4Science/DSpace#19 to include this change in the pull request refactoring the config files for DSpace 7.

PR #1874, opened 2017-10-26
DS-3729: Set the Bitstream deletion flag in the database

marsaoua:
https://jira.duraspace.org/browse/DS-3729

Needs to be cherry-picked to master too.

### DIFF
--- a/dspace-api/src/main/java/org/dspace/content/BundleServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/content/BundleServiceImpl.java
@@ -191,11 +191,15 @@ public class BundleServiceImpl extends DSpaceObjectServiceImpl<Bundle> implement
             bundle.unsetPrimaryBitstreamID();
         }
 
-        // Check if we our bitstream is part of a single bundle:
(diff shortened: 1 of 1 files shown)


pnbecker (approved):
~~+1 This PR comes from my colleague, so other committers should review that.~~
Travis is complaining, obviously something is still wrong here. We'll take a close look. -> marked as work in progress.

pnbecker:
@tomdesair You were right about the order of cleaning a bitstream's bundle and deleting it. Nevertheless there was a bug related to this. We added a code comment that should help understand the bug and our change.

I'm +1, but as noted before: it comes from my colleague, so I would appreciate more reviews.

tomdesair:
+1

Makes sense and validated by Travis that the delete (in the unit tests) still works.

We shouldn't forget to port this to master.

hardyoyo (approved):
+1 by review, looks fine and passes tests

pnbecker:
@hardyoyo thank you for merging. Did you cherry-picked this to master too or shall I?

kshepherd:
apparently this is already ported to master, removing label

PR #1877, opened 2017-10-30
DS-3734 Fixed missing trigger of item last modified date when adding a bitstream

christian-scheible:
https://jira.duraspace.org/browse/DS-3734

When a bitstream is added to an item after it was archived the last modified date of the owning item is not triggered.

To reproduce the error:

Go to http://demo.dspace.org/xmlui/
Log in as administrator
Click on any item and hit "Edit this Item"
Click on "Item Bitestreams"
Click on "Upload a new Bitstream"
Upload the file
Click on Item Status
Expected behavior: Last Modified equals current date and time.
Observed behavior: Last Modified didn't change

### DIFF
--- a/dspace-api/src/main/java/org/dspace/content/BundleServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/content/BundleServiceImpl.java
@@ -147,6 +147,14 @@ public class BundleServiceImpl extends DSpaceObjectServiceImpl<Bundle> implement
                 return;
             }
         }
+        
+        //Ensure that the last modified from the item is triggered !
+        Item owningItem = (Item) getParentObject(context, bundle);
+        if(owningItem != null)
+        {
+            itemService.updateLastModified(context, owningItem);
+            itemService.update(context, owningItem);
+        }
 
         bundle.addBitstream(bitstream);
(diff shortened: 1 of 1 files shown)


AlexanderS:
Does this also update the last modified date when f.e. a thumbnail is created?

christian-scheible:
@AlexanderS I am not sure about this as we don't use thumbnails in our DSpace instance. But a far as I understand this, a thumbnail is a bitstream in the bundle THUMBNAIL. In this case adding a thumbnail to an item will change the date which is consistent with removing it. Because at `org.dspace.content.BundleServiceImpl.removeBitstream(Context, Bundle, Bitstream)` we already  trigger the update of the item last modified date.

terrywbrady (approved):
Reviewed code, Verified change, per Jira did a quick test for performance impacts when generating derived bitstreams.  The impact seems small compared to the time it takes to create derivative bitstreams.

My approval assumes that we do not want to explore setting bundle-specific behaviors at this time.  If we decide that is needed in the future, a new Jira issue should be logged.

kshepherd (approved):
+1, tested successfully, merging now
thanks @christian-scheible !

PR #1880, opened 2017-11-02
[VSIM-90] use get not post

hardyoyo:
people like to bookmark and share search results, we should let them


hardyoyo:
whoops, wrong upstream, sorry!

PR #1872, opened 2017-10-25
DS-3728 Avoid duplicate metadata on bitstream clones

jonas-atmire:
PR for
https://jira.duraspace.org/browse/DS-3728

Avoid duplicating bitstream metadata when creating a new version

PR #1883, opened 2017-11-09
DS-3702 & DS-3703: Rebuild the old behavior of bitstreams during vers…

marsaoua:
…ioning in DSpace 6. Bitstreams are pointing to the same file on the disk and reuse the same internal id of the predecessor item's bitstreams version as long as the new bitstreams do not differ from them. Bitstream's metadata are also duplicated only for the new version of the item.

### DIFF
--- a/dspace-api/src/main/java/org/dspace/content/BitstreamServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/content/BitstreamServiceImpl.java
@@ -96,6 +96,37 @@ public class BitstreamServiceImpl extends DSpaceObjectServiceImpl<Bitstream> imp
     }
 
     @Override
+    public Bitstream clone(Context context, Bitstream bitstream)
+            throws SQLException 
+    {
+        // Create a new bitstream with a new ID.
+        Bitstream clonedBitstream = bitstreamDAO.create(context, new Bitstream());
+        // Set the internal identifier, file size, checksum, and 
+        // checksum algorithm as same as the given bitstream. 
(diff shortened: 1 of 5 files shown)


tomdesair:
I did not review this yet. I just thinks this relates to https://github.com/DSpace/DSpace/pull/1872.

pnbecker:
Should also be cherry-picked to master.

jcreel (approved):
I deployed tuub:DS-3702_AND_DS-3703 under XMLUI and versioned a couple of items with multiple bitstreams.  I found that the new versions of the items had the same bitstream listings, save that the UUIDs were different.  I could see the differing UUIDs for different versions' bitstreams by navigating to the bitstreams in the edit item UI.  Looking up the different UUIDs for a given bitstream in the database, I found that they all shared the same internal_id and, ergo the same file in the assetstore.

kshepherd (approved):
+1 tested successfully, works as advertised.
just a note: the braces placement in this commit doesn't conform to new dspace code style so it might need some work during the port to master

kshepherd:
@hardyoyo : This PR has 2 successful tests, if you could give the changes a quick inspection and add a +1 if you are happy, that will let us merge :) Thanks!

hardyoyo (approved):
+1 looks good to me

pnbecker:
@hardyoyo did you cherry-picked this to master too?

pnbecker:
@hardyoyo thank you for merging. Did you cherry-picked this to master too or shall I?

PR #1886, opened 2017-11-16
[DS-3757] increase default clamav socket timeout to 6 minutes

hardyoyo:


hardyoyo:
https://jira.duraspace.org/browse/DS-3757

mwoodiupui (approved):
I'm not sure what the value should be, but 6 minutes seems much more realistic than <1/8 second.

tdonohue (approved):
👍 LGTM

PR #1888, opened 2017-11-18
DS-3740 Endpoint to create new submissions

abollini:
Second step toward the implementation of the workspaceitems endpoint.
This PR needs to be reviewed after #1887


abollini:
in the DSpace 7 meeting it was decided to just review the final PR #1889 using integration test and curl looking to the doc.
https://wiki.duraspace.org/display/DSPACE/2017-11-30+DSpace+7+Working+Group+Meeting+notes

PR #1896, opened 2017-12-11
[DS-3087] removed inlineMath setting from MathJax config in Mirage1/2

hardyoyo:
the MathJax defaults are sensible and preferable, this PR replaces #1608, which was closed after it was observed to target the wrong branch


kshepherd (approved):
+1, small and sensible fix, reverting this delimiter back to the MathJax defaults

tdonohue (approved):
👍 LGTM

tomdesair:
+1

We're running this in production for a client as using `$` can give weird results when talking about dollars in your abstract (e.g. https://research-repository.griffith.edu.au/handle/10072/339488)

`['\\(','\\)']` is the default, so it can indeed be removed.

PR #1897, opened 2017-12-12
DS-3787 Impossible to retrieve the community or collection logo

abollini:
Trying to access the logo of a community or collection results in a 500 error code.
After investigation I found that the issue is related to the absence of a dc.title for such bitstreams that is used to store the name of the file. The Bitstream endpoint (MultipartFileSender) want to force a name in the response header so flag the response as not valid in absence of the file name

I have also added tests to show the bug see the result of the build of the first commit https://travis-ci.org/4Science/DSpace/builds/315096151#L5562

nevertheless, we now have a failure on the findAllBitstreamTest...

### DIFF
--- a/dspace-spring-rest/src/main/java/org/dspace/app/rest/BitstreamContentRestController.java
+++ b/dspace-spring-rest/src/main/java/org/dspace/app/rest/BitstreamContentRestController.java
@@ -24,6 +24,7 @@ import org.dspace.app.rest.utils.MultipartFileSender;
 import org.dspace.authorize.AuthorizeException;
 import org.dspace.authorize.service.AuthorizeService;
 import org.dspace.content.Bitstream;
+import org.dspace.content.BitstreamFormat;
 import org.dspace.content.service.BitstreamService;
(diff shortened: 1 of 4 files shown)


abollini:
@tomdesair  can you help with the test failure? what I'm missing?

tomdesair:
Hi @abollini, there was a problem with deleted bitstreams. The logo's are marked as "deleted" but not removed from the database. In the `BitstreamBuilder` class, we do expunge bitstreams after deleting them. But since the logo is created by DSpace API (and not via a `BitstreamBuilder`), it is  not expunged automatically.

Therefore I've added an "expunge deleted bitstreams" step in the test cleanup method here https://github.com/4Science/DSpace/pull/22

In the future, this will also catch bitstreams created by other processes like for example filter-media.

abollini:
thanks @tomdesair can you formally approve the PR once travis confirm that it works?

tomdesair (approved):
Tests prove that this works now. Good catch and nice job @abollini !

PR #1901, opened 2017-12-20
DS-3791 : Missing date values while faceting

jonas-atmire:
Fixes #7138
Make sure the "yearDifference" takes into account that a gap of 10 year contains 11 years


bram-atmire (approved):
reviewed by inspection & ok

kshepherd (approved):
+1, tests OK. Thanks @jonas-atmire !

PR #1902, opened 2017-12-20
[DS-3556] utf8 normalization for oai interface

pbroman:
NKFC normalization of metadata values in the oai core to solve DS-3556
https://jira.duraspace.org/browse/DS-3556


pbroman:
Closed in favor of a solution with rollback to Xalan 2.7.0

PR #1905, opened 2017-12-21
DS-3794 issue when pressing enter in submission lookup popup window

samuelcambien:
In the ChoiceLookupTransformer, when creating the interactive div, it should be given an action, to prevent the default 'submit' action.
I gave it the 'DSpaceChoiceLoad() action, which takes a form as argument.
I also refactored the choice-lookup.js file, basically:

* in the DSpaceChoiceLookup function: copying the input values to the popup at the time the popup is opened
* in the DSpaceChoiceLoad function: reading the value from the popup input (rather than from a param, which never worked and then from the submission form, thus ignoring the popup input text)
* in the DSpaceChoiceLoad function: deleting some unused variables
* in the DSpaceChoiceLoad function: don't increase the start and end of the results, because this shouldn't be done when altering the search term. Also renamed the 'oldStart' and 'newStart' variables to 'start' and 'end' for clarity.
* in the DSpaceChoicesMoreOnClick function: rather increase the pagination of the results here

### DIFF
--- a/dspace-xmlui-mirage2/src/main/webapp/scripts/choice-support.js
+++ b/dspace-xmlui-mirage2/src/main/webapp/scripts/choice-support.js
@@ -145,6 +145,35 @@ function DSpaceChoiceLookup(url, field, formID, valueInput, authInput,
             $( "body" ).append(modal);
             modal.modal();
             var form = document.getElementById('aspect_general_ChoiceLookupTransformer_div_lookup');
+
+            // copy the value from the input form to the popup window
+            var of = $(window.document).find('#' + formID);
+
+            if (isName) {
+                $('*[name = text1]').val(
+                    of.find('*[name = ' + dspace_makeFieldInput(valueInput, '_last') + ']').val()
+                );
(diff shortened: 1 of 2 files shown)


samuelcambien:
also committed to dspace 5

https://github.com/DSpace/DSpace/pull/1904

qzhao-ulb:
I got the same problem by using XMLUI. Opened a lookup dialog in submission, typed a keyword in the help text input area, then entered "enter" button, then all input values in other fields in the submission dialog were deleted unfortunately.
This commit can fix this bug. I tested in a lookup dialog in submission, it works, tested in edit metadata dialog of a published item, it works as well.
Thanks for this PR!

alanorth:
@samuelcambien Thanks for the PR. Sorry for the late followup on this, but I'm having difficulty reproducing this issue in DSpace 6.3. I am using controlled vocabularies in XMLUI, but I don't know how to trigger this bug—I click "Click to see available values", select a value and the popup disappears, then I press the "Add" button. It works the same before and after the patch for me...?

PR #1911, opened 2018-01-04
DS-3796 Fix missing bitstream detection and reporting for Checksum Checker

aprilherron:
Currently, Checksum Checker fails to correctly process missing bitstreams. If the bitstream is not found, it just uses the historical checksum for comparison instead of marking the bitstream as missing. In addition, the MostRecentChecksumDAOImpl date queries had start and end date parameters reversed.

https://jira.duraspace.org/browse/DS-3796


tdonohue (approved):
👍 Code looks correct based on the description of the problem in DS-3796. I haven't tested this though, so, I'd rely on others to do so and report back.

PR #1916, opened 2018-01-11
DS-3542 Spring security authorizations

tomdesair:
Work in progress PR to integrate Spring Security authorizations checks with the new DSpace 7 REST API.

THIS IS NOT READY YET!


Frederic-Atmire:
For now I put a most of the EPersonRepository integration tests on ignore cause it seems that an EPerson (according to the AuthorizeService) doesn't get READ access to itself. This might be something we should look into.

tomdesair:
Closed in favor of https://github.com/DSpace/DSpace/pull/1976

PR #1921, opened 2018-01-17
[DS-3804] Jar dependency issues running DSpace 7x REST

terrywbrady:
### This PR is a temporary fix to restore stability to the REST build process.
- https://jira.duraspace.org/browse/DS-3804
- https://jira.duraspace.org/browse/DS-3805

### Recommended approach
The more comprehensive approach to the problem will be to upgrade the version of log4j that is in use.  DSpace currently uses a version that is at end of life.
- https://jira.duraspace.org/browse/DS-3135
- https://docs.spring.io/spring-boot/docs/current/reference/html/howto-logging.html#howto-configure-log4j-for-logging

### DIFF
--- a/dspace-oai/pom.xml
+++ b/dspace-oai/pom.xml
@@ -159,10 +159,6 @@
                      <groupId>org.apache.commons</groupId>
                      <artifactId>commons-lang3</artifactId>
                 </exclusion>
-                <exclusion>
-                     <groupId>org.codehaus.jackson</groupId>
-                     <artifactId>jackson-mapper-asl</artifactId>
-                </exclusion>
             </exclusions>
         </dependency>
 
--- a/dspace-oai/pom.xml
+++ b/dspace-oai/pom.xml
@@ -159,6 +159,10 @@
                      <groupId>org.apache.commons</groupId>
                      <artifactId>commons-lang3</artifactId>
                 </exclusion>
+                <exclusion>
+                     <groupId>org.codehaus.jackson</groupId>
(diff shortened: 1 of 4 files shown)


dspace-spring-rest/pom.xml
```diff
@@ -169,6 +169,20 @@
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-web</artifactId>
 			<version>${spring-boot.version}</version>
+                        <exclusions>
```
tdonohue: Could we add a simple comment here that says something like:

tdonohue (approved):
Looks good overall. One minor request to add in a comment about the temporary dependency exclusions. But, after that, I think this is good to merge

dspace-oai/pom.xml
```diff
@@ -159,6 +159,10 @@
                      <groupId>org.apache.commons</groupId>
                      <artifactId>commons-lang3</artifactId>
                 </exclusion>
+                <exclusion>
```
abollini: I'm a bit worried about this exclusion as it is an artifact from an old, different layout, jackson version. The mapper class are now in the jackson-databinder artifact so if we really use this staff we could hit a classnotfound exception at runtime somewhere... BTW the jtwig-spring version in our depency is quite old the new one is: http://jtwig.org/

abollini (commented):
if the jackson exclusion is not required to have the OAI running properly I will prefer to skip it

PR #1922, opened 2018-01-18
DS-3765: File visit statistics should only be counted for archived items, not for items in workflows

MW3000:
Fix: https://jira.duraspace.org/projects/DS/issues/DS-3765

### DIFF
--- a/dspace-api/src/main/java/org/dspace/statistics/SolrLoggerUsageEventListener.java
+++ b/dspace-api/src/main/java/org/dspace/statistics/SolrLoggerUsageEventListener.java
@@ -8,6 +8,9 @@
 package org.dspace.statistics;
 
 import org.apache.log4j.Logger;
+import org.dspace.content.Bitstream;
+import org.dspace.content.Bundle;
+import org.dspace.content.Item;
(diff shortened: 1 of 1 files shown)


terrywbrady (commented):
I have presumed that the usage statistics could be computed entirely from SOLR query output.  I am concerned about the performance of this change.  If there are a large number of results found in the SOLR query, the process of iterating through the objects could be time consuming.

Unfortunately, I am not aware of a field in the SOLR repo that would provide this same information.

Taking a closer look at your summary in the ticket, I see why this issue is a concern.  If this is only being invoked at an item level (vs collection or community), perhaps my concerns are not as significant. I am interested to hear other opinions.

ssolim:
I dont think this is a concern, too. After the UsageEvent.VIEW for "one" bitstream that belongs to only "one" Bundle that belongs to "one" Item, only one call of item.isArchived is added. This does not depend on the amount in and from SOLR.

The other possible solution would be to drop SOLR logentries when a bitstream is removed. This could make more sense when the scope leaves the workflow.

This is more a general question: Should logentries be available from removed bitstreams ?

PR #1923, opened 2018-01-18
DS-3489: Fix links of resources in embedded page

tomdesair:
This PR restores the links on resources that are inside an embedded page. 

Currently they are not present on the logos of collections embedded inside a community resources returned by the `/api/core/communities` endpoint.


abollini (approved):
the code looks good. Build on the demo server fix the issue on the angular demo site so I'm going to merge

PR #1925, opened 2018-01-19
[DS-3434] DSpace fails to start when a database connection pool is supplied through JNDI

mwoodiupui:
https://jira.duraspace.org/browse/DS-3434

### DIFF
--- a/dspace-api/src/main/java/org/dspace/storage/rdbms/DatabaseUtils.java
+++ b/dspace-api/src/main/java/org/dspace/storage/rdbms/DatabaseUtils.java
@@ -1381,6 +1381,7 @@ public class DatabaseUtils
     /**
      * Get a reference to the configured DataSource (which can be used to
      * initialize the database using Flyway).
+     * The DataSource is configured via our ServiceManager (i.e. via Spring).
      * <P>
      * This is NOT public, as we discourage direct connections to the database
      * which bypass Hibernate. Only Flyway should be allowed a direct connection.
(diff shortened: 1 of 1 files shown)


dspace-api/src/main/java/org/dspace/storage/rdbms/DatabaseUtils.java
```diff
@@ -1374,15 +1381,21 @@ else if (dbms_lc.contains("h2")) // Used for unit testing only
...
      * <P>
      * This is NOT public, as we discourage direct connections to the database
      * which bypass Hibernate. Only Flyway should be allowed a direct connection.
      * @return DataSource
      */
     protected static DataSource getDataSource()
     {
-        // DataSource is configured via our ServiceManager (i.e. via Spring).
-        return DSpaceServicesFactory.getInstance().getServiceManager().getServiceByName("dataSource", BasicDataSource.class);
+        Object dataSource = DSpaceServicesFactory.getInstance()
```
mwoodiupui: This may be too lax.  We might request a javax.jdbc.DataSource.  Both the Tomcat and Commons packages implement DataSource, and it seems unlikely that any other pooling packages would not.

alanorth:
+1 from me. I've just tested this on top of the latest `dspace-6_x` (currently 6.3-SNAPSHOT) and I am able to start DSpace with a database pool provided by through JNDI. Testing environment is macOS, Tomcat 8.5.24, PostgreSQL 9.5.10. Thanks, @mwoodiupui.

tdonohue (approved):
👍 Code looks great to me. It's just adding in null checks, and better exceptions, essentially.

terrywbrady (approved):
Code reviewed

PR #1926, opened 2018-01-22
[DS-3764] Add SSO login to HAL login

terrywbrady:
https://jira.duraspace.org/browse/DS-3764

This PR is intended to start a discussion about the best way to trigger an SSO login from a REST client.  I am marking this as discussion needed.

The HAL browser "Login" page provides a good illustration of the login process for the rest api.

This enhancement illustrates how to implement an SSO login using the same mechanism.

This enhancement also provides a potential mechanism for exposing configuration variables to REST client applications.  

WARNING: The current implementation can expose sensitive configuration variables (db.password), so some discussion is needed on the proper way to filter configuration settings for the client.

### DIFF
--- a/dspace-spring-rest/src/main/java/org/dspace/app/rest/ConfigurationRestController.java
+++ b/dspace-spring-rest/src/main/java/org/dspace/app/rest/ConfigurationRestController.java
@@ -1,0 +1,81 @@
+/**
+ * The contents of this file are subject to the license and copyright
+ * detailed in the LICENSE and NOTICE files at the root of the source
+ * tree and available online at
+ *
+ * http://www.dspace.org/license/
+ */
+package org.dspace.app.rest;
+
+import org.apache.commons.lang.StringUtils;
+import org.apache.log4j.Logger;
+import org.dspace.app.rest.link.HalLinkService;
+import org.dspace.app.rest.model.*;
+import org.dspace.app.rest.model.hateoas.*;
+import org.dspace.app.rest.parameter.SearchFilter;
+import org.dspace.app.rest.repository.DiscoveryRestRepository;
+import org.dspace.app.rest.utils.ConfigurationExtractor;
+import org.dspace.app.rest.utils.ScopeResolver;
+import org.dspace.app.rest.utils.Utils;
+import org.springframework.beans.factory.InitializingBean;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.data.domain.Pageable;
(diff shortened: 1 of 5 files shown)


tomdesair (changes requested):
Hi @terry,

I like the idea of only using the `/api/authn/login` endpoint and not using a dedicated shibboleth login endpoint. 

And while the Configuration REST endpoint in itself is a good implementation, I'm not sure if we should use it in the authentication procedure. This is because each REST client has to be aware of the "authentication-shibboleth.lazysession.loginurl" property key. We should find a way for the REST API to tell the client what it needs to do, so that it can discover the login URL. Therefor I propose the pass the IdP login page URL as a HTTP Location header (and to move the Configuration REST endpoint implementation to a separate PR). That also means we would need only one login button.

I would slightly modify your procedure as follows:

1. A user opens the `/login.html` Angular page or HAL browser page.
2. JavaScript on that page calls `/api/authn/login`
3. The REST API answers with a 401 Unauthorized because no credentials are provided. In case the `org.dspace.authenticate.AuthenticationMethod#loginPageURL` method returns a non-blank value, the REST API also passes this URL in the [Location response header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Location). For `org.dspace.authenticate.ShibAuthentication` this URL should point to the IdP login URL + `?target=<Referer request header value>`. In case multiple AuthenticationMethod implementations are enabled, we will have multiple "Location" headers, but that's something we can address later.
4. JavaScript sees the "Location" header in the response and redirects the user to that page, which in the case of Shibboleth is the IdP login page.
5. The user authenticates with the IdP and is redirected to `/login.html` (because of the `?target=`)
6. The JavaScript on that page re-executes the call to `/api/authn/login` but that call now contains the Shibboleth session cookie. 
7. The Shibboleth daemon on the (REST API) server picks up the Shibboleth session and passes the correspondig Shibboleth attributes to the REST API
8. The DSpace authentication manager translates these Shibboleth attributes to the correct EPerson object
9. The REST Spring security implementation creates an authentication token for that EPerson which is returned to the JavaScript client under the "Authorization" header.

What do you think about this approach?

terrywbrady:
@tomdesair , I like your suggestions for authn.  I will spend some time on implementation today.

I also think that it makes sense to separate the config endpoint.  I am curious to start some discussion around that the config settings that should/should not be exposed.

terrywbrady:
Recommendations implemented in https://github.com/DSpace/DSpace/pull/1927

PR #1931, opened 2018-01-25
DS-3782 upgrade commons collections to version 4

Raf-atmire:
This is the suggested upgrade to commons collections version 4 as requested on jira ticket: https://jira.duraspace.org/browse/DS-3782

### DIFF
--- a/dspace-api/src/main/java/org/dspace/authenticate/IPAuthentication.java
+++ b/dspace-api/src/main/java/org/dspace/authenticate/IPAuthentication.java
@@ -9,6 +9,7 @@ package org.dspace.authenticate;
 
 import java.sql.SQLException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -16,7 +17,6 @@ import java.util.UUID;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
-import org.apache.commons.collections.ListUtils;
 import org.apache.log4j.Logger;
 import org.dspace.core.ConfigurationManager;
 import org.dspace.core.Context;
@@ -164,7 +164,7 @@ public class IPAuthentication implements AuthenticationMethod {
     public List<Group> getSpecialGroups(Context context, HttpServletRequest request)
         throws SQLException {
         if (request == null) {
-            return ListUtils.EMPTY_LIST;
+            return Collections.EMPTY_LIST;
         }
         List<Group> groups = new ArrayList<Group>();
 
--- a/dspace-api/src/main/java/org/dspace/eperson/dao/impl/EPersonDAOImpl.java
+++ b/dspace-api/src/main/java/org/dspace/eperson/dao/impl/EPersonDAOImpl.java
@@ -16,8 +16,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.UUID;
 
-import org.apache.commons.collections.CollectionUtils;
-import org.apache.commons.collections.ListUtils;
+import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.dspace.content.MetadataField;
 import org.dspace.core.AbstractHibernateDSODAO;
@@ -65,7 +64,7 @@ public class EPersonDAOImpl extends AbstractHibernateDSODAO<EPerson> implements 
     public List<EPerson> search(Context context, String query, List<MetadataField> queryFields,
                                 List<MetadataField> sortFields, int offset, int limit) throws SQLException {
         String queryString = "SELECT " + EPerson.class.getSimpleName()
-                                                      .toLowerCase() + " FROM EPerson as " + EPerson.class
+            .toLowerCase() + " FROM EPerson as " + EPerson.class
             .getSimpleName().toLowerCase() + " ";
         if (query != null) {
             query = "%" + query.toLowerCase() + "%";
@@ -84,7 +83,7 @@ public class EPersonDAOImpl extends AbstractHibernateDSODAO<EPerson> implements 
     @Override
     public int searchResultCount(Context context, String query, List<MetadataField> queryFields) throws SQLException {
         String queryString = "SELECT count(*) FROM EPerson as " + EPerson.class.getSimpleName().toLowerCase();
-        Query hibernateQuery = getSearchQuery(context, queryString, query, queryFields, ListUtils.EMPTY_LIST, null);
+        Query hibernateQuery = getSearchQuery(context, queryString, query, queryFields, Collections.EMPTY_LIST, null);
 
         return count(hibernateQuery);
     }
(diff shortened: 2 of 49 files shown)


dspace-api/src/main/java/org/dspace/authenticate/IPAuthentication.java
```diff
@@ -176,7 +177,7 @@ public boolean isImplicit()
     {
         if (request == null)
         {
-            return ListUtils.EMPTY_LIST;
+            return Collections.EMPTY_LIST;
```
tdonohue: In several places in the PR, it looks like you replace `ListUtils` with `Collections`, but leave the `import org.apache.commons.collections4.ListUtils` in place?  Is this on purpose, or should the `import` simply be removed here?

Raf-atmire:
Incase the ListUtils is not used anywhere else in the class, it should
indeed be removed as Collections.EMPTY_LIST is not dependent on that
import. I will remove the unused imports as soon as possible.

tdonohue:
NOTE: This PR looks to currently be failing approval from Snyk (which checks for possible security issues) because we still have a dependency on "commons-beanutils" v1.9.2 which pulls in "commons-collections" version 3.2.1 :(  Here's the report from Snyk

> Introduced through: DSpace/DSpace@DSpace/DSpace#d742d1a86f02eea69d6f982ddbf787f8046e8776 › commons-beanutils:commons-beanutils@1.9.2 › commons-collections:commons-collections@3.2.1

Looking at commons-beanutils, it looks like they have a ticket to update to commons-collections version 4.0 here: https://issues.apache.org/jira/browse/BEANUTILS-500   But, their version 2.0.0 isn't yet released.

So, while this PR is looking good, we won't completely get rid of commons-collections version 3 until we are able to update commons-beanutils.

tomdesair:
Hi @tdonohue, `commons-beanutils` seems to be an optional dependency of Commons Configuration. Do you know if we actual need and use that library?

tdonohue:
@tomdesair : Aha, glad you found the reason why it's in our dependencies. I couldn't recall, but we do use Configuration Builders to load our configs and provide the ability to "override" configs in dspace.cfg with those in local.cfg, etc.  Here's where they are used:
https://github.com/DSpace/DSpace/blob/master/dspace-services/src/main/java/org/dspace/servicemanager/config/DSpaceConfigurationService.java#L491

So, it's rather annoying here that we aren't going to be able to get rid of `commons-collections` version 3 until `commons-beanutils` is updated, and then `commons-configuration` updates to the latest `commons-beanutils`.  I don't see any easier way around this.

pom.xml
```diff
@@ -1096,16 +1096,16 @@
...
-                <groupId>commons-collections</groupId>
-                <artifactId>commons-collections</artifactId>
-                <version>3.2.2</version>
-                <!-- <version>3.1</version> xmlui - wing -->
-            </dependency>
             <dependency>
                 <groupId>org.apache.commons</groupId>
                 <artifactId>commons-collections4</artifactId>
                 <version>4.1</version>
+                <!-- <version>3.1</version> xmlui - wing -->
```
mwoodiupui: We don't need this comment now, since XMLUI is no more.

tdonohue:

tdonohue:
@Raf-atmire : unfortunately, this PR seems to have similar issues to #1964 now (though not to the extreme of #1964) In aligning this PR with the new Code Style, the number of lines changed has now increased, as code you didn't change is being re-aligned / re-wrapped by your IDE.  I'm guessing you may have done some sort of bulk IDE rewrapping that has caused both of these PRs to now try and update the wrapping of unrelated code?

If you are bulk-rewrapping code in IntelliJ, I'd recommend using the same settings that were used to align everything with the Code Style. Otherwise, your IDE will likely continue to make unnecessarily realignment / rewrapping changes that will clutter your PRs. For these IntelliJ settings see: https://wiki.duraspace.org/display/DSPACE/Code+Style+Guide#CodeStyleGuide-ConfigureIDEAforbulkcleanup

The other option is to simply install the Checkstyle plugin for your IDE, in which case it will highlight code style issues *as you write your code*, allowing you to manually fix them as you go, instead of relying on bulk tools to properly format your code.  For the proper checkstyle plugin see: https://wiki.duraspace.org/display/DSPACE/Code+Style+Guide#CodeStyleGuide-IDESupport

tomdesair:
@tdonohue I've restored the lines which should not have been modified. I've also rebased on master.

tdonohue:
@tomdesair : thanks! Looks like there's a minor compilation failure though from Travis: https://travis-ci.org/DSpace/DSpace/builds/405388711#L3363 

In meantime, I'll start to give this an updated review

pom.xml
```diff
@@ -1149,17 +1149,16 @@
...
-            </dependency>
             <dependency>
                 <groupId>org.apache.commons</groupId>
                 <artifactId>commons-collections4</artifactId>
                 <version>4.1</version>
             </dependency>
+            <dependency>
+                <groupId>commons-collections</groupId>
+                <artifactId>commons-collections</artifactId>
+                <version>3.2.2</version>
```
tdonohue: This dependency should be dropped, correct?

dspace-api/src/main/java/org/dspace/eperson/dao/impl/EPersonDAOImpl.java
```diff
@@ -96,14 +95,14 @@ public int searchResultCount(Context context, String query, List<MetadataField>
                                                       .toLowerCase() + " FROM EPerson as " + EPerson.class
             .getSimpleName().toLowerCase();
 
-        List<MetadataField> sortFields = ListUtils.EMPTY_LIST;
+        List<MetadataField> sortFields = Collections.EMPTY_LIST;
 
-        if (metadataSortField != null) {
+        if (metadataSortField != null) {ItemRestRepositoryIT
```
tdonohue: Here's the compilation failure.  A random `ItemRestRepositoryIT` was accidentally included here.

pom.xml
tomdesair: No, we still have the issue with `beanutils` mentioned above. Removing that dependency results in:

```
[WARNING] 
Dependency convergence error for commons-collections:commons-collections:3.2.1 paths to dependency are:
+-org.dspace:dspace-api:7.0-SNAPSHOT
  +-commons-validator:commons-validator:1.5.0
    +-commons-beanutils:commons-beanutils:1.9.2
      +-commons-collections:commons-collections:3.2.1
and
+-org.dspace:dspace-api:7.0-SNAPSHOT
  +-commons-validator:commons-validator:1.5.0
    +-commons-collections:commons-collections:3.2.2

[WARNING] Rule 0: org.apache.maven.plugins.enforcer.DependencyConvergence failed with message:
Failed while enforcing releasability the error(s) are [
Dependency convergence error for commons-collections:commons-collections:3.2.1 paths to dependency are:
+-org.dspace:dspace-api:7.0-SNAPSHOT
  +-commons-validator:commons-validator:1.5.0
    +-commons-beanutils:commons-beanutils:1.9.2
      +-commons-collections:commons-collections:3.2.1
and
+-org.dspace:dspace-api:7.0-SNAPSHOT
  +-commons-validator:commons-validator:1.5.0
    +-commons-collections:commons-collections:3.2.2
]
```

tdonohue: @tomdesair : Aha, yes, I neglected to reread the discussion above.  You are correct, we do need it still.

tdonohue (approved):
👍 Code looks correct, and it passes all tests/build.

I am going to add only minor addition to this PR -- a comment to the main POM about why we still need `commons-collections` v3.  However, once that is applied, I think this can be merged.

PR #1933, opened 2018-01-25
[DS-3434] DSpace fails to start when a database connection pool is supplied through JNDI

mwoodiupui:
Ported from the v6 patch.
https://jira.duraspace.org/browse/DS-3434


tdonohue (approved):
👍 LGTM. It's a port of #1925, which was already approved/merged to `dspace-6_x`

hardyoyo (approved):
+1 by inspection

PR #1934, opened 2018-01-25
[DS-3811] Integrate Shibboleth into DSpace REST Report Tools

terrywbrady:
https://jira.duraspace.org/browse/DS-3811

To enable password authentication or shibboleth authentication for the reports, see the comments at the top of restCollReport.js and restQueryReport.js.

This change has permitted the removal of the optional configuration setting that bypassed authentication checks when running the report tools from a protected rest api.

### DIFF
--- a/dspace-rest/src/main/java/org/dspace/rest/FilteredCollectionsResource.java
+++ b/dspace-rest/src/main/java/org/dspace/rest/FilteredCollectionsResource.java
@@ -85,9 +85,6 @@ public class FilteredCollectionsResource extends Resource {
         try
         {
             context = createContext();
-            if (!configurationService.getBooleanProperty("rest.reporting-authenticate", true)) {
-                context.turnOffAuthorisationSystem();            	
-            }
 
             if (!((limit != null) && (limit >= 0) && (offset != null) && (offset >= 0)))
             {
@@ -171,9 +168,6 @@ public class FilteredCollectionsResource extends Resource {
         FilteredCollection retColl = new org.dspace.rest.common.FilteredCollection();
         try {
             context = createContext();
-            if (!configurationService.getBooleanProperty("rest.reporting-authenticate", true)) {
-                context.turnOffAuthorisationSystem();            	
-            }
 
             org.dspace.content.Collection collection = collectionService.findByIdOrLegacyId(context, collection_id);
             if(authorizeService.authorizeActionBoolean(context, collection, org.dspace.core.Constants.READ)) {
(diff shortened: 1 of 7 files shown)


tdonohue (approved):
👍  Nice!  I haven't tested, but the code looks good/clean.

We would want to ensure this also gets merged to `master` too.  Also, we'd need to ensure documentation is updated, since several configs are removed, and it looks like you have to decide which form of authentication to enable in each of the various `*.js` report files.

kshepherd (approved):
+1, a nice addition. this is by inspection, i can't test shibboleth at the moment. thanks @terrywbrady !

i'm not too familiar with the way the shib link is created, i've typically configured the webserver to protect/redirect for a certain URL pattern rather than hardcoding a link but that's possibly just personal preference

terrywbrady:
Added commit based on discussion with @tdonohue and @kshepherd .

Tim Donohue [2:24 PM]
@terrywbrady: RE: 1934 / 1991, Do we need to document that hardcoded Shib Link question somewhere? Or turn it into a variable?
Other than that, those PRs look fine. I'm not noting it seems like there's a question on whether that Shib URL is always the same...and if there's a doubt, maybe it should be a variable or something

Terry Brady [2:25 PM]
I will post a quick commit to each PR.

Tim Donohue [2:25 PM]
(Or cover this in documentation...that's fine too)

Kim Shepherd [2:26 PM]
yeh that was my question... since i'm used to managing that stuff in apache config rather than in the app itself, it's always been a 'variable' for me in as much as it's part of a virtualhost config or something

Tim Donohue [2:27 PM]
And it looks to me like there are two hardcoded URL paths here:
https://github.com/DSpace/DSpace/pull/1934/files#diff-2bfc0af825f2a8b5f28b58da4f60338bR250
and
https://github.com/DSpace/DSpace/pull/1934/files#diff-2bfc0af825f2a8b5f28b58da4f60338bR272

Terry Brady [2:27 PM]
In REST 7, we will have the ability to return config values to the client.  For this short term app, I am thinking of the following change
   //Indicate if Shibboleth Authentication is supported
   this.makeShibLink = function(){return false;};
   this.shibPath = "/Shibboleth.sso/Login";

Tim Donohue [2:28 PM]
@terrywbrady: that seems fine for the second hardcoded path.

Terry Brady [2:29 PM]
The other path is the official endpoint.  The rest of the tools assume /rest is the root path.  (Look for status, logout, registries...)

Tim Donohue [2:29 PM]
The first hardcoded path seems to assume the REST API will always be at `/rest`.  I'd guess that'd usually be correct, but I've found it's best never to assume things are installed on particular paths.

Terry Brady [2:30 PM]
I would be happy to address the root path issue in a new ticket/PR.  I think to close out the Shibb issue, we should keep the scope of the changes small..

Tim Donohue [2:30 PM]
Oh, ok. If that `/rest` path assumption is throughout these JS files, then that's fine...leave it as is.  If someone gets annoyed, they can change it and submit a PR.

Terry Brady [2:31 PM]
I wil add the commits I described above and alert you all when they are in place.  I will test in 6 and 7 on my end.

Tim Donohue [2:31 PM]
And I agree, I'm not trying to slow down this PR. Getting this ready ASAP is better

PR #1936, opened 2018-01-26
[DS-3524] Restore maven-enforcer-plugin on rest7

mwoodiupui:
https://jira.duraspace.org/browse/DS-3524
Small version adjustments to satisfy the enforcer rule.
This will conflict with DS-3816, which also converges one of these dependencies.  If this patch is adopted, DS-3816 can be discarded.
This patch builds cleanly with all unit and integration tests enabled.  I have not done any other testing yet.

### DIFF
--- a/dspace-api/pom.xml
+++ b/dspace-api/pom.xml
@@ -492,7 +492,7 @@
         </dependency>
         <dependency>
             <groupId>org.hamcrest</groupId>
-            <artifactId>hamcrest-all</artifactId>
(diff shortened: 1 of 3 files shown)


abollini:
Add a link to the PR related to the DS-3816 https://github.com/DSpace/DSpace/pull/1932

tdonohue (approved):
👍 Looks good to me

terrywbrady (approved):

PR #1942, opened 2018-02-01
[DS-3824] Bitstream metadata duplicated when creating a version

KevinVdV:


KevinVdV:
Oracle query to remove the duplicate rows has NOT been tested yet.

pnbecker:
This would need to be cherry-picked to master.
This seems to be related to #1883

kshepherd:
@KevinVdV does this PR fix the same issue as https://github.com/DSpace/DSpace/pull/1883 and if so, do you have a preference for which implementation to use?

kshepherd:
No answer from author and PR now has conflicts, moving to 6.4 until we can resolve

PR #1946, opened 2018-02-05
DS-3827 LazyInitializationException in formats.jsp

philip-muench:
https://jira.duraspace.org/browse/DS-3827

This JSP produces an internal error due to a LazyInitializationException. The exception is produced because the dspace context is aborted before completing all database queries. JIRA ticket: DS-3827


tdonohue (approved):
👍  Untested, but the code change looks obvious to me

pnbecker (approved):
+1
Code looks right, passes my tests, thanks for the PR!

PR #1947, opened 2018-02-05
DS-3828 Add latest version information to versioning Navigation validity

jonas-atmire:
PR for https://jira.duraspace.org/browse/DS-3828
This will make sure, that the versioning Navigation, the version history is taken into account.
Adding this will make sure that the appropriate checks are being done to see what should or should not be shown


tdonohue (approved):
👍 Untested, but seems like an obvious fix to me

terrywbrady:
@jonas-atmire , I have attempted to reproduce the problem described in Jira, and I am not able to reproduce this issue.  Could you revise those instructions with more details?

I am not very familiar with DSpace versioning.  I have enabled it in my repository and I have created an item with 3 versions.  As I navigate from version to version, the sidebar links appear to be correct for the version that I am viewing.

kshepherd:
Shifting to 6.4 milestone as this PR has an unanswered question/clarification regarding reproduction of the original issue

PR #1948, opened 2018-02-05
[DS-3803] Remove db.jndi setting from dspace.cfg

mwoodiupui:
Cherry-picked from 6_x branch:
As of DSpace 6.x this setting is no longer used and is not customizable
by the user. Now DSpace always looks for a pool named "jdbc/dspace" in
JNDI and falls back to creating a pool with the db.* settings located
in dspace.cfg.

See: https://wiki.duraspace.org/display/DSDOC6x/Configuration+Reference
See: dspace/config/spring/api/core-hibernate.xml
See: https://jira.duraspace.org/browse/DS-3434

### DIFF
--- a/dspace/config/dspace.cfg
+++ b/dspace/config/dspace.cfg
@@ -95,12 +95,6 @@ db.maxwait = 5000
 # (default = 10)
 db.maxidle = 10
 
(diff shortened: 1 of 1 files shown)


tdonohue (approved):
👍 Looks good to me

PR #1949, opened 2018-02-06
DS-3830 Item retrieval fallbacks for versioning navigation

jonas-atmire:
PR for https://jira.duraspace.org/browse/DS-3830
This tries additional fallbacks to retrieve the item when creating the validity for the Item versioning navigation

### DIFF
--- a/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/versioning/Navigation.java
+++ b/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/versioning/Navigation.java
@@ -11,6 +11,7 @@ import org.apache.cocoon.caching.CacheableProcessingComponent;
 import org.apache.cocoon.environment.ObjectModelHelper;
 import org.apache.cocoon.environment.Request;
 import org.apache.cocoon.util.HashUtil;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.excalibur.source.SourceValidity;
 import org.apache.excalibur.source.impl.validity.NOPValidity;
 import org.dspace.app.util.Util;
@@ -124,8 +125,10 @@ public class Navigation extends AbstractDSpaceTransformer implements CacheablePr
 		            }
 
                     DSpaceObject dso = HandleUtil.obtainHandle(objectModel);
-                    if(dso != null)
-                    {
+		            if(dso == null){
+                        dso = getItemById();
+                    }
+                    if (dso != null) {
                         validity.add(context, dso);
                     }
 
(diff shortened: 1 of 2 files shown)


terrywbrady (approved):
Tested and reviewed, the change looks good.

dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/versioning/Navigation.java
```diff
@@ -124,8 +125,10 @@ public SourceValidity getValidity()
...
 
                     DSpaceObject dso = HandleUtil.obtainHandle(objectModel);
-                    if(dso != null)
-                    {
+		            if(dso == null){
+                        dso = getItemById();
+                    }
+                    if (dso != null) {
                         validity.add(context, dso);
                     }
```
kshepherd: The inserted `if` test here is intended to far, and makes the block a bit hard to follow.
Also, instead of two explicit `if` tests with logically opposite conditions, I think it would be easier to follow if we changed the whole block to an `if / else`
I will commit these minor changes directly rather than make new work for you :)

dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/versioning/Navigation.java
```diff
@@ -201,6 +204,12 @@ private Item getItemById() throws SQLException
         if (itemId != null)
         {
             item = itemService.find(this.context, itemId);
+        } else {
+            String itemIDParam = parameters.getParameter("itemID", null);
+            if (StringUtils.isNotBlank(itemIDParam)) {
+                itemId = UUID.fromString(itemIDParam);
+                item = itemService.find(context, itemId);
+            }
```
kshepherd: Just another minor formatting/styling nitpick... this has some muddled use of blocks / brackets which makes it a little hard to parse the tests, and doesn't match the DSpace convention of
`if(test) {
    // ...
}
else {
    // ...
}`

(again, I'll change this myself just to save time, wanted to make a note here though so you can see why I'm messing with your PR!)

kshepherd (changes requested):
Hi @jonas-atmire , thanks for the contribution! Looks great, I'm +1 by inspection once we just tidy up a few code style issues as per inline comments.

I've made some quick code style changes and simplified that if/else test, but I don't have permission to push to this PR branch on your fork. Could you please give me permission to push? (https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/)
This will let DSpace maintainers push changes to this PR only, doesn't affect any other permissions.

Alternatively, you could commit and push the changes yourself. I now realise that all the versioning classes have this different code style, but I figure while we are committing and testing changes to Navigation, we might as well bring it in line with DSpace convention.

Cheers

Kim

dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/versioning/Navigation.java
mwoodiupui: Looks like mixed tabs/spaces indenting again.

kshepherd:
no response from author, decided in dev meeting to merge anyway and let code style be fixed in dspace 7. thanks for the patch @jonas-atmire !

kshepherd (approved):

PR #1951, opened 2018-02-08
[DS-3835] Add js work around to preserve current scope

terrywbrady:
https://jira.duraspace.org/browse/DS-3835

This fix should also be ported to DSpace 5x.


kshepherd (approved):
+1, reproduced issue in dspace-6_x and confirmed that scope is kept in XMLUI when PR is applied

kshepherd:
small fix and quick win, so merging now, thanks @terrywbrady !

PR #1958, opened 2018-02-16
[DS-3839] backporting DSPR#1956 for dspace-5_x

hardyoyo:
https://jira.duraspace.org/browse/DS-3839


tdonohue (approved):
👍 Looks good, merging immediately as it's already been merged to 6.x

Thanks @hardyoyo !

PR #1965, opened 2018-02-24
[DS-3852]OAI indexer message not helpful in locating problems

mwoodiupui:
https://jira.duraspace.org/browse/DS-3852
This identical patch is in production, in our somewhat modified 5.6 instance, and the affected command has been run multiple times while chasing another problem.


mwoodiupui:
This patch helped me track down a crash in the indexer (caused by local database corruption).

tdonohue (approved):
👍 Looks good to me. An obvious quick win

PR #1966, opened 2018-02-24
[DS-3852] OAI indexer message not helpful in locating problems

mwoodiupui:
https://jira.duraspace.org/browse/DS-3852
Ported from #1965.


tdonohue (approved):
👍 Looks good to me. An obvious quick win

PR #1968, opened 2018-02-24
[DS-3202] PMH Responder returns 500 Internal Error if item is in no Community or no Collection

mwoodiupui:
https://jira.duraspace.org/browse/DS-3853
This is running in production and the instance now satisfies the openarchives.org validator.


mwoodiupui:
Oops!  This was already fixed in master by #1401 and seems to be present in branches dspace-6_x and master (will be DSpace 7).  I'm rewording this PR to attach it to DS-3202 and closing DS-3853.

tdonohue (approved):
👍 Code looks good to me. This is an obvious, minor refactor to add in `if (null != [obj])` sorts of checks.

terrywbrady (approved):
+1 Tested

kshepherd:
merging! thanks @mwoodiupui

PR #1972, opened 2018-02-28
DS-3858 add HttpRequestRetryHandler to DataCiteCon

ssolim:
https://jira.duraspace.org/browse/DS-3858

### DIFF
--- a/dspace-api/src/main/java/org/dspace/identifier/doi/DataCiteConnector.java
+++ b/dspace-api/src/main/java/org/dspace/identifier/doi/DataCiteConnector.java
@@ -19,6 +19,8 @@ import org.apache.http.HttpResponse;
 import org.apache.http.StatusLine;
 import org.apache.http.auth.AuthScope;
 import org.apache.http.auth.UsernamePasswordCredentials;
+import org.apache.http.client.CredentialsProvider;
+import org.apache.http.client.HttpClient;
 import org.apache.http.client.methods.HttpDelete;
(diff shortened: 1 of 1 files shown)


pnbecker (changes requested):
Hi @ssolim,

thank you for the Pull-Request. DataCite's test systems has some problems today, which make it hard to test this PR. So I tested it against the production system using the test Prefix 10.5072 (this was not possible in earlier versions of DSpace, but works since DSpace 6).

I like your changes as they make the DataCiteConnector a bit more reliable. Unfortunately the way we construct the DefaultHttpClient is deprecated. I added some more changes and will create a PR against you feature branch. Please consider them.

Could you please also squash your two commits into one? Your second commit misses the issue number (DS-3858) and there is no need for two commits here.

Thanks for pulling our attention to this and helping out with code!

pnbecker:
@ssolim Sorry, I'm unable to create a PR against your repository as github doesn't see it as official fork of DSpace. Could you please cherry-pick my latest commit from [here](https://github.com/the-library-code/DSpace/tree/DS-3858)?

pnbecker (approved):

pnbecker:
The changes I requested were fulfilled. +1
I tested this code quite well, would be great if it could be included into 6.3 and cherry-picked to master. @terrywbrady, @kshepherd could you please give it another short review?

tdonohue (approved):
👍 Looks good to me. Haven't tested, but the code looks correct.

PR #1977, opened 2018-03-02
[DS-3775] Hibernate session bug when submitting item with Versioning enabled

jsavell:
https://jira.duraspace.org/browse/DS-3775

### DIFF
--- a/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/versioning/Navigation.java
+++ b/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/versioning/Navigation.java
@@ -126,7 +126,7 @@ public class Navigation extends AbstractDSpaceTransformer implements CacheablePr
                     DSpaceObject dso = HandleUtil.obtainHandle(objectModel);
                     if(dso != null)
                     {
-                        validity.add(context, context.reloadEntity(dso));
+                        validity.add(context, dso);
                     }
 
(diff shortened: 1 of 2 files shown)


tomdesair (changes requested):
Thanks for the PR @jsavell! This week we encountered a similar issues but in another part of the code.

The root cause of the problem seems to be the `HandleUtil.obtainHandle` method that retrieves an internally cached version of a DSO. This DSO is potentially detached if a commit happened before.

Do you agree that it would be better to move the `context.reloadEntity(dso)` call to the `return` statement in `HandleUtil.obtainHandle` method? That would fix this issue also in other places.

jsavell:
That sounds great.

I'm on vacation right now, but will update the PR within a few days.

jsavell:
@tomdesair The reloadEntity call has been moved to HandleUtil as requested.

tomdesair (approved):

tomdesair:
Thanks @jsavell ! That will also prevent future issues with the `HandleUtil.obtainHandle` method.

tdonohue (approved):
👍 Haven't tested, but this fix seems quite in line with other `reloadEntity()` fixes applied in 6.1 and 6.2.  So, the code looks good to me.

terrywbrady (approved):
Tested locally and verified the fix.  @jsavell , thanks for the good instructions for re-creating the issue.

PR #1982, opened 2018-03-07
DS-3865: HTML5 Upload Pause Button does not work

philip-muench:
Make pause button work again
https://jira.duraspace.org/browse/DS-3865


kshepherd (approved):
+1, I reproduced the problem in dspace-6_x JSPUI and with this PR applied, I can pause and resume large uploads as many times as I like, no JS errors and the upload completes successfully. Thanks @philip-muench !

hardyoyo (approved):
+1 good catches

kshepherd:
This is a quick, easy change with no master port needed so I'm not going to wait for another +1, merging it to 6.3 now! Thanks @philip-muench !

PR #1983, opened 2018-03-08
[DS-3841] REBASE enable CORS header for security controller

PTrottier:
This PR is a rebase for #1959

### DIFF
--- a/dspace-spring-rest/src/main/java/org/dspace/app/rest/Application.java
+++ b/dspace-spring-rest/src/main/java/org/dspace/app/rest/Application.java
@@ -145,10 +145,11 @@ public class Application extends SpringBootServletInitializer {
             public void addCorsMappings(CorsRegistry registry) {
                 String[] corsAllowedOrigins = configuration.getCorsAllowedOrigins();
                 if (corsAllowedOrigins != null) {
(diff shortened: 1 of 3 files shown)


kshepherd (commented):
Thanks @PTrottier! This looks good, and I'm almost +1 (by inspection), I just have one comment / question, about the removal of the context.setCurrentUser call that was added recently:
It's true that in this specific case, the eperson is passed explicitly to authenticationProvider.getGrantedAuthorities which in turn calls the isAdmin signature that looks only for the eperson in the eperson object, not context... I just wonder whether this will always be the case -- if some code is added later that relies on context holding correct data about the current user, will we remember to add this line back in? Is it tidier to remove it since it's technically redundant right here & now, or safer to leave it in so that we don't accidentally use a bad context object in the future? (or am I too focused on context / thinking in a pre-7 way?)

kshepherd (approved):
@tomdesair explained the move to JWT meaning context current user should always be set by EPersonClaimProvider, so that clears up my earlier question. +1 by inspection

tdonohue (approved):
👍 LGTM too, and it matches what was in #1959 (but rebased).  Thanks, @PTrottier!

PR #1986, opened 2018-03-15
DS-3869 Record stats events even if location service fails to init

aprilherron:
When the server starts up, `SolrLoggerServiceImpl` (previously `SolrLogger`) initializes a location `LookupService` using the Geo Database file configured in `usage-statistics.dbfile`. By default this is `${dspace.dir}/config/GeoLiteCity.dat`.

However, if this file is removed, or fails to load, the location service fails to initialize and, as a result, usage events are no longer recorded by `SolrLoggerServiceImpl`. This is due to a `null` check which skips the creation of the solr document. However, it makes more sense to simply omit the location metadata from the Solr document in this case. In fact, a null check is already in place for these code-blocks.


KingKrimmson:
Closing, re-opened new pull-request from personal branch

PR #1988, opened 2018-03-16
[DS-3714] port pr 1862 to master

terrywbrady:
https://github.com/DSpace/DSpace/pull/1862


tdonohue (approved):
👍 Looks good to me! Matches with changes in #1862, and @terrywbrady mentioned on Slack that it was tested locally.

PR #1989, opened 2018-03-16
[DS-3704] port pr 1854 to master

terrywbrady:
https://github.com/DSpace/DSpace/pull/1854


tdonohue (approved):
👍 Looks good to me (same changes as in #1854), and @terrywbrady mentioned on Slack was tested locally.

PR #1991, opened 2018-03-16
[DS-3811] port pr 1934 to master

terrywbrady:
PR #1934


.checkstyle
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mwoodiupui: Was this file supposed to be included?  It seems to be local to your development environment.

terrywbrady:
Thanks @mwoodiupui , I cleaned up the pr and removed that file.

We may want to add .checkstyle to .gitignore to support Eclipse users

tdonohue (approved):
👍 LGTM. Identical changes to #1934. This is just a forward-port. Merging immediately.

PR #1993, opened 2018-03-19
Avoid NullPointerException with empty sub-communities.

hcayless:
If there are sub-communities that contain nothing, a community metadata export will fail with a NullPointerException, because the buildFromCommunity method will return null when called recursively and then the code will try to add that null to the result Iterator. This patch just checks whether the buildFromCommunity call returned null before trying to append it, avoiding the exception.

### DIFF
--- a/dspace-api/src/main/java/org/dspace/app/bulkedit/MetadataExport.java
+++ b/dspace-api/src/main/java/org/dspace/app/bulkedit/MetadataExport.java
@@ -122,7 +122,9 @@ public class MetadataExport {
                 System.out.print(" ");
             }
             Iterator<Item> items = buildFromCommunity(context, subCommunity, indent + 1);
-            result = addItemsToResult(result, items);
+            if (items != null) {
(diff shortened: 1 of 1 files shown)


mwoodiupui:
That should fix the immediate problem.  A more rugged fix would be to instead have addItemsToResult immediately return 'result' if 'items' is null.  That way, if it is ever called elsewhere, we don't have to remember to check for null before calling.

(buildFromCommunity *should* have returned an empty Iterator rather than null....)

hcayless:
That would work fine. I can modify the pull request to do that instead if you'd prefer.

mwoodiupui:
Yes, please do.

hcayless:
addItemsToResult looked sort of unbalanced when I added the new condition and return, so I rewrote it without the if/else. If items is null, it immediately returns result; if result is empty it immediately returns items, and if it makes it past those conditions, it returns items added to result.

PR #1997, opened 2018-03-21
[DS-3832] Upgrade to GeoIP2.

mwoodiupui:
https://jira.duraspace.org/browse/DS-3832

Not yet tested.


kshepherd (approved):
+1 merging immediately as this is an almost identical commit to 6.x (apart from some pom changes in 6.x that already existed in master)

PR #1998, opened 2018-03-21
DS-3875: Fix the attribute name of <identifier> tag to 'identifierType'.

dineshmendhe1:
Fix the attribute name of <identifier> tag to 'identifierType' in DIM2EZID.xsl file. 

https://jira.duraspace.org/browse/DS-3875

### DIFF
--- a/dspace/config/crosswalks/DIM2EZID.xsl
+++ b/dspace/config/crosswalks/DIM2EZID.xsl
@@ -63,7 +63,7 @@
             <!--
                  For EZID we need an empty field here.
             -->
-                <identifier type='DOI'/>
+                <identifier identifierType='DOI'/>
 
             <!--
                 DataCite (2)


kshepherd (approved):
+1, based on inspection and review of the datacite schema http://schema.datacite.org/meta/kernel-3/metadata.xsd which specifies identifierType as the correct attribute. Thanks @dineshmendhe1 !

Possibly out of scope, but there is a kernel-4 metadata schema now, do we also need to think about updating this, or does that involve a new API version that we're not ready for? (see APIv2 docs, https://ezid.cdlib.org/doc/apidoc.html)

tdonohue (approved):
👍 Looks good / obvious to me.

@kshepherd : I'd suggest keeping the scope of this fix small (as it is currently an obvious bug). Any other issues/updates could be dealt with separately in a new ticket/PR

kshepherd:
This is a good, small fix so I'm merging to 6.3 now rather than waiting for another +1.
Thanks @dineshmendhe1 !

PR #1999, opened 2018-03-22
DS-3870 S3 bitstore leaves connections to AWS open

benbosman:
Fixes https://jira.duraspace.org/projects/DS/issues/DS-3870

Will only open a connection to the file (using the retrieve method) if it will be downloaded and closed afterwards

### DIFF
--- a/dspace-xmlui/src/main/java/org/dspace/app/xmlui/cocoon/BitstreamReader.java
+++ b/dspace-xmlui/src/main/java/org/dspace/app/xmlui/cocoon/BitstreamReader.java
@@ -179,6 +179,7 @@ public class BitstreamReader extends AbstractReader implements Recyclable
     protected ItemService itemService = ContentServiceFactory.getInstance().getItemService();
     protected CitationDocumentService citationDocumentService = DisseminateServiceFactory.getInstance().getCitationDocumentService();
 
+    private boolean hasNotBeenModified = false;
 
     /**
      * Set up the bitstream reader.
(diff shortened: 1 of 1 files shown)


kshepherd (commented):
Looks good, thanks @benbosman , i guess my main question is.. this looks like it keeps things more efficient when responding to spiders and when the file isn't modified, but how does it resolve a general problem with closing S3 connetions? Doesn't the "finally" block at https://github.com/DSpace/DSpace/blob/e5cb62997adf56ef5235418cd0856f1f8724afc7/dspace-xmlui/src/main/java/org/dspace/app/xmlui/cocoon/BitstreamReader.java#L752 handle that? Or if not, doesn't that also need fixing?

kshepherd:
Shifting to 6.4 until open questions are answered and we get some S3 testers to give it a whirl (if we get some action in the next week we can move back to 6.3)

kshepherd:
ok thanks @benbosman. i don't know any S3 users so it'll be hard to test. we'll discuss at next meeting as over the last month we've been really trying to get things tied down and not continue postponing 6.3

tdonohue (approved):
👍 This code looks correct to me. Haven't had a chance to test it, but the process seems to be the following:
* If client is a know "spider", then check to see if bitstream has changed since the "If-Modified-Since" header.  If not, set our `hasNotBeenModified=true`
* If `hasNotBeenModified==true`, then don't let the client download the file. Instead, return a "Last-Modified" header (corresponding to last modified date), along with a 304 Not Modified response.
* Finally reset `hasNotBeenModified=false` in `recycle()` method.

This logic all seems sound to me, especially as it only is triggered for clients which are identified as a spider.

kshepherd (approved):
tests well for XMLUI downloads and now that i understand the logic, this makes sense. merging now, thanks @benbosman

benbosman:
Thank you @kshepherd and @tdonohue for the reviews

PR #1996, opened 2018-03-21
DS-3511: Fix HTTP 500 errors on REST API bitstream updates

AndrewBennet:
This is a replacement pull request for https://github.com/DSpace/DSpace/pull/1840 - the commit has been cherry-picked onto the `dspace-6_x` branch.

PR #2001, opened 2018-03-27
DS-3511: Fix HTTP 500 errors on REST API bitstream updates (7.0)

minurmin:
This is a replacement pull request for #1840, rebased against the master branch.

When trying to update the content of a bitstream or delete a bitstream policy HTTP 500 error occurs (tested with DSpace 6.0-6.2). Fixed by adding context.complete() to the end of the respective methods (adapted the original patch submitted to DS-3511).


kshepherd (approved):
+1 merging immediately as this is a master port of #1996 which is already tested, reviewed, merged

PR #2002, opened 2018-03-28
DS-3877 Trim bitstream name during filter-media comparison

aprilherron:
While the UI usually protects against this case, it is possible to have a bitstream name with a leading space. When `MediaFilterService` attempts to check if a bitstream has already been generated for the filter, the name comparison fails due to the leading space and a duplicate bitstream is generated.

This is solved by simply trimming the name during the comparison.


terrywbrady:
@KingKrimmson , @tdonohue , 

I see that this PR has been categorized for the 6.3 release but the pull request is targeted for the master branch.  I imagine that we would want this change for both branches.

@KingKrimmson , could you create a second PR with a target branch of dspace-6_x?  That would be easier to test.

terrywbrady (approved):
Tested locally and verified the change.

KingKrimmson:
Submitted a 6.x version: [PR 2012](https://github.com/DSpace/DSpace/pull/2012)

kshepherd:
merging this as tested and reviewed (see #2012), thanks @KingKrimmson !

PR #2006, opened 2018-03-28
DS-3889 Change Content-Type in OAI-Response (DSpace 6 port)

kshepherd:
DSpace 6 Port for https://github.com/DSpace/DSpace/pull/1804, commit cherrypicked against dspace-6_x:

As per OAI 2.0 spec (3.1.2.1) the response content type *must* be text/xml.
http://www.openarchives.org/OAI/openarchivesprotocol.html#HTTPResponseFormat

Our OAI client is rejecting the response from DSpace OAI server.


mwoodiupui (approved):
+1  Correct according to OAI spec.  This is a port of a patch which has already been merged.

PR #2008, opened 2018-03-28
DS-3719: Porting PR 2007 to 6.x

philip-muench:


philip-muench:
Issue: https://jira.duraspace.org/browse/DS-3719

terrywbrady (changes requested):
@philip-muench , I attempted to reproduce this error on the existing DSpace 6.3 branch, and the problem is no longer present.  I suspect that some other fix has been made within the metadata service since the 6.2 release.  I am not certain that this change is needed.

philip-muench:
You're right. The change can be found in MetadataFieldDAOImpl,java, Whereas in my local version (from before Oct'17), the qualifier of the field is just being compared to null in findByElement(), now  StringUtils.isNotBlank() is used to check if the qualifier needs to be part of the db query (commit from 14th Nov). My PR is not necessary any more. I will close it.

PR #2011, opened 2018-03-30
[DS-3832] GeoIP-API(com.maxmind.geoip:geoip-api) needs to be replaced by GeoIP2 ( com.maxmind.geoip2:geoip2 )

mwoodiupui:
https://jira.duraspace.org/browse/DS-3832
Cherry picked from #2003.
Passes tests.  Not yet run.


kshepherd (approved):
+1 merging immediately as port of tested and reviewed 6.x PR

PR #2012, opened 2018-04-04
DS-3877 Trim bitstream name during filter-media comparison

aprilherron:
This is a 6.x version of [PR 2002](https://github.com/DSpace/DSpace/pull/2002)

While the UI usually protects against this case, it is possible to have a bitstream name with a leading space. When `MediaFilterService` attempts to check if a bitstream has already been generated for the filter, the name comparison fails due to the leading space and a duplicate bitstream is generated.

This is solved by simply trimming the name during the comparison.


terrywbrady (approved):
tested and verified

tdonohue (approved):
👍 Looks good to me.  Should also be cherry-picked or ported to `master`

PR #2017, opened 2018-04-11
drop indexes, update, recreate

kshepherd:
DSpace 6.x port of DS-3788 (master PR is https://github.com/DSpace/DSpace/pull/1912, already tested, reviewed, merged to master)


tdonohue (approved):
👍 Looks identical to #1912, which was tested, approved and merged into `master`.  Merging immediately.

PR #2018, opened 2018-04-12
DS-3886: Adopt tests to versioning and DOI-support

pnbecker:
https://jira.duraspace.org/browse/DS-3886

Three simple fixes to make DSpace's tests pass even if versioning is activited and the DOIIdentifierProvider is used. It is not necessary to cherry pick this pr, neither to dspace-6_x nor to master.

This PR was created with the support of the Technische Universität Hamburg.


kshepherd (approved):
+1 i can confirm that with versioning enabled / DOI ID + connector enabled, tests fail without this PR and pass with the PR

kshepherd:
@tdonohue i've tested this, could you give it a review and +1 / merge it?

tdonohue (approved):
👍 Looks good to me

PR #2023, opened 2018-04-15
[DS-3800] master port of pull/1914

kshepherd:
Master port of #1914 (DS-3800 The metadata.hide configuration property does not take into account the boolean value assigned to it.)
Hand edited as I couldn't get the commits in the dspace-6_x PR applied to master cleanly with rebasing / cherry-picking


tdonohue (approved):
👍 LGTM.  Same code changes as applied in #1914 (along with some minor code realignment to follow new code style rules).  Merging immediately as this is a direct port.

PR #2025, opened 2018-04-17
DS-3892 add the findBySchema name method

abollini:
https://jira.duraspace.org/browse/DS-3892

### DIFF
--- a/dspace-api/pom.xml
+++ b/dspace-api/pom.xml
@@ -81,6 +81,7 @@
                         <exclude>**/src/test/resources/**</exclude>
                         <exclude>**/src/test/data/**</exclude>
                         <exclude>**/.gitignore</exclude>
+                        <exclude>**/src/main/resources/rebel.xml</exclude>
                         <exclude>src/test/data/dspaceFolder/config/spiders/**</exclude>
                         <exclude>src/main/java/org/apache/solr/handler/extraction/ExtractingParams.java</exclude>
                     </excludes>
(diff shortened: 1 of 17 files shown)


abollini:
This is now ready to be reviewed.
Integration Tests are included (also fixing previous test that were flagged as Ignored or inaccurate).
Documentation has been added to the REST contract https://github.com/DSpace/Rest7Contract/blob/master/metadatafields.md

This PR has been rebased on top of #2073 that should be merged asap.
The most relevant change here is commit 6b74ef5 that allows us to specify the precedence between builder to be able to cleanup "complex" context (the metadatafields must be deleted before cleanup the metadataschema to avoid violation of DB constraints)

tomdesair (commented):
Thank you for the integration tests! The code looks good. I only have one remark.

I was also wondering if it would make sense to add a `fields` link to https://dspace7.4science.it/dspace-spring-rest/api/core/metadataschemas/1 ? For example

`"fields" : "/api/core/metadatafields/search/bySchema?schema=<prefix>"`

That would allow easy navigation of a REST schema object to the corresponding fields.

abollini:
Hi @tomdesair 
I have done the refactor that you have suggested to avoid code duplication.
It seems a good idea to add a "fields" link in the metadataschema HAL document but it is not trivial to do it well so I have created a separate issue for that:
Finally, I have also noted that the MetadataSchemaRestRepositoryIT was named incorrectly (was Metadataschema...) I have renamed it and enable the findOne test that currently success

tdonohue (approved):
👍 This code looks good to me, and includes passing Integration Tests.  Thanks!

PR #2026, opened 2018-04-18
[DS-3894] Fix "No enum constant" exception for facet "View More" links

aprilherron:
https://jira.duraspace.org/browse/DS-3894

The "order" parameter is being erroneously reused for "View More" links which is causing an exception to be thrown. To reproduce issue: Perform a discovery search with results, sort results asc or desc. Click a facet's "View More" link. `SearchFacetFilter` tries to interpret "asc" or "desc" as a `DiscoveryConfigurationParameters.SORT` enum element. The solution to the issue is simply to rename the parameter to something else. I renamed to "filterorder".


tdonohue (approved):
👍 Looks like an obvious fix to me.

kshepherd (approved):
+1 Tested in XMLUI, reproduced error and confirmed this PR fixes it. Code looks good. Merging. Thanks @KingKrimmson !

PR #2031, opened 2018-04-23
[DS-3895] Bitstream size can't be referenced in HQL queries

mwoodiupui:
https://jira.duraspace.org/browse/DS-3895
All automated tests succeed.


tdonohue (approved):
👍 Obvious fix, looks correct to me & tests all pass!

pnbecker (approved):
:+1:

PR #2036, opened 2018-04-24
DS-3900 admin/groups browse dont show arrows to navegate

arvoConsultores:
https://jira.duraspace.org/browse/DS-3900


ssolim:
https://github.com/DSpace/DSpace/pull/1864

this is already open for some time and could be ready to merged

PR #2038, opened 2018-04-26
DS-3507: Fixes the issues with the colon and Solr special characters …

marsaoua:
…at the end of a searching string by escaping them.

### DIFF
--- a/dspace-jspui/src/main/java/org/dspace/app/webui/discovery/DiscoverUtility.java
+++ b/dspace-jspui/src/main/java/org/dspace/app/webui/discovery/DiscoverUtility.java
@@ -17,6 +17,8 @@ import javax.servlet.http.HttpServletRequest;
 
 import org.apache.commons.lang.StringUtils;
 import org.apache.log4j.Logger;
+import org.apache.solr.client.solrj.util.ClientUtils;
+
 import org.dspace.app.webui.util.UIUtil;
 import org.dspace.content.DSpaceObject;
 import org.dspace.core.Context;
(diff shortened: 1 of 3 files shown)


pnbecker:
@terrywbrady @kshepherd I marked this for 6.3, hoping you can consider it as quit win. If 6.3 is already to close to the release, please just change the Milestone.

kshepherd:
I'm wondering if #2045 (changes solr GETs to POSTs) might also accidentally fix this issue? Seems so from my testing.
Could someone give that PR a try and let me know what you think? If we decide to go ahead with replacing GETs with POSTS in discovery, this might be a less finnicky way of resolving it (POST params don't rely on special characters as separators, etc., and i haven't triggered any errors using other special characters in values yet). If you discover any problems there, please add them to teh convo and we'll look at applying both PRs

pnbecker:
@kshepherd I haven’t tested it, but it is a problem related to solr and not to urlencoding. The colon is used to separate fields from search terms for n queries. If a colonnis not escaped at the end of the query, solr gets problems parsing the query.

kshepherd:
Thanks @pnbecker - the reason I thought the POST might fix things is not just related to urlencoding, it's the fact that when we post form key / value pairs, separator characters like colons are not used.
However, I can see in the JIRA issue (https://jira.duraspace.org/browse/DS-3507) that other characters are referred to, and Solr seems to not be very conditional as to how it interprets them, so I'll keep testing this PR too.

kshepherd (approved):
+1, I can reproduce the errors with special solr characters in dspace-6_x and the queries are successful / return expected results with this PR applied.
I haven't inspected code thoroughly so it'd be good to get an additional code inspection, but it all looks pretty sensible to me. Thanks @marsaoua !

kshepherd:
Noted in dev meeting that it would be cleaner and more compatible with a DSpace 7 forward port to handle special character encoding / escaping in the SolrServiceImpl class, rather than in the user interfaces. Depending on how/when the DSpace 7 version of this bug is fixed, we might have to revert any merges in DSpace 6 to avoid double escaping, etc.

pnbecker (approved):
+1
Thank you @marsaoua for the PR. :-)

pnbecker:
As discussed in today's dev meeting I merge this now as it is. As remarked by @kshepherd we will have to check if we run in this error again with DSpace 7. We can use this PR as blueprint solution then but will have to lookout for the right place to fix this (probably in the REST-API or DSPACE-API).

PR #2042, opened 2018-05-01
[DS-3895] Bitstream size can't be referenced in HQL queries

mwoodiupui:
https://jira.duraspace.org/browse/DS-3895
All automated tests succeed.

### DIFF
--- a/dspace-api/src/main/java/org/dspace/app/itemexport/ItemExportServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/app/itemexport/ItemExportServiceImpl.java
@@ -579,7 +579,7 @@ public class ItemExportServiceImpl implements ItemExportService
                                 for (Bitstream bitstream : bitstreams)
                                 {
                                     // add up the size
-                                    size += bitstream.getSize();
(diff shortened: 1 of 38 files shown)


mwoodiupui:
Something strange happened to Travis.  BitstreamServiceImpl doesn't contain the string "getSize", and that line is calling Bitstream.getFormat().  I've restarted the build.  I'm sure that I built this code locally, without the noted error.  Will build it again.

mwoodiupui:
I checked my local source tree:  it is in sync. with GitHub, and on the DS-3895-6x branch.  Here, the code builds without error and all tests pass.  Travis rebuilt it and got the same nonsensical error at the same place.  ???

mwoodiupui:
Aha, I had a stale version of dspace-6_x.  Upstream has a new method in BitstreamServiceImpl that uses getSize().  Rebased, testing now.

mwoodiupui:
...aaaand now I've actually committed the update.

tdonohue (approved):
👍 Obvious fix, looks correct to me & tests all pass!

pnbecker (approved):
:+1:

PR #2043, opened 2018-05-01
[DS-3694] Re-add the assembly exclusion for the Mirage 2 war

kshepherd:
The Mirage 2 war exclusion from assembly.xml was accidentally removed during port from master in PR #2014 ... since we still use XMLUI and Mirage 2 in DSpace 6, we need to add this exclusion back in so we don't unnecessarily copy a war for Mirage2 over during assembly (the theme is already included in the XMLUI war)

### DIFF
--- a/dspace/src/main/assembly/assembly.xml
+++ b/dspace/src/main/assembly/assembly.xml
@@ -124,6 +124,10 @@
          <includes>
             <include>org.dspace.modules:*:war:*</include>
          </includes>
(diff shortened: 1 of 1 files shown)


hardyoyo (commented):
+1 obvious fix, thanks!

hardyoyo (approved):

kshepherd:
thanks @hardyoyo , merging now as it's obvious quick win and we're both +1 (as is terry on the last ticket)

PR #2044, opened 2018-05-02
[DS-3901] Small bug in DefaultItemVersionProvider. deleteVersionedItem()

pnbecker:
https://jira.duraspace.org/browse/DS-3901

I did run the unit tests, but I'm not sure if this is covered at all and didn't run any manual tests. I think the change is quite obvious. If some DSpace committers agree, I don't think we need manual tests. If you want to test this, the following steps could be helpful:

1. Create three versions of an item, delete the third version, check if the second one is marked as "in_archive", and can be found in search an browsing.
2. Allow submitters to create new versions (in config/modules/versioning.cfg). Open the item in JSPUI, login with the credentials of the submitter and create a new version, but don't finish the submission process. Instead delete the item while being in submission.
3. Run a test compared to the second one, but delete the item enduring the Workflow process.

This should be cherry-picked to master.

### DIFF
--- a/dspace-api/src/main/java/org/dspace/versioning/DefaultItemVersionProvider.java
+++ b/dspace-api/src/main/java/org/dspace/versioning/DefaultItemVersionProvider.java
@@ -80,8 +80,8 @@ public class DefaultItemVersionProvider extends AbstractVersionProvider implemen
                 // version gets archived.
                 Item item = versionHistoryService.getPrevious(c, history, versionToDelete).getItem();
                 if (!item.isArchived()
-                        || workspaceItemService.findByItem(c, versionToDelete.getItem()) != null
-                        || workflowItemService.findByItem(c, versionToDelete.getItem()) != null)
+                        && workspaceItemService.findByItem(c, versionToDelete.getItem()) == null
+                        && workflowItemService.findByItem(c, versionToDelete.getItem()) == null)
                 {
                     item.setArchived(true);
                     itemService.update(c, item);


pnbecker:
In DSpace 6 we added the possibility to create new versions for submitters. The code in DSpace 5 ignores Workspace and WorkflowItems completely: https://github.com/DSpace/DSpace/blob/dspace-5_x/dspace-api/src/main/java/org/dspace/versioning/DefaultItemVersionProvider.java#L54-L62

tdonohue (approved):
👍 The if statement definitely does not match the larger comment just before it.  It seems like it's meant to be verifying that the version you are deleting is *not* a WorkspaceItem or WorkflowItem.  So, the new code is the correct logic to do that.

dspace-api/src/main/java/org/dspace/versioning/DefaultItemVersionProvider.java
```diff
@@ -80,8 +80,8 @@ public void deleteVersionedItem(Context c, Version versionToDelete, VersionHisto
                 // version gets archived.
                 Item item = versionHistoryService.getPrevious(c, history, versionToDelete).getItem();
                 if (!item.isArchived()
-                        || workspaceItemService.findByItem(c, versionToDelete.getItem()) != null
-                        || workflowItemService.findByItem(c, versionToDelete.getItem()) != null)
+                        && workspaceItemService.findByItem(c, versionToDelete.getItem()) == null
+                        && workflowItemService.findByItem(c, versionToDelete.getItem()) == null)
```
mwoodiupui: I wonder why the test is not (previous item is not archived and item to be deleted is archived).  An item which is not in a workspace, and not in a workflow, and not archived, should represent an inconsistency in the model -- every item should be in one of these three states, yes?

PR #2048, opened 2018-05-03
[DS-3902] Upgrade Flyway version to resolve dspace install issues on Windows

terrywbrady:
https://jira.duraspace.org/browse/DS-3902

### DIFF
--- a/dspace-api/pom.xml
+++ b/dspace-api/pom.xml
@@ -643,7 +643,7 @@
         <dependency>
             <groupId>org.flywaydb</groupId>
             <artifactId>flyway-core</artifactId>
-            <version>4.0.3</version>
+            <version>4.1.2</version>
         </dependency>
 
         <!-- Google Analytics -->


terrywbrady:
I have only tested on Windows so far. This needs to be verified on linux.

kshepherd (approved):
+1 for linux: tested fresh_install and explicit database migrate on ubuntu linux and it's working fine there

kshepherd:
@tdonohue commented in dev meeting:
RE:2048 The one risk to Flyway upgrades is that all Flyway upgrades are not equal... Flyway sometimes adds in new features or version incompatibilities in new releases.  So, we might want to verify in going from Flyway 4.0.3 -> 4.1.2 that we are just getting bug fixes

kshepherd:
after we noted new features and one incompatible change in 4.1.0 (https://flywaydb.org/documentation/releaseNotes#4.1.0) and uncertainty around cause/conditions of reported issue, we decided (in dev meeting) to move this to 6.4

PR #2057, opened 2018-05-12
DS-3777: Allow only POST on the authn/login endpoint.

jcreel:
Allow only POST on the authn/login endpoint.


abollini (approved):
The PR was good. I have fixed some codestyle issues, test that were broken after the change and add a specific test to check that GET is disallowed

PR #2058, opened 2018-05-12
[DS-3695] Upgrade Solr to 7.3.1.

mwoodiupui:

This splits Maven property "solr.version" into "solr.client.version" and "solr.server.version"; upgrades our use of the Solr-J client to 7.3.0; *removes the Solr server*; and as usual takes up a number of innocuous IDE hints along the way.

At some point "solr.server.version" will disappear.

### How to use it
1. Unpack Solr to somewhere convenient.  Use version 7.3.0 or newer.  Call this place SOLR_HOME.
1. Copy [DSpace]/solr/* to $SOLR_HOME/server/configsets.  Optionally remove any directories already in server/configsets.
1. Start Solr.
1. Configure DSpace:  solr.server = http://localhost:8983/solr
1. Start DSpace

As discussed 23-Jan-2019 in #dev-mtg Slack channel, this PR will focus on a `fresh_install` without annual sharding.  Subsequent PRs will fill out other cases.  To be done here:

- [x] Take up schema updates from recent Docker work.


mwoodiupui:
At this writing, all of DSpace builds successfully and all of the test suites are satisfied.  As I haven't much experience with Solr at all, I'm uncertain how to do further testing and would welcome suggestions.

mwoodiupui:
This may need some exclusions to get rid of e.g. Jetty.  I saw some downloads go flitting by the first time I built it....

mwoodiupui:
*sigh*  I have my Maven settings.xml set up to un-disable tests by default, but at home it wasn't properly doing that for ITs so they didn't run.

helix84:
A better approach regarding the schema and config files would be to start with the Solr 7 default ones and change what's needed for DSpace rather than the other way around. As far as I remember, these files contained a lot of the default junk from an older Solr version anyway.

tdonohue:
NOTE: I've started playing with updating/restarting this PR. I'll have come additional commits to add tomorrow.

I've also noticed that this **PR depends on the Solr Statistics Schema changes in #1810**, as Solr v7 seems more strict about Schema validation and immediately throws errors along the lines of "Invalid number/integer" when encountering unexpected UUIDs.

tdonohue:
@mwoodiupui : I've pushed up some bug fixes to this PR.  The PR now builds successfully and passes *most tests*. However, integration tests in the `dspace-spring-rest` project have some failures, mostly because our Solr Statistics schema is invalid (resulting in lots of `Invalid Number: [uuid]` errors).

So, this PR has now become dependent on the schema fixes in #1810.  (However, there are a few more fixes that can be made to this PR in the meantime, provided you manually apply the schema changes locally.)

tdonohue:
I've rebased this PR again (force update) to include the schema changes in #2249 (which were just merged into `master`)

mwoodiupui:
Rebased to resolve conflict.

mwoodiupui:
Another forced push.  Repair conflicts caused by merging #2241 .

mwoodiupui:
@terrywbrady please inspect commit 31becba to see if I broke anything.

terrywbrady:
@mwoodiupui , that commit looks reasonable to me.  Ping me if you would like me to run a test.

mwoodiupui:
I took a look at the latest build failure, in DiscoveryRestControllerIT.  Could someone show me where documents are cleaned out of Solr between the tests?

tdonohue:
Just a note of the status here.  This PR has a single failing test (which continues to be a bit of a mystery)

* This is the failing test: https://github.com/DSpace/DSpace/blob/master/dspace-spring-rest/src/test/java/org/dspace/app/rest/DiscoveryRestControllerIT.java#L2683
* And the failing Solr query is here: https://github.com/DSpace/DSpace/blob/master/dspace-spring-rest/src/test/java/org/dspace/app/rest/DiscoveryRestControllerIT.java#L2735-L2736

It's trying to search by `dc.date.issued`.  Earlier in that test method, it adds 3 Items with different dates, and tries to find just *one* of them.  But, instead the result is always *3* (presumably matching all 3 items, which is odd)

If anyone else watching this ticket has immediate ideas of what could be going on, please feel free to share them on this PR (as we'll continue to investigate).

mwoodiupui:
The queried field is declared as a 'text' field, which means it is tokenized on ingestion.  Putting quotes around the required value of the field turns the subsequent search into a phrase search, which finds only the correct document.  But I'm uneasy about doing it that way.  Does phrase searching really work the way we need it to for all supported queries?

Modifying the schema by adding:
+    <dynamicField name='dc.date.*'
+          type='string'
+          indexed='true'
+          stored='true'
+          multiValued='false'
+          omitNorms='true'
+          docValues='true'/>
fixes the broken test, but breaks another which wants to search on year only.  "2010" doesn't match "2010-02-13" as a plain string.

Where do we document what date precisions are supported on ingestion, what precisions on search, and what are the matching rules between dates of different precision?

tdonohue:
@mwoodiupui : While I understand your concerns, I'm starting to wonder if you are running into an *existing* DSpace issue regarding date precision.  I'm pretty sure (not 100% positive though) we treat all metadata fields identically in DSpace 6.x...so, I think dates are just treated as strings.

Your questions are good ones, and we likely can & should improve this. But, I'm wondering whether that should come in a *followup* ticket/PR (improve date precision in Solr), rather than lumping it into this Solr upgrade PR.  

So, I guess I'm wondering out loud if putting quotes around dates is "good enough" for this PR (to just get us running on latest Solr), and we can improve date precision after that?  I'd like to hear other opinions (from anyone following this discussion/ticket) though.

mwoodiupui:
Capturing a thought for later:  (1) We need a specification for DSpace "imprecise dates", as noted above; (2) We might represent imprecise dates in Solr using DateRangeField, which I have NOT tried.

mwoodiupui:
I cherry-picked #2286 in an attempt to let Travis run to completion and it seems to have done.

mwoodiupui:
Rebased to fix a conflict.  NOTE:  force-pushed the result.

mwoodiupui:
Index upgrading changes are not yet tested.

AlexanderS:
Just a note: The whole sharding stuff of the statistics core (especially `SolrLoggerServiceImpl.initSolrYearCores()` and `SolrLoggerServiceImpl.CreateCore(...)`) does not work anymore, if the solr datadir is not `[DSpace]/solr` because it is using the filesystem to detect the additional cores for the previous years.

mwoodiupui:
Correct.  I'm looking at using Time Routed Aliases instead of directly sharding the index.  That should work remotely and we won't have to do anything to create new shards.

mwoodiupui:
Above commit:  also replace deprecated Trie*Field with *PointField, which is what made those attributes irrelevant.  Deleted various comments which described field implementation classes no longer shipped with Solr, and thus had become misleading.

dspace/solr/search/conf/schema.xml
```diff
@@ -16,198 +16,32 @@
...
-       version="1.4" is Solr's version number for the schema syntax and semantics.  It should
-       not normally be changed by applications.
-       1.0: multiValued attribute did not exist, all fields are multiValued by nature
-       1.1: multiValued attribute introduced, false by default
-       1.2: omitTermFreqAndPositions attribute introduced, true by default except for text fields.
-       1.3: removed optional field compress feature
-       1.4: default auto-phrase (QueryParser feature) to off
-       1.5: omitNorms defaults to true for primitive field types (int, float, boolean, string...)
-     -->
+<schema name="discovery" version="1.5">
```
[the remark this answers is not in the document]
mwoodiupui: Yes.  I would like to say a lot more in commentary here, if only I *knew* more about the uses of some of these fields.  I will try to capture what I learn along the way.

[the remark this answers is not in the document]
tdonohue: @mwoodiupui : sounds great to me. I saw your latest commit which adds at least basic comments/notes for each. I think that's "good enough" for now.  I agree with you though, that as you learn more, it'd be good to enhance any comments with additional information.  Thanks! I'll resolve this comment.

dspace/solr/oai/conf/schema.xml
```diff
@@ -176,26 +132,24 @@
...
+   <dynamicField name="*_td" type="double"  indexed="true"  stored="true"/>
+   <dynamicField name="*_tdt" type="date"   indexed="true"  stored="true"/>
+   <dynamicField name="*_pi"  type="pint"   indexed="true"  stored="true"/>
    <dynamicField name="ignored_*" type="ignored" multiValued="true"/>
    <dynamicField name="attr_*" type="textgen" indexed="true" stored="true" multiValued="true"/>
    <dynamicField name="random_*" type="random" />
    
  </fields>
  <uniqueKey>item.handle</uniqueKey>
- <defaultSearchField>item.handle</defaultSearchField>
```
terrywbrady: Is testing needed to confirm that the defaultSearchField and defaultOperator are not needed?

dspace/solr/statistics/conf/schema.xml
```diff
@@ -186,158 +60,68 @@
...
+    <field name="owner" type="string" indexed="true" stored="true" required="false" multiValued="true" docValues="true"/>
+    <field name="submitter" type="string" indexed="true" stored="true" required="false" />
+    <field name="actor" type="string" indexed="true" stored="true" required="false" />
+    <field name="workflowItemId" type="integer" indexed="true" stored="true" required="false" />
 
  </fields>
 
  <uniqueKey>uid</uniqueKey>
 
- <!-- field for the QueryParser to use when an explicit fieldname is absent -->
```
terrywbrady: The defaultSearchField and defaultOperator change removal will need to be tested.  When DSpace used to default the operator to OR, results were not very reliable.

dspace/src/main/config/build.xml
```diff
@@ -947,94 +947,89 @@ You may manually install this file by following these steps:
         <antcall target="update_geolite" />
     </target>
 
-    <!-- Check if any Solr indexes need updating to the version of Solr/Lucene we are using. -->
+    <!-- Check if any Solr indexes need updating to the version of
```
terrywbrady: Is this upgrade working now?  It is good to see that this could work. I presume new documents inherit the the version from the schema/config files but existing index documents are aware of their old versions.

terrywbrady (commented):
The code changes look very good.  I would like to help test as well.  Let's chat about how I can help with testing.

It is good to see the removal of the large comment blocks from the Solr schemas especially if some of those comments are no longer up to date.  Those comments did provide some instruction to a developer on how to modify the file.  Would it be possible to embed a link to some Solr documentation that would provide similar examples?

dspace/src/main/config/build.xml
mwoodiupui: Good point.  This is untested.  The current focus is on fresh_install, which won't be running this target.  Arguably I should remove this change to a separate PR for supporting upgrade.

mwoodiupui:
The extensive commentary on all the things you can do with a schema is still available (with up-to-date text and examples) in [SOLR]/server/solr/configsets/_default/conf.  I will add mention of that.

dspace-spring-rest/pom.xml
```diff
@@ -320,28 +322,146 @@
...
+        <!-- The following Solr / Lucene dependencies also support integration tests -->
         <dependency>
-            <groupId>org.dspace</groupId>
-            <artifactId>dspace-solr</artifactId>
-            <classifier>classes</classifier>
+            <groupId>org.apache.solr</groupId>
+            <artifactId>solr-core</artifactId>
+            <version>${solr.client.version}</version>
             <scope>test</scope>
+            <exclusions>
```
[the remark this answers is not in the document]
mwoodiupui: Wouldn't that be nice.  I've never found a way to declare a group of artifacts that can be excluded (or included) as a unit.

dspace/solr/oai/conf/schema.xml
mwoodiupui: Well caught!  these settings no longer exist, and had to be removed from the schemas, but I need to ensure that the corresponding 'df' and 'q.op' query parameters are set when needed.

mwoodiupui: I've added code to replace these obsolete settings.

dspace/src/main/config/build.xml
mwoodiupui: I've now inadvertently tested the trivial case:  when the cores are already v7 this target correctly does not try to upgrade them.  Actually upgrading older cores remains untested.

dspace/solr/oai/conf/schema.xml
mwoodiupui: Hmmm.  Looks now as though the proper fix would be to add these parameters as defaults in the search RequestHandler, and back out the Java code changes.

dspace/solr/statistics/conf/schema.xml
mwoodiupui: Changing these was unintentional, the result of a half-done fix.

terrywbrady (changes requested):
## Changes requested
- For the search repo, the default operator is currently OR'ing terms.
- The statistics repo seems to be incorrectly formatting data values.  Here is the one record I managed to generate.

```
{
        "ip":"172.29.0.1",
        "referrer":"http://localhost:3000/items/2032e716-dfe4-4874-ba43-41c51aed1c22",
        "dns":"172.29.0.1",
        "userAgent":"Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.98 Safari/537.36",
        "isBot":false,
        "id":"java.util.UUID:a67ea636-2fcd-4236-b885-553396d41514",
        "type":0,
        "owningItem":["java.util.UUID:2032e716-dfe4-4874-ba43-41c51aed1c22"],
        "owningColl":["java.util.UUID:7ac14f26-4161-4f9e-b0ca-39720924a4e7"],
        "owningComm":["java.util.UUID:e5757f12-ab22-4ed9-b023-fe09cd741b84"],
        "time":"2019-02-15T01:40:49.986Z",
        "bundleName":["THUMBNAIL"],
        "statistics_type":"view",
        "uid":"5174c8bd-3fe2-4298-bd84-03fa3de7d894",
        "_version_":1625497115115388928}]
  }
```

## Code review
The code looks good.  The schema changes look like a good mapping.  Question: were there any dynamic fields or data types that were dropped because they were presumed to be unused?  If so, it might be good to explicitly enumerate the items that were assumed to be unneeded.

## Test results
I used DSpace 7 to test the behavior of the search, statistics, and oai cores.  I did not attempt to test authority.
- oai - I was able to import into the core and browse content with the oai web app.  I was able to filter by set.  I did not do a date-restricted query.  The uketd_dc format did filter out all objects (since no ETD's are present)
- search - keyword search and faceted searches behaved aside from the operator issue listed above
- statistics - in my usage of the angular ui, I managed to trigger the creation of only one stats record.  I presume there is not enough functionality in place yet to verify the creation and retrieval of statistics records.  See my issue above

mwoodiupui:
I did not remove any dynamicField declarations because I have no idea whether any of them is used.  I would like to figure that out someday and delete anything that we are not using.

What is incorrect about the statistics record?  I presume it is only the "java.util.UUID:" jazz.  I think I need to back out any use of the UUID type and make them strings again.  Solr is not treating declared UUID fields the way I expected.  Are there any other problems with this record?

terrywbrady:
@mwoodiupui , I did not expect the UUID to display with the class name as a prefix.  Perhaps that is the way this version of Solr renders it.  It looked like I was seeing a Java toString() formatted String rather than a value.  I can do more testing to see if that is the correct representation of a UUID in Solr.

In the search repo, the UUID's just look like strings.
```
"search.uniqueid":"2-4fee1213-f704-4d70-bf78-80d100d7dd3c",
        "search.resourcetype":2,
        "search.resourceid":"4fee1213-f704-4d70-bf78-80d100d7dd3c",
        "handle":"123456789/17",
        "location":["me5757f12-ab22-4ed9-b023-fe09cd741b84",
          "l7ac14f26-4161-4f9e-b0ca-39720924a4e7"],
        "location.comm":["e5757f12-ab22-4ed9-b023-fe09cd741b84"],
        "location.coll":["7ac14f26-4161-4f9e-b0ca-39720924a4e7"],
```

mwoodiupui:
I haven't been able to create a statistics document yet, but I think you are right:  SolrJ must've changed its treatment of a UUID object added to a string field.  I've changed the stat. code to add object.getId().toString() instead of object.getId().  (UUID overrides Object.toString() and should do what we want.)

mwoodiupui:
Never mind:  I had to wait for the auto-commit to time out, or something.  Now I have statistics.  And the UUID-valued fields are plain UUIDs.  I think I fixed it.

mwoodiupui:
In case it wasn't clear:  I have set the default search operator again in the 'oai', 'search', and 'statistics' cores.  'authority' never had one, so it uses the "default default" operator (OR).

terrywbrady:
For the external Solr service, I pulled together the documentation for building and running it with Docker

https://github.com/DSpace-Labs/DSpace-Docker-Images/issues/85

mwoodiupui:
Did I make all the requested changes?

artlowel:
Thanks @mwoodiupui!

I tested it, and it looks good. I was able to get the search working in the HAL browser and in the UI (albeit on an older commit, it doesn't look like the [metadata-as-a-map PR](https://github.com/DSpace/DSpace/pull/2287) is already merged in this PR). The results for search were identical to the old solr backend.

I was also able to generate download statistics and they look good as well. I'm still going to try to migrate old statistics to see whether that causes any issues.

terrywbrady (approved):
My prior test results looked good.  

I re-reviewed the code and it looks good.  I did not re-test.

Give me a shout if you would like me to re-test any specific aspects of this PR.

artlowel:
I tested the statistics migration by using the statistics core from demo.dspace.org.

According to [the solr docs](https://lucene.apache.org/solr/guide/7_4/indexupgrader-tool.html), you can't migrate from v4 to v7 at once, you need to migrate one version at a time, and to do so, you need a few jars from each version.

So I did the following:
```bash
# download all the needed jars from maven
wget --content-disposition "https://search.maven.org/remotecontent?filepath=org/apache/lucene/lucene-core/5.5.5/lucene-core-5.5.5.jar" "https://search.maven.org/remotecontent?filepath=org/apache/lucene/lucene-backward-codecs/5.5.5/lucene-backward-codecs-5.5.5.jar" "https://search.maven.org/remotecontent?filepath=org/apache/lucene/lucene-core/6.6.5/lucene-core-6.6.5.jar" "https://search.maven.org/remotecontent?filepath=org/apache/lucene/lucene-backward-codecs/6.6.5/lucene-backward-codecs-6.6.5.jar" "https://search.maven.org/remotecontent?filepath=org/apache/lucene/lucene-core/7.7.1/lucene-core-7.7.1.jar" "https://search.maven.org/remotecontent?filepath=org/apache/lucene/lucene-backward-codecs/7.7.1/lucene-backward-codecs-7.7.1.jar"

# run the migrations in order
java -cp lucene-core-5.5.5.jar:lucene-backward-codecs-5.5.5.jar org.apache.lucene.index.IndexUpgrader -verbose /path/to/statistics/data/index

java -cp lucene-core-6.6.5.jar:lucene-backward-codecs-6.6.5.jar org.apache.lucene.index.IndexUpgrader -verbose /path/to/statistics/data/index

java -cp lucene-core-7.7.1.jar:lucene-backward-codecs-7.7.1.jar org.apache.lucene.index.IndexUpgrader -verbose /path/to/statistics/data/index
```
As an aside: we should try to come up with a more convenient way to do the migration by the final release.

When I queried the solr 7 stats core afterwards, the number of results I ended up with was identical to the core I started from, but if I used a `rows` param greater than `0`, or tried to do a facet query, I got this exception: 
```
java.lang.AssertionError: Unexpected state. Field: &apos;stored,indexed,tokenized,omitNorms&lt;latitude:60.225693&gt;&apos;
``` 
so it has a problem with `float` fields. 

I tried changing all `float` fields to `string` in `schema.xml`, and then it complained about `date` fields, `long` fields, etc. After I replaced everything to `string` except for `boolean` fields, it worked, and I got results identical to the old core, and could facet.

@mwoodiupui, can you think of a reason for this?

mwoodiupui:
@artlowel Thank you for testing migration manually.  Your results support my growing expectation that we will have to rebuild all indexes when upgrading a site.  Solr 4 has been representing data types using classes which no longer exist in Solr 7, so those fields were reconfigured to use the current classes.  I don't know how these classes are stored in the index, but it seems reasonable that e.g. TrieFloatField would be stored differently than FloatPointField, and that FloatPointField might be unable to interpret records created using TrieFloatField.

I should note that this PR has been focused on the fresh_install case.  Your test provides valuable and timely information, but such problems should not block this PR; they are input to further work aimed at supporting existing sites.  There was some discussion of this on Slack last week.

The Ant build description includes targets for upgrading Solr cores, and there is some work in this PR to fit it for upgrading all the way to Solr 7.  That work went untested until late last week, when a problem was discovered.  Further work on automatic upgrading will take place in another PR.  But I am becoming more strongly convinced that this upgrade will require (a) reindexing 'search, and 'oai', and dumping/restoring 'authority' and 'statistics'.  Only 'authority' currently lacks the necessary tools, and I've begun work on that, which will be useful in other cases as well.

artlowel (approved):
>I should note that this PR has been focused on the fresh_install case. Your test provides valuable and timely information, but such problems should not block this PR; they are input to further work aimed at supporting existing sites

I agree. I'll add my approval.

pom.xml
```diff
@@ -26,14 +26,17 @@
...
+        <log4j.version>2.6.2</log4j.version>
         <slf4j.version>1.7.22</slf4j.version>
         <jackson.version>2.8.11</jackson.version>
         <jersey.version>2.26</jersey.version>
         <hibernate.version>5.2.8.Final</hibernate.version>
         <hibernate-validator.version>5.4.2.Final</hibernate-validator.version>
+        <!-- PIN Solr to 7.3.x until SOLR-12858 is fixed. This bug affects all integration tests that use Solr
+             https://issues.apache.org/jira/browse/SOLR-12858 -->
+        <solr.client.version>7.3.1</solr.client.version>
+        <solr.server.version>4.10.4</solr.server.version>
```
tdonohue: @mwoodiupui : This line is confusing.  Is there a reason we have an old `solr.server.version` specified here?  It looks like this is unused and can be completely removed.  But, maybe I'm missing something?

mwoodiupui: You are correct:  it is unused, and it is confusing.  I had already made a note to remove it in a subsequent PR, but I'll do that now.

tdonohue (approved):
👍 Gave this a code review, and it also looks great to me.

tdonohue:
Merging, as this is at +3.  Developers on `master` (pre-7) need to be aware that you will now need to install Solr v7 _alongside_ DSpace.  Documentation has begun at (see prerequisites and step 11):  https://wiki.duraspace.org/display/DSDOC7x/Installing+DSpace

PR #2060, opened 2018-05-14
[DS-3808] Return 204 Status when return json object is empty

terrywbrady:
https://jira.duraspace.org/browse/DS-3808

### DIFF
--- a/dspace-spring-rest/src/test/java/org/dspace/app/rest/BitstreamRestRepositoryIT.java
+++ b/dspace-spring-rest/src/test/java/org/dspace/app/rest/BitstreamRestRepositoryIT.java
@@ -347,7 +347,8 @@ public class BitstreamRestRepositoryIT extends AbstractControllerIntegrationTest
 
 
         getClient().perform(get("/api/core/bitstreams/" + bitstream.getID() + "/self"))
-                   .andExpect(status().isOk())
+                   //Does this status make sense?  This is what is being returned.
+                   .andExpect(status().isNoContent())
         ;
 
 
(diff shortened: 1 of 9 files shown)


dspace-spring-rest/src/test/java/org/dspace/app/rest/BitstreamRestRepositoryIT.java
```diff
@@ -347,7 +347,8 @@ public void findOneBitstreamRelsTest() throws Exception {
 
 
         getClient().perform(get("/api/core/bitstreams/" + bitstream.getID() + "/self"))
-                   .andExpect(status().isOk())
+                   //Does this status make sense?  This is what is being returned.
```
abollini: no this has no sense... this endpoint doesn't exist at all so a 404 should be returned. Nevertheless it is a separate issue. Can you remove the check of the /self and fire a bug report?

dspace-spring-rest/src/test/java/org/dspace/app/rest/SubmissionDefinitionsControllerIT.java
```diff
@@ -95,8 +95,8 @@ public void findByCollection() throws Exception {
     public void findCollections() throws Exception {
         //Match only that a section exists with a submission configuration behind
         getClient().perform(get("/api/config/submissiondefinitions/traditional/collections"))
-                   //The status has to be 200 OK
-                   .andExpect(status().isOk());
+                   //Does this status change make sense?
```
abollini: yes and no, acording to the rest contract https://github.com/DSpace/Rest7Contract/blob/master/submissiondefinitions.md#collections
as there is no explicit mapping toward the traditional submissionform an empty list is returned...
I'm not sure if no content is the right return here and more in general where list are expected... maybe a zero size page is a better option?

terrywbrady:
I created the following Jira tickets to capture the outstanding issues related to empty return objects.

- https://jira.duraspace.org/browse/DS-3916
- https://jira.duraspace.org/browse/DS-3915

Hopefully, these changes will be an improvement in the interim.

abollini (approved):
tested it works as expected

PR #2066, opened 2018-05-20
DS-3915 RestResourceController hides not existing endpoints under other errors

abollini:
This early PR is an attempt to try a more test driven development process.
Right now I have only created test cases to make evident the issue


abollini:
ready to test!

tdonohue (approved):
👍 Looks good to me, and it comes with Integration tests to prove the new code works.  Thanks @abollini

PR #2071, opened 2018-05-23
DS-3919: Rename identifier-service-test.xml to identifier-service.xml

pnbecker:
https://jira.duraspace.org/browse/DS-3919

Should be cherry-picked to master too.


tdonohue (approved):
👍 Obvious change, and verified by our testing framework.  Merging.

PR #2073, opened 2018-06-04
 DS-3923 manage null return in Search Methods

abollini:
This PR includes the implementation of the DS-3923 ticket and minor fixes discovered during the OR18 workshop preparation


tomdesair (approved):
Definitely an improvement and tests show that the code still works.

PR #2076, opened 2018-06-07
Configurable workflow: Enable workflow item deletion without publishing it (DS-3926)

santit96:
The whole issue is described here:
https://jira.duraspace.org/browse/DS-3926

### DIFF
--- a/dspace-api/src/main/java/org/dspace/xmlworkflow/state/actions/processingaction/AcceptEditRejectAction.java
+++ b/dspace-api/src/main/java/org/dspace/xmlworkflow/state/actions/processingaction/AcceptEditRejectAction.java
@@ -9,13 +9,23 @@ package org.dspace.xmlworkflow.state.actions.processingaction;
 
 import org.dspace.app.util.Util;
 import org.dspace.authorize.AuthorizeException;
+import org.dspace.authorize.factory.AuthorizeServiceFactory;
+import org.dspace.authorize.service.AuthorizeService;
 import org.dspace.content.DCDate;
 import org.dspace.content.MetadataSchema;
+import org.dspace.content.factory.ContentServiceFactory;
+import org.dspace.content.service.ItemService;
+import org.dspace.core.Constants;
 import org.dspace.core.Context;
 import org.dspace.xmlworkflow.factory.XmlWorkflowServiceFactory;
+import org.dspace.xmlworkflow.service.WorkflowRequirementsService;
+import org.dspace.xmlworkflow.service.XmlWorkflowService;
 import org.dspace.xmlworkflow.state.Step;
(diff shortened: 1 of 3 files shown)


J4bbi:
Closing and opening to restart tests

J4bbi:
@santit96, it's been a while since you submitted, but a belated thanks for the pr

I am having some difficulty actually testing the functionality described, I haven't used the workflow process very much

could you explain how to use the described feature in this PR?

J4bbi:
Thanks @santit96 

I don't know why but I don't get that option, I am pretty sure I have done everything correctly, I am building your branch with the two commits and using docker-compose.

image

these are the only, and standard options I get.

Is it possible this branch is missing a commit?

J4bbi:
@santit96  should there perhaps be something in either

https://github.com/DSpace/DSpace/blob/dspace-6_x/dspace-xmlui/src/main/resources/aspects/Workflow/workflow.js

or

https://github.com/DSpace/DSpace/blob/dspace-6_x/dspace-xmlui/src/main/resources/aspects/XMLWorkflow/xmlworkflow.js

?

J4bbi:
Or possibly, rather

https://github.com/DSpace/DSpace/blob/dspace-6_x/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/workflow/PerformTaskStep.java

santit96:
@J4bbi I'm pretty sure that i have pushed all the commits, but i made this PR long ago so i'll check just in case!

J4bbi:
@santit96, can you also confirm if this PR is intended for configurable xmlworkflow or the basic workflow?

see: https://wiki.lyrasis.org/display/DSDOC6x/Configurable+Workflow

santit96:
It is for configurable xmlworkflow

santit96:
I checked the feature again with the branch of this PR (also checked that i have the same commits this PR has), and everything works fine for me. I seems to me that there is a permissions issue. Perhaps the problem is that the the group assigned to WF2 Accept/Reject/Edit Metadata Step does not have DELETE permissions over the items, you can try giving DELETE permissions over the collection to that particular admin.

tdonohue:
@santit96 and @J4bbi :  I'd recommend this PR be updated to go against `main` and DSpace 7.x.  It looks, to me, like it's adding a new feature to Configurable Workflow (the ability to delete a workflowitem while it's in workflow) and XMLUI.  Since this feature doesn't exist in DSpace 7.0, adding it into 6.x this late is potentially problematic unless it's also added to 7.x.

Therefore, I'd like to ask that you consider adding this into 7.x instead of 6.x.  In the 7.0 platform, it is possible for Administrators in the system to delete any item in Workflow (which would solve the DS-3926 ticket in a different manner).  But, I don't believe it's possible yet for reviewers to delete items currently in workflow.

PR #2077, opened 2018-06-09
DS-3838 return 204 for successful logout

abollini:
Also updated the IT


dspace-spring-rest/src/main/java/org/dspace/app/rest/security/WebSecurityConfiguration.java
```diff
@@ -86,7 +87,7 @@ protected void configure(HttpSecurity http) throws Exception {
             //Configure the logout entry point
             .logoutRequestMatcher(new AntPathRequestMatcher("/api/authn/logout"))
             //When logout is successful, return OK (200) status
```
tomdesair: I would align this comment with the code

tomdesair (commented):
Looks good! I only have one minor remark.

tomdesair (approved):
Ok for me now!

PR #2096, opened 2018-06-24
Update copyright year range in LICENSE file in preparation for 6.3 release

kshepherd:


kshepherd:
merging without review as this PR just changes a '7' to an '8' and we need it in there prior to release

PR #2113, opened 2018-07-10
DS-3945: XMLUI: filtering by "starts with" field twice on browsing pages doesn't work

toniprieto:
Pull request for: https://jira.duraspace.org/browse/DS-3945

To reproduce this error:

1) Go to browse by title: http://demo.dspace.org/xmlui/browse?type=title

2) Filter by some text (by example: document)

3) Filter again by some other text (by example: test)

Result: With Mirage2 theme, the browse page goes to the browse by date because it lost the "type" parameter.

With Mirage theme, the user-submitted value "starts with" is ignored on the second search.

### DIFF
--- a/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/artifactbrowser/ConfigurableBrowse.java
+++ b/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/artifactbrowser/ConfigurableBrowse.java
@@ -412,7 +412,9 @@ public class ConfigurableBrowse extends AbstractDSpaceTransformer implements
         // Add all the query parameters as hidden fields on the form
         for (Map.Entry<String, String> param : queryParamsPOST.entrySet())
         {
-            jump.addHidden(param.getKey()).setValue(param.getValue());
+            if (!BrowseParams.STARTS_WITH.equals(param.getKey())) {
+                jump.addHidden(param.getKey()).setValue(param.getValue());
+            }
         }
 
         // If this is a date based browse, render the date navigation


J4bbi:
This is replicated in #2191.
I would prefer this PR because it uses equals method.

kshepherd:
@toniprieto sorry for letting this PR sit for so long!
is it possible that #1730 already fixes this, since the Map.put() call ensures no duplicates? Unless I'm misunderstanding the solution.
Also not sure how this PR fixes the Mirage2 (loses type) issue mentioned
cc @J4bbi

toniprieto:
@kshepherd thanks for reviewing it.

The Map.put() call ensures no duplicates of hidden fields but the problem is that the starts_with field is duplicated because another "starts_with" input text is defined in this form. This input text field should be used when the form is submitted but the hidden one is used instead. Removing the hidden field solves the problem.

For Mirage2, the behaviour is different because of this template:

https://github.com/DSpace/DSpace/blob/b0470eb59da0b43534de24021b72ea9970ab4d5a/dspace-xmlui-mirage2/src/main/webapp/xsl/aspect/artifactbrowser/browse.xsl#L58

This template is for displaying the "starts_with" input text field and the button on an input-group. The problem is that the hidden "starts_with" field is also processed by this template and others hidden fields are not displayed because this template ignores them.

Changing a bit the definition of this template should avoid this problem, but it is not needed if we remove the hidden starts_with field to solve the previous problem.

```
    <xsl:template match="dri:p[dri:field[@id='aspect.artifactbrowser.ConfigurableBrowse.field.starts_with' and @type='text']
    or dri:field/@id='aspect.discovery.SearchFacetFilter.field.starts_with'
    or dri:field/@id='aspect.administrative.WithdrawnItems.field.starts_with'
    or dri:field/@id='aspect.administrative.PrivateItems.field.starts_with']">
```

J4bbi:
Yeah, I'm for merging this one.

kshepherd:
closing and reopening to force travis CI rebuild

kshepherd (approved):
+1 as discussed, tested and reviewed, thanks @toniprieto !

PR #2116, opened 2018-07-13
[DS-3953] Externalize hard-coded list of metadata registry files in updater

mwoodiupui:
https://jira.duraspace.org/browse/DS-3953
This patch moves the list to dspace.cfg.  Lightly tested, appears to work.
Also cleans up a few annoying IDE nags and some minor logging inefficiencies.


tdonohue (approved):
👍 This looks great to me (untested, but it looks like it obviously should work).  This improvement will also likely be important to the new Configurable Entities work (see #2246), as configuring new Entities can require adding new metadata schemas/fields for those Entities.

mwoodiupui:
Build failure is caused by missing files which will be supplied by the Configurable Entities patch.  That patch should be merged before this one.

mwoodiupui:
Rebased to resolve conflict with Configurable entities.

dspace/config/dspace.cfg
```diff
@@ -766,6 +766,22 @@ org.dspace.app.batchitemimport.work.dir = ${dspace.dir}/imports
 # XSL file configured by: crosswalk.dissemination.DataCite.stylesheet file.)
 #identifier.doi.ezid.publisher = a publisher
 
+##### Registry Loader #####
+# Metadata namespaces.
```
benbosman: It would be useful to add a comment that this will import the metadata from ${dspace.dir}/config/registries/ during flyway
I assume this config change should also be documented on the wiki

benbosman (approved):
This looks good to me, I only included a small note inline about comments
I haven't tested it, but don't see any reason why it would fail, and I assume the ITs would otherwise capture it as well because they need the standard metadata

PR #2129, opened 2018-07-20
[DS-3952] Expose Request Copy

mwoodiupui:
Fixes #7299 
https://jira.duraspace.org/browse/DS-3952
Also document some existing classes that I needed to understand.

REST Contract: https://github.com/DSpace/RestContract/pull/173 (which builds off existing  https://github.com/DSpace/RestContract/blob/main/item-requests.md)

Frontend PR: https://github.com/DSpace/dspace-angular/pull/1327


dspace-spring-rest/src/main/java/org/dspace/app/rest/model/RequestItemRest.java
```diff
@@ -0,0 +1,220 @@
...
+    protected boolean accept_request;
+    protected boolean allfiles;
+
+    /**
+     * @return the bitstream requested.
+     */
+    @LinkRest(linkClass = BitstreamRest.class)
+    @JsonIgnore
+    public Bitstream getBitstream() {
+        return new BitstreamConverter().toModel(bitstream);
```
abollini: the BitstreamConverter needs to be injected

dspace-spring-rest/src/main/java/org/dspace/app/rest/model/RequestItemRest.java
```diff
@@ -0,0 +1,220 @@
...
+        this.expires = expires;
+    }
+
+    /**
+     * @return the item requested.
+     */
+    @LinkRest(linkClass = ItemRest.class)
+    @JsonIgnore
+    public Item getItem() {
+        return new ItemConverter().toModel(item);
```
abollini: please use spring to inject the ItemConverter

dspace-spring-rest/src/main/java/org/dspace/app/rest/converter/RequestItemConverter.java
```diff
@@ -0,0 +1,88 @@
...
+    public RequestItemRest fromModel(RequestItem requestItem) {
+        RequestItemRest requestItemRest = new RequestItemRest();
+        requestItemRest.setAccept_request(requestItem.isAccept_request());
+        requestItemRest.setAllfiles(requestItem.isAllfiles());
+        requestItemRest.setBitstream(bitstreamConverter.fromModel(requestItem.getBitstream()));
+        requestItemRest.setDecision_date(requestItem.getDecision_date());
+        requestItemRest.setExpires(requestItem.getExpires());
+        requestItemRest.setId(requestItem.getID());
+        requestItemRest.setItem(itemConverter.fromModel(requestItem.getItem()));
+        requestItemRest.setReq_email(requestItem.getReqEmail());
```
abollini: personally I prefer the camel case approach that is widely used in the rest of the code. Anyway, it is better than the current "JAVA" request item class that mix both approach

abollini (changes requested):
Thanks again for the PR and the additional javadoc really appreciated.
Other than the two inline comments I would like to ask for some Integration Test. Can you write an initial RequestCopyBuilder and matcher to test the findOne implementation?
if you need help on that I will be happy to assist you from the 13th Aug when I will come back to the office

dspace-spring-rest/src/main/java/org/dspace/app/rest/converter/RequestItemConverter.java
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * @author Mark H. Wood <mwood@iupui.edu>
+ */
+@Named
+public class RequestItemConverter
+        extends DSpaceConverter<RequestItem, RequestItemRest> {
+    private static final Logger LOG = LoggerFactory.getLogger(RequestItemConverter.class);
+
+    @Inject
+    protected BitstreamConverter bitstreamConverter;
```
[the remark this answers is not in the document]
mwoodiupui: Strictly speaking, @Inject = @Autowired(required=true).  I'll change it.

dspace-spring-rest/src/main/java/org/dspace/app/rest/converter/RequestItemConverter.java
```diff
@@ -0,0 +1,88 @@
...
+        requestItemRest.setReqEmail(requestItem.getReqEmail());
+        requestItemRest.setReqMessage(requestItem.getReqMessage());
+        requestItemRest.setReqName(requestItem.getReqName());
+        requestItemRest.setRequestDate(requestItem.getRequest_date());
+        requestItemRest.setToken(requestItem.getToken());
+        return requestItemRest;
+    }
+
+    @Override
+    public RequestItem toModel(RequestItemRest obj) {
```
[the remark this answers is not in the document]
mwoodiupui: Good catch.  I've added logic to handle updates.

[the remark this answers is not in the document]
mwoodiupui: Is this change request satisfied?

tdonohue (commented):
@mwoodiupui Overall, I think this is looking great! I only have a few minor notes/comments inline.  But, to me, this looks nearly ready to merge.

dspace-spring-rest/src/main/java/org/dspace/app/rest/converter/RequestItemConverter.java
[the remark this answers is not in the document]
abollini: yes, thanks

dspace-spring-rest/src/test/java/org/dspace/app/rest/builder/RequestItemHelperDAO.java
```diff
@@ -0,0 +1,40 @@
...
+import javax.persistence.Query;
+
+import org.dspace.app.requestitem.RequestItem;
+import org.dspace.core.AbstractHibernateDAO;
+import org.dspace.core.Context;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Supply missing "delete" operation on RequestItem, to support testing.
```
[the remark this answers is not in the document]
abollini: why not create it right now in the RequestItemService?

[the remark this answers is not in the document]
mwoodiupui: It seemed so strange to me that delete was not implemented, that it might have been a conscious decision to make these requests un-deletable.  This needs discussion.  I have no objection to implementing deletion, but I don't want to undo a design choice through ignorance.

dspace-spring-rest/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
```diff
@@ -0,0 +1,197 @@
...
+        fail("The test case is a prototype.");
+    }
+*/
+
+    /**
+     * Test of save method, of class RequestItemRepository.
+     */
+/*
+    @Test
+    public void testSave()
```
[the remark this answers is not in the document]
mwoodiupui: No other *RepositoryIT tests save().  Shall I just delete the test?

dspace-spring-rest/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
```diff
@@ -0,0 +1,197 @@
...
+        Class instanceClass = instance.getDomainClass();
+        assertEquals("Wrong domain class", instanceClass, RequestItemRest.class);
+    }
+
+    /**
+     * Test of wrapResource method, of class RequestItemRepository.
+     */
+/*
+    @Test
+    public void testWrapResource()
```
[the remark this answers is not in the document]
mwoodiupui: No other *RepositoryIT tests wrapResource().  Shall I just delete the test?

dspace-spring-rest/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
```diff
@@ -0,0 +1,197 @@
...
+ */
+public class RequestItemRepositoryIT
+        extends AbstractControllerIntegrationTest {
+    /** Where to find {@link RequestItem}s in the local URL namespace. */
+    public static final String URI_ROOT = REST_SERVER_URL
+            + RequestItemRest.CATEGORY + '/'
+            + RequestItemRest.NAME + 's';
+
+/*
+    @BeforeClass
```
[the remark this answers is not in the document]
mwoodiupui: Done.

dspace-spring-rest/src/test/java/org/dspace/app/rest/builder/RequestItemHelperDAO.java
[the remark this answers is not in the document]
mwoodiupui: Delete operation moved to main code.

mwoodiupui:
Rebased to cure a recent merge conflict.

dspace-spring-rest/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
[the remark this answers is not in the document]
abollini: I agree this is not something that should be tested in an integration test (it is eventually something for a UnitTest)

dspace-spring-rest/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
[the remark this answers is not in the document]
abollini: The save method is indeed to deal with PUT request (update)
https://github.com/DSpace/DSpace/blob/94aeb7c18a1177807a2444dd031e856a14c152c0/dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/DSpaceRestRepository.java#L72
Looking to the code I don't think the implementation reflect that. Instead than implement the save method you should implement the createAndReturn method
https://github.com/DSpace/DSpace/blob/94aeb7c18a1177807a2444dd031e856a14c152c0/dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/DSpaceRestRepository.java#L261

See for example how this as been done in the EPersonRestRepository
https://github.com/DSpace/DSpace/blob/94aeb7c18a1177807a2444dd031e856a14c152c0/dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/EPersonRestRepository.java#L67
and how it was tested
https://github.com/DSpace/DSpace/blob/e92852faca085b26335b2cb14dbba35c321d0ee0/dspace-spring-rest/src/test/java/org/dspace/app/rest/EPersonRestRepositoryIT.java#L52
and other create*test methods

Finally, please note that in the current dspace there is no why to update a request a copy, once done you cannot update your comment or apply the request to all the files etc. I'm not against adding such implementation but it is not strictly required so I suggest to remove it now (save implementation and test). Eventually, to test a save update your should create the request item object using a builder and send a PUT request to update some attributes. Retrieve it using GET and check that the requested changes have been applied

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/RequestItemRepository.java
```diff
@@ -0,0 +1,105 @@
...
+    protected RequestItemConverter requestItemConverter;
+
+    @Autowired(required = true)
+    protected BitstreamService bitstreamService;
+
+    @Autowired(required = true)
+    protected ItemService itemService;
+
+    @Override
+    public RequestItemRest findOne(Context context, String id) {
```
benbosman: What permissions are applicable here? Are item requests restricted to the current user?
I would also recommend to create ITs to verify permissions are applied correctly.

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/RequestItemRepository.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    @Override
+    public Page<RequestItemRest> findAll(Context context, Pageable pageable) {
+        // TODO ? There is no enumerator in RequestItemService
+        throw new UnsupportedOperationException("Not supported yet.");
+    }
+
+    @Override
+    public RequestItemRest save(Context context, RequestItemRest ri) {
```
benbosman: What permissions are applicable here? Are item requests restricted to the current user or can you request a copy for a different user?

Which parameters are mandatory when creating the request? Can you create a request without specifying the email address?

I would also recommend to create ITs to verify permissions are applied correctly.

@abollini also mentioned in 
https://github.com/DSpace/DSpace/pull/2129/files#r239523981 the save method can is used for updates, the createAndReturn is used for creating new requests

dspace-spring-rest/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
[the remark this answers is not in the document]
benbosman: Agreed with @abollini

benbosman (changes requested):
@mwoodiupui Thanks for the PR. I think this is almost ready. I have included a few comments inline.
The feedback is primarily tailored to permissions, the creation of a new request, and ITs

dspace-spring-rest/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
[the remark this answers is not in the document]
mwoodiupui: The (commented out) test has been moved to a unit test suite.

dspace-spring-rest/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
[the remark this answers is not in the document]
mwoodiupui: save() replaced with a new createAndReturn().

tdonohue:
@mwoodiupui : Just a note that this work is now "high priority" for 7.1.  Are you interested in updating this old PR to get it into a reviewable state?  Or would you rather I reassign this to someone who can start from your PR?

(Sidenote, it appears at a glance this PR also might solve #2886 )

mwoodiupui:
I'm halfway through rebasing, but it appears that much has changed since 2019.

[time passes] First rough results of the latest rebase.  I've used DefaultProjection in places where it wasn't obvious how to get a Projection.

mwoodiupui:
Still to do:
- [ ] Add permission checks
- [ ] Test permission checks
- [ ] Which request parameters are mandatory?

mwoodiupui:
So far, the only nod to authorization that I've seen is that the user making a request must be logged in.  After that the plan seems to be security-by-obscurity:  only a designated person receives the request, thus only that person possesses the token that identifies the request.  But the token is an `int`, which makes guessing tokens rather simple.  We may want to follow up in a future feature request by hardening the design in some way.

dspace-server-webapp/src/main/java/org/dspace/app/rest/repository/RequestItemRepository.java
```diff
@@ -0,0 +1,119 @@
...
+
+    @Autowired(required = true)
+    protected ItemService itemService;
+
+    @Autowired(required = true)
+    protected RequestItemConverter requestItemConverter;
+
+    @Override
+    //@PreAuthorize(expr)
+    public RequestItemRest findOne(Context context, String id) {
```
tdonohue, line 57: @mwoodiupui : Regarding your previous comment about security, additional security could/should be added here.  This likely should have a `@PreAuthorize("hasAuthority('AUTHENTICATED')")` -- to verify the user is authenticated.   We also should have a test (in the IT) which verifies that an unauthenticated user cannot access any individual requests.

(Sidenote: in your comment you said the token is an `int`, but here it's shown to be a String.  It appears to me it's actually a random hex key, see https://github.com/DSpace/DSpace/blob/main/dspace-api/src/main/java/org/dspace/app/requestitem/RequestItemServiceImpl.java#L45)

mwoodiupui: Thanks for the suggested expression.

v6 behavior seems to have been to send the user to the login page if not logged in.

I don't know where I got the `int`, unless it is the debug log message in `RequestItemServiceImpl#createRequest` which says it created a "requestitem_token" followed by the RequestItem's ID (which is Integer).  (I'm going to fix that.)  Anyway, random hex is at least a good deal harder to guess.

mwoodiupui: @tdonohue @benbosman I've been re-thinking the permission issue.  https://wiki.lyrasis.org/display/DSDOC7x/Request+a+Copy describes the feature.  The whole point of Request a Copy is to cause DSpace to punt the access control decision to a human.  The request form collects contact information, so we don't need an EPerson to supply it.  Request a Copy operates outside of DSpace's normal access control -- the request is routed to someone determined by the DSpace configuration, and granting it causes the file(s) to be sent.

All that to say this:  probably this feature should *not* interact with DSpace access control at all.

I must've been tired when I wrote my previous post.  A new look at [the XMLUI code ](https://github.com/DSpace/DSpace/blob/da3154b04511389b3886207ed91d0d18a0adaaa4/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/artifactbrowser/ItemRequestForm.java#L150) seems to show that, if the session is not logged in, the request form will provide a link to login *if desired*, but the form is presented; however, if the session *is* logged in *and* has READ access to the Item, the browser is redirected to the item page.  Unauthenticated users, and authenticated users without READ, get the form and may use it to send a request to bypass DSpace access control.

tdonohue: @mwoodiupui : In the DSpace 7 realm, I'm a little worried about providing an unauthenticated REST endpoint where anyone/anything can kickoff a ton of "requests for copies" anonymously (that seems like an invitation for a DOS attack).  That's why I wonder if it's more appropriate to place this behind a requirement to authenticate, even if it slightly changes how it works.  

So, for instance:
1. Request a Copy can only be submitted by authenticated users.  They can perhaps change their email address if they want the response sent back to a different email address. 
2. Request a Copy can be approved by anyone with the generated (random) link (no need to login to approve).

That would in turn mean that the `findOne()` method needs to allow for anonymous access (for a user to be able to approve it anonymously).  So, that means it should have a `@PreAuthorize("permitAll()")`.

However, the `createAndReturn()` method would need to require authentication...i.e. `@PreAuthorize("hasAuthority('AUTHENTICATED')")`... to force you to be authenticated before you can create a new "request for copy".

mwoodiupui: That change is not "slight", IMHO.  As I recall, requiring the requester to have an account was specifically rejected during the initial discussion of alternatives to this feature.  What is the typical use case?  (Ours is precisely the case of anonymous users to be vetted by those controlling access.)  I think such changes need wider discussion.

We might be better off with something like a rate limit.

tdonohue: @mwoodiupui : That's reasonable then.  To be honest, now that I think about this again, since it's a POST request, I'm realizing this `createAndReturn` would require CORS & CSRF to pass....so, the requests must already be coming from a "trusted" source/origin, which also must be passing a valid CSRF token.  This provides us with a bit of a safety net already that it's not just anyone from anywhere generating these "requests for copies".

So, I'm OK with your approach. I'd just ask that you explicitly add a `@PreAuthorize("permitAll()")` to both methods and add ITs which prove the behavior you expect.  If we run into later issues with this behavior we can fix it at a later time.

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
```diff
@@ -0,0 +1,368 @@
...
+        final String uri = URI_ROOT + '/' + request.getToken();
+        getClient(authToken).perform(get(uri))
+                   .andExpect(status().isOk()) // Can we find it?
+                   .andExpect(content().contentType(contentType))
+                   .andExpect(jsonPath("$", Matchers.is(
+                       RequestCopyMatcher.matchRequestCopy(request))));
+
+        // Clean up.
+        bitstream.setDeleted(true);
+        context.restoreAuthSystemState();
```
tdonohue, line 117: I don't believe the cleanup is needed...Builders should cleanup after themselves.

Also, please move the `context.restoreAuthSystemState()` up to *before* the `getClient()` call.  This just helps us ensure that we are always turning auth back on BEFORE tests run.   In general, turn the auth back on as soon as you can, which usually is right after you create any test content you need.

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
```diff
@@ -0,0 +1,368 @@
...
+        final String uri = URI_ROOT + '/' + request.getToken();
+        getClient().perform(get(uri))
+                   .andExpect(status().isOk()) // Can we find it?
+                   .andExpect(content().contentType(contentType))
+                   .andExpect(jsonPath("$", Matchers.is(
+                       RequestCopyMatcher.matchRequestCopy(request))));
+
+        // Clean up.
+        bitstream.setDeleted(true);
+        context.restoreAuthSystemState();
```
tdonohue, line 156: Same as previous.  Auth needs to be turned back on sooner.  Cleanup likely isn't necessary.

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
```diff
@@ -0,0 +1,368 @@
...
+        rir.setItemId(item.getID().toString());
+        rir.setRequestEmail(RequestItemBuilder.REQ_EMAIL);
+        rir.setRequestMessage(RequestItemBuilder.REQ_MESSAGE);
+        rir.setRequestName(RequestItemBuilder.REQ_NAME);
+        rir.setAllfiles(false);
+
+        // Create it and see if it was created correctly.
+        ObjectMapper mapper = new ObjectMapper();
+        String authToken = getAuthToken(admin.getEmail(), password);
+        MvcResult mvcResult = getClient(authToken)
```
tdonohue, line 195: This entire `getClient()` call can be simplified.  Returning the entire `mvcResult` uses a lot of extra time/memory...and it's not needed when all you want is the token created. 

Here's an example of how to do this much cleaner:
https://github.com/DSpace/DSpace/blob/main/dspace-server-webapp/src/test/java/org/dspace/app/rest/CollectionGroupRestControllerIT.java#L159-L165

The key here is you want to use an `andDo()` clause to parse the value out of the response **without** having to save the entire response to a variable.  You can find many more examples in the codebase by searching for "andDo("

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
```diff
@@ -0,0 +1,368 @@
...
+                        hasJsonPath("$._links.self.href", not(is(emptyOrNullString())))
+                )))
+                .andReturn();
+
+        // Clean up the created request.
+        String content = mvcResult.getResponse().getContentAsString();
+        Map<String,Object> map = mapper.readValue(content, Map.class);
+        String requestToken = String.valueOf(map.get("token"));
+        RequestItem ri = requestItemService.findByToken(context, requestToken);
+        requestItemService.delete(context, ri);
```
tdonohue, line 219: The cleanup can also be simplified if you use an `andDo()` obviously.

I'd also recommend moving the deletion logic to a static method like `RequestItemBuilder.deleteRequestItem([requestToken])`, as this would better mirror other tests.  See for example how ItemBuilder has an `ItemBuilder.deleteItem([uuid])` method for deleting a specific item.

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
```diff
@@ -0,0 +1,368 @@
...
+        rir.setBitstreamId(bitstream.getID().toString());
+        rir.setItemId(item.getID().toString());
+        rir.setRequestEmail(RequestItemBuilder.REQ_EMAIL);
+        rir.setRequestMessage(RequestItemBuilder.REQ_MESSAGE);
+        rir.setRequestName(RequestItemBuilder.REQ_NAME);
+        rir.setAllfiles(false);
+
+        // Create it and see if it was created correctly.
+        ObjectMapper mapper = new ObjectMapper();
+        MvcResult mvcResult = getClient()
```
tdonohue, line 260: Same as above... use an `andDo()` clause to simplify

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
```diff
@@ -0,0 +1,368 @@
...
+                .andReturn();
+
+        // Clean up the created request.
+        String content = mvcResult.getResponse().getContentAsString();
+        Map<String,Object> map = mapper.readValue(content, Map.class);
+        String requestToken = String.valueOf(map.get("token"));
+        RequestItem ri = requestItemService.findByToken(context, requestToken);
+        requestItemService.delete(context, ri);
+
+        context.restoreAuthSystemState();
```
tdonohue, line 286: Same here, move to Builder delete method.  Also move the auth restore up to the top after you create content.

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
```diff
@@ -0,0 +1,368 @@
...
+
+    /**
+     * Verify that Spring Security's CSRF protection is working as we expect.
+     * We must test this using a simple non-GET request, as CSRF Tokens are not
+     * validated in a GET request.
+     *
+     * @throws java.lang.Exception passed through.
+     */
+    @Test
+    public void testRefreshTokenWithInvalidCSRF()
```
tdonohue, line 308: This method needs renaming & comments changed.  It's a test to verify that an item request cannot be made with an invalid CSRF token.  It's not about "RefreshToken"

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
```diff
@@ -0,0 +1,368 @@
...
+                        .andExpect(status().isNoContent());
+    }
+
+    /**
+     * Verify that Spring Security's CORS settings are working as we expect.
+     *
+     * @throws java.lang.Exception passed through.
+     */
+    @Test
+    public void testCannotReuseTokenFromUntrustedOrigin()
```
tdonohue, line 344: Same here, rename the test.  It's not about Token reuse...it's verifying that an item request cannot be made from an untrusted origin.

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
```diff
@@ -0,0 +1,368 @@
...
+        String token = getAuthToken(eperson.getEmail(), password);
+
+        // Remove "Bearer " from that token, so that we are left with the token itself
+        token = token.replace("Bearer ", "");
+
+        // Save token to an Authorization cookie
+        Cookie[] cookies = new Cookie[1];
+        cookies[0] = new Cookie(AUTHORIZATION_COOKIE, token);
+
+        getClient().perform(post(URI_ROOT)
```
tdonohue, line 320: This POST appears to be posting nothing... what does it normally return when you post nothing? We should perhaps have a separate test with an invalid POST to show that it does NOT throw a "Forbidden" error (I doubt it would, but having a test that shows how an invalid POST would respond would be nice anyways)

tdonohue (changes requested):
@mwoodiupui : Thanks for picking this back up & getting it ready for review.  I gave this a code review today, and overall it looks good.  But, as I've noted below, some of your code needs to be cleaned up based on some established "best practices" (in other existing classes).  Additionally, I'd like this PR to implement the endpoint at `/api/tools/itemrequests`, as that's the agreed endpoint name in our REST Contract: https://github.com/DSpace/RestContract/blob/main/item-requests.md

(Sidenote, we also should verify the REST Contract is still accurate based on this PR...as I expect it needs some updating/correcting based on this implementation.)

Overall, even though I added a lot of comments below, I'm in favor of the direction here...just trying to help you clean things up & align this code with other similar code in the codebase.  Thanks again!

dspace-server-webapp/src/test/java/org/dspace/app/rest/matcher/RequestCopyMatcher.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @author Mark H. Wood <mwood@iupui.edu>
+ */
+public class RequestCopyMatcher {
+    private RequestCopyMatcher() { }
+
+    public static Matcher<? super Object> matchRequestCopy(RequestItem request) {
+        return allOf(
+                //hasJsonPath("$._links.bitstream", Matchers.not(Matchers.empty())),
+                //hasJsonPath("$._links.item", Matchers.not(Matchers.empty())),
```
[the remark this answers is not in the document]
mwoodiupui: Done.

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
mwoodiupui: Yes, that should be posting a fresh item request.

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
mwoodiupui: Done.

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
mwoodiupui: Oops, copy/paste error.  Done, but I don't see the issue with the comments.

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
mwoodiupui: Done.

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
mwoodiupui: Done.

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
mwoodiupui: Done.

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
mwoodiupui: Isn't the MvcResult built anyway, and merely a reference returned?  I don't see how we save anything but a single store.

However, to make it more like the code elsewhere, done.

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
mwoodiupui: Done.

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
mwoodiupui: Right.  Done.

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
```diff
@@ -0,0 +1,368 @@
...
+/**
+ *
+ * @author Mark H. Wood <mwood@iupui.edu>
+ */
+public class RequestItemRepositoryIT
+        extends AbstractControllerIntegrationTest {
+    /** Where to find {@link RequestItem}s in the local URL namespace. */
+    public static final String URI_ROOT = REST_SERVER_URL
+            + RequestItemRest.CATEGORY + '/'
+            + RequestItemRest.NAME + 's';
```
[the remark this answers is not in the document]
mwoodiupui: Done

dspace-server-webapp/src/main/java/org/dspace/app/rest/repository/RequestItemRepository.java
```diff
@@ -0,0 +1,147 @@
...
+        ri.setAccept_request(false); // Not accepted yet.  Must set:  DS-4032
+        requestItemService.update(ctx, ri);
+        return requestItemConverter.convert(ri, new DefaultProjection());
+    }
+
+    // NOTICE:  there is no service method for this -- requests are never deleted?
+    @Override
+    public void delete(Context context, String token)
+            throws AuthorizeException, RepositoryMethodNotImplementedException {
+        throw new RepositoryMethodNotImplementedException("RequestItemRest", "delete");
```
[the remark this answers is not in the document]
mwoodiupui: Done.

dspace-server-webapp/src/main/java/org/dspace/app/rest/repository/RequestItemRepository.java
```diff
@@ -0,0 +1,147 @@
...
+            return null;
+        } else {
+            return requestItemConverter.convert(requestItem, new DefaultProjection());
+        }
+    }
+
+    @Override
+    public Page<RequestItemRest> findAll(Context context, Pageable pageable) {
+        // TODO ? There is no enumerator in RequestItemService
+        throw new UnsupportedOperationException("Not supported yet.");
```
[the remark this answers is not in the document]
mwoodiupui: Done.

dspace-server-webapp/src/main/java/org/dspace/app/rest/model/RestModel.java
```diff
@@ -33,6 +33,8 @@
     public static final String AUTHORIZATION = "authz";
     public static final String VERSIONING = "versioning";
     public static final String AUTHENTICATION = "authn";
+    public static final String COPY_REQUEST = "copy_request";
```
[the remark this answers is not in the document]
mwoodiupui: Right.  Removed.

dspace-server-webapp/src/main/java/org/dspace/app/rest/model/RequestItemRest.java
```diff
@@ -0,0 +1,221 @@
...
+ *
+ * @author Mark H. Wood <mwood@iupui.edu>
+ */
+@LinksRest(links = {
+    @LinkRest(name = "bitstream", method = "getUuid"),
+    @LinkRest(name = "item", method = "getUuid")
+    })
+public class RequestItemRest
+        extends BaseObjectRest<Integer> {
+    public static final String NAME = "copyrequest";
```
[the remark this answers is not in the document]
mwoodiupui: Sensible.  Done.

mwoodiupui:
Yup, three years ago this was to be my first attempt to learn my way around the REST code, so I'm not surprised that I missed some conventions and such.

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
tdonohue: @mwoodiupui : a while back we found out that Tests which accessed MvcResult directly ran _much slower_ than those that used `andDo()`.  So, somehow, the `andDo()` is somehow much more performant.  I admit, I don't know the *why*, but I do know that's why we adopted this "best practice" & avoid accessing MvcResult directly.

mwoodiupui:
I'm seeing consistent ConcurrentModificationException failures in the IT.  Several tests fail and, curiously, they all fail trying to delete the *same* Item (judging by the identical UUIDs).  Your insights welcomed.

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
```diff
@@ -0,0 +1,393 @@
...
+                        hasJsonPath("$.requestName", is(RequestItemBuilder.REQ_NAME)),
+                        hasJsonPath("$.allfiles", is(false)),
+                        hasJsonPath("$.requestDate", not(is(emptyOrNullString()))), // TODO should be an ISO datetime
+                        hasJsonPath("$._links.self.href", not(is(emptyOrNullString())))
+                )))
+                .andDo((var result) -> saveToken(read(result.getResponse().getContentAsString(), "token")))
+                .andReturn();
+
+        // Clean up the created request.
+        RequestItemBuilder.deleteRequestItem(requestToken);
```
tdonohue, line 274: @mwoodiupui : I _think_ the issue is with your `andDo()` syntax.  I'm not sure this this syntax is correct -- and it may result in some ItemRequests not being deleted, which seems to be the root issue. 

Other examples in our codebase use this syntax instead...
```
AtomicReference<String> idRef = new AtomicReference<>();
getClient().perform(post(
....
.andDo(result -> idRef.set(read(result.getResponse().getContentAsString(), "token")));
```
Then, you get the value out of the `idRef` variable like this...
```
// I'd recommend doing a try/finally to ensure this deletion ALWAYS runs
} finally {
    RequestItemBuilder.deleteRequestItem(idRef.get());
}
```

Essentially, my suspicion here is that this cleanup isn't working as you expected. I'd bet that `requestToken` ends up not being the value you expect and something is not being deleted.

tdonohue: More background info on my suspicions... in checking the test failures, I saw the **first** failure came out of your `testCreateAndReturnAuthenticated` test method.  It appeared to show that the CollectionBuilder was trying to cleanup an Item, and that Item couldn't be deleted cause it was still related to an ItemRequest.

So, I started looking at ItemRequest deletion, as that sort of failure implies that one (or more) of those requests are "hanging around" even after the RequestItemBuilder ran it's cleanup task.

When these sorts of failures occur, I find that the **first** failure is _usually_ the most significant one...the remainder are usually just side effects of the first one.

mwoodiupui: I made your suggested changes.  Tests still broke.  Other than atomicity, I think my code was doing the same thing as the suggestion, but at least now mine looks familiar if you know the others.

The actual problem was that RequestItemService.deleteRequestItem has its own Context and did not commit its work.

tdonohue: Good catch then on finding the real problem!  Sorry if I misunderstood what your code was doing...it just wasn't the normal syntax I was familiar with.

tdonohue:
@mwoodiupui : Does this also fix #2886 ? It appears to...though it's unclear how the expiration date is currently set/saved (I need to dig more into the code though)

mwoodiupui:
@tdonohue No, it does not add any behavior with respect to expiration of requests.  It *supports* expiration by retaining the `expires` field of the request, which was always defined but apparently never implemented.  Actually expiring requests is a new feature, and thus I avoided it for 7.0 development.  RequestItemService offers a `delete` method which should facilitate the removal of expired (or completed) requests.

mwoodiupui:
See also #7417.

dspace-server-webapp/src/main/java/org/dspace/app/rest/repository/RequestItemRepository.java
```diff
@@ -0,0 +1,146 @@
...
+        ri.setAccept_request(false); // Not accepted yet.  Must set:  DS-4032
+        requestItemService.update(ctx, ri);
+        return requestItemConverter.convert(ri, new DefaultProjection());
+    }
+
+    // NOTICE:  there is no service method for this -- requests are never deleted?
+    @Override
+    public void delete(Context context, String token)
+            throws AuthorizeException, RepositoryMethodNotImplementedException {
+        throw new RepositoryMethodNotImplementedException(RequestItemRest.NAME, "delete");
```
tdonohue, line 139: It looks like you've implemented a delete service method in `RequestItemService.delete()`.  Should we therefore update this REST endpoint to support deletion?  That'd also seemingly allow us to close #7417 

If there's a reason you'd rather not implement deletion in this PR, that's OK.  Was just noting the discrepancy between this method and the new `RequestItemService.delete()` method.

dspace-server-webapp/src/main/java/org/dspace/app/rest/repository/RequestItemRepository.java
```diff
@@ -0,0 +1,146 @@
...
+
+            boolean allFiles = rir.isAllfiles();
+
+            String email = rir.getRequestEmail();
+            if (isBlank(email)) {
+                throw new IncompleteItemRequestException("A submitter's email address is required");
+            }
+
+            String username = rir.getRequestName();
+            String message = rir.getRequestMessage();
```
tdonohue, line 119: Since these are user inputs, I think `message`, `username` and maybe even `email` should all be HTML encoded, to avoid any risk of an XSS style attack. I don't think any of these fields should support HTML/Javascript anyways, so best to encode or strip it before we even create the request.

(I guess this encoding could also be put into the `createRequest()` call below. Either way is fine.)

tdonohue (commented):
@mwoodiupui : Gave this another review today. Overall, it looks good now! The tests also prove that it works as described (I admit, I haven't tested it as this might be more easily done during development of the UI portion, to make sure both parts work together.)

I did notice a few small improvements that could be made (see inline below).

Also, I noticed that the corresponding REST Contract is a bit outdated now: https://github.com/DSpace/RestContract/blob/main/item-requests.md  It'd be good if you could update this with information about how to create a new `itemrequest` (and possibly how to delete one if you plan to add that endpoint -- see my inline comment below).

Beyond that, I'm in approval of the direction here. Looks good.

dspace-server-webapp/src/main/java/org/dspace/app/rest/repository/RequestItemRepository.java
mwoodiupui: This would allow anyone to delete any request, if he can guess a token.  Item requests bypass the usual DSpace access controls -- that's what they are for.  I doubt that requests would ever be deleted via REST calls.

We need some discussion of how, and whether, requests are deleted.  I would suggest discussion in #7417.

The `RequestItemService.delete()` method was needed for testing.

dspace-server-webapp/src/main/java/org/dspace/app/rest/repository/RequestItemRepository.java
mwoodiupui: Would you (or anyone) please point me toward an example.

tdonohue: @mwoodiupui : it's likely sufficient to use one of the `StringEscapeUtils` methods here... likely `StringEscapeUtils.escapseHtml4()`: https://commons.apache.org/proper/commons-text/javadocs/api-release/org/apache/commons/text/StringEscapeUtils.html

This would just ensure that any attempts to insert HTML/JS in these fields would be turned instead into HTML entities, which should ensure they aren't rendered on any client.

Another option would be to see if we can use something like Apache Tika to *strip* the HTML out (I've never tried this, but seems possible): https://stackoverflow.com/a/12259801/3750035

UPDATE: Just realized we also already have a `Utils.addEntities()` method to do HTML escaping. So, you could just use that. https://github.com/DSpace/DSpace/blob/main/dspace-api/src/main/java/org/dspace/core/Utils.java#L270

mwoodiupui: So, since I haven't found any other cases in dspace-server-webapp to study, let me see if I have this right:

- The contract should be updated to show that these fields are percent-encoded in flight.
- RequestItemRepository.createAndReturn must percent-DEcode them when creating the database row.
- RequestItemConverter.convert must percent-ENcode them when creating a RequestItemRest.

tdonohue: @mwoodiupui : 
* The contract doesn't necessarily need to mention that the fields are encoded/decoded (it's up to you).  But, it DOES need to be updated to fully describe the endpoint. Currently, the contract is missing information about how to create a new Item Request.
* The end goal here is that we need to find a way to protect ourselves from XSS **because these endpoints allow for anonymous item request submissions**.   Ideally, the endpoint should NOT accept HTML at all -- it should only accept plain text.  So, this means either we strip any HTML tags...or we simply encode everything (which will invalidate HTML tags but potentially make the result look odd if someone attempts to submit an HTML formatted request).  It seems to me that we should never decode it, as doing so could result in the user reviewing the Item Request accidentally triggering an XSS attack (if the decoded request contains HTML/Javascript which then ends up being accidentally executed in the UI)

Does that help at all?

mwoodiupui: So these fields should be HTML-encoded *in the database*?  That would be a change to the dspace-api layer and require fixup of existing requests during upgrade.

I really would like to look at some other code where we are already doing this.

mwoodiupui: I'm beginning to wonder how this XSS would work.  Requests never appear in the UI.  Submission of the request form results in a text/plain email to the appropriate party.  This email contains a link which uses an opaque token to identify the request.  Following the link results in another text/plain email being returned to the requester, indicating acceptance or rejection and (in the case of acceptance) with the requested bitstream(s) attached.

I also begin to think that validating addresses and stripping markup would be more appropriate than this encoding that I still don't understand.

tdonohue: @mwoodiupui : I sent a private response in Slack as well... but, my assumption was that Item Requests *do* get displayed in the UI in order to be reviewed/approved.  We had a ticket for that here: https://github.com/DSpace/dspace-angular/issues/779  You are correct that, if these requests _never appear in a UI_ then XSS isn't possible.  I had just been working on the assumption that the request needs to appear in the UI to be approved... in which case, we need to ensure that the Item Request cannot trigger an XSS attack (since that request can come from an anonymous user).

I admit, I am not very familiar with this feature overall...so, if there's no need for it to *ever appear* in the UI, then we can end this XSS discussion.  However, should we ever decide it needs a UI component (even to delete old requests or similar from an Admin), then we need to ensure safety from XSS at that time.

mwoodiupui:
As discussed in today's meeting:  the helpdesk workflow is not yet supported, but may land in 7.2.

tdonohue (commented):
@mwoodiupui : I re-reviewed this today & tested creating an itemrequest via `curl` and it worked.  The code looks good overall as well, and I only have minor comments (see inline below).

I'm assuming this PR might still need to be updated to support `PUT` (for accepting/approving requests) as designed in your contract (https://github.com/DSpace/RestContract/pull/173)?   If so, I'd like to wait to submit approval until that's ready too.  If you plan on implementing that *separately* though, then let me know (as I'm +1 this current PR once minor updates are made).

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
```diff
@@ -0,0 +1,485 @@
...
+
+    /**
+     * Test of findOne method, with an authenticated user.
+     *
+     * @throws java.lang.Exception passed through.
+     */
+    @Test
+    public void testFindOneAuthenticated()
+            throws Exception {
+        System.out.println("findOne (authenticated)");
```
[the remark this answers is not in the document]
mwoodiupui: NetBeans puts these in when it generates a test class.  I find that it helps me sort out where a test method's output begins, particularly in the large volume of testing output generated (necessarily) by Spring MVC.  I would really rather leave them in.

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
```diff
@@ -0,0 +1,485 @@
...
+            // Test bad email
+            rir.setRequestEmail("<script>window.location='http://evil.example.com/';</script>");
+            getClient(authToken)
+                    .perform(post(URI_ROOT)
+                            .content(mapper.writeValueAsBytes(rir))
+                            .contentType(contentType))
+                    .andExpect(status().isUnprocessableEntity());
+        } finally {
+            // Clean up the created request.
+            RequestItemBuilder.deleteRequestItem(requestTokenRef.get());
```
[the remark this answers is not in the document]
mwoodiupui: Yes, in fact `requestTokenRef` is never filled so the cleanup does nothing.

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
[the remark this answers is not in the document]
mwoodiupui: But I have removed an unnecessary stack trace left over from debugging.

dspace-server-webapp/src/main/java/org/dspace/app/rest/repository/RequestItemRepository.java
mwoodiupui: I think that XSS has been addressed now.

dspace-api/src/main/java/org/dspace/app/requestitem/RequestItemServiceImpl.java
```diff
@@ -78,4 +98,28 @@ public void update(Context context, RequestItem requestItem) {
...
+
+    @Override
+    public boolean isRestricted(Context context, DSpaceObject o)
+            throws SQLException {
+        List<ResourcePolicy> policies = authorizeService
+                .getPoliciesActionFilter(context, o, Constants.READ);
+        for (ResourcePolicy rp : policies) {
+            if (resourcePolicyService.isDateValid(rp)) {
+                return false;
+            }
```
tdonohue, line 121: This seems odd to me, but maybe I'm misunderstanding. It looks like you are looping through READ policies to see if *any* have valid dates (or no dates attached at all). If they are valid, you assume it's not restricted?  What about a READ policy with a valid date which limits to a specific, non-Anonymous group?  In that scenario, it seems your `isRestricted()` would return false, when the object is restricted to a specific group.

Maybe I'm misunderstanding the logic here though... or maybe this method should be renamed so it doesn't imply that it's checking whether the object is access restricted (e.g. could be called `hasValidReadPolicies()`)

dspace-server-webapp/src/main/java/org/dspace/app/rest/utils/RegexUtils.java
```diff
@@ -36,4 +40,10 @@ private RegexUtils(){}
      */
     public static final String REGEX_REQUESTMAPPING_IDENTIFIER_AS_DIGIT = "/{id:\\d+}";
 
+    /**
+     * Regular expression to accept a string of 32 hexadecimal digits.
+     */
+    public static final String REGEX_REQUESTMAPPING_IDENTIFIER_AS_HEX32
+            = "/{id:[0-9a-zA-Z]{32}}";
```
tdonohue, line 47: This looks to be matching *too much*.  Aren't any alphabetic letters past F invalid in hexidecimal?  I think this should be more like
`/{id:[0-9a-fA-F]{32}`

I should also point out here that it looks like it _could be possible_ for a very large, 32-digit number to match both `REGEX_REQUESTMAPPING_IDENTIFIER_AS_DIGIT` and `REGEX_REQUESTMAPPING_IDENTIFIER_AS_HEX32`.  But, I'm not sure how to work around that, and the likelihood seems very slim that would ever occur (as numerical identifiers are sequential in DSpace)

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
```diff
@@ -0,0 +1,518 @@
...
+                .writer()
+                .writeValueAsString(parameters);
+
+        // Send the request.
+        String authToken = getAuthToken(eperson.getEmail(), password);
+        getClient(authToken).perform(put(URI_ROOT + '/' + itemRequest.getToken())
+                .contentType(contentType)
+                .content(content))
+                .andExpect(status().isOk()
+                // .andMore?
```
tdonohue, line 504: If it's possible to do so, we may want to verify that the PUT resulted in the updates we expected...not just that it returned 200.  So, is there any way to verify that the ItemRequest was approved at the Java API layer, or via another `GET` call after this `PUT`?

Also, we should add a few more `testPut()` methods here which verify incorrect behavior throws the expected errors.  For instance, we should verify that an anonymous user, or different authenticated user cannot "approve" the request.

dspace-server-webapp/src/main/java/org/dspace/app/rest/repository/RequestItemRepository.java
```diff
@@ -0,0 +1,296 @@
...
+     */
+    private String getLinkTokenEmail(String token)
+            throws URISyntaxException, MalformedURLException {
+        String base = configurationService.getProperty("dspace.server.url");
+
+        URI link = new URIBuilder(base)
+                .setPath("/api/" + RequestItemRest.CATEGORY
+                        + '/' + RequestItemRest.PLURAL_NAME)
+                .addParameter("token", token)
+                .build();
```
tdonohue, line 292: I think this link is incorrect, as we'll need to create a form/page on the UI for the user to read the request and respond.  This currently is a REST API link, which won't take the user anywhere useful.

I think you'll need to work with @artlowel to craft a UI URL where the user should be sent if they want to respond to an Item Request.  That UI URL will obviously need to be passed the token, so that it can lookup the request (via the REST API) and eventually craft the proper `PUT` to respond to the request.

tdonohue (changes requested):
@mwoodiupui : Gave this another code review today.  The code is still looking good overall, though I have some comments on the newly added `PUT` functionality and a request for a few more `PUT` tests (to verify proper errors occur).  

I also tested this for the first time today with the UI PR from @artlowel : https://github.com/DSpace/dspace-angular/pull/1327

The process of *creating* an item request (and sending the email to the author) works perfectly.  However, the response back isn't yet working because you and @artlowel will need to talk about what UI page you need to send the user to, so that @artlowel can build that page and send the `PUT` back.  See further comments inline below.

Overall, mostly it's good, just needs finishing work & ensuring that your PR and the UI PR are "synced" up properly.

dspace-server-webapp/src/main/java/org/dspace/app/rest/utils/RegexUtils.java
mwoodiupui: Right, thank you for catching that.  Indeed we only need a-fA-F.

32 hex digits (16 bytes) is much larger than a Java `long`.  It could only be held in a BigInteger.  I studied the problem for some time before deciding that I could do no better than this.

dspace-api/src/main/java/org/dspace/app/requestitem/RequestItemServiceImpl.java
mwoodiupui: The logic here is copied from https://github.com/DSpace/DSpace/blob/dspace-5_x/dspace-jspui/src/main/java/org/dspace/app/webui/util/RequestItemManager.java and I didn't like it either.

But consider:  the point of an item request is to bypass the normal DSpace authorization and let a human make the decision.  The intent of the code which calls this method (which I also copied from JSPUI logically if not quite verbatim) seems to be that if all bitstreams of an item are requested, a bitstream is not of an internal type (like licenses) and the bitstream is restricted *for anyone* (not necessarily everyone) then it should be attached to the email.  IOW we don't send copies of things that an anonymous user could get for himself.

I think that XMLUI did not make this distinction -- it exists there but is commented out.  Maybe we should just drop the distinction, and the method which embodies it.

dspace-server-webapp/src/main/java/org/dspace/app/rest/repository/RequestItemRepository.java
mwoodiupui: I do need to coordinate with the UI code, but it's a bit more complicated.  Since this link appears in an email, it will result in a GET request.  That means we need to link to the back-end, to bring up the UI at the approval page with the token pre-loaded.  THAT page will do the POST which accepts/rejects the request.  On the back-end, I need to see how similar things are done during user registration, and do likewise.  Thanks for reminding me that this part of the process is not yet coded.  Sorry that I requested re-review too soon!

tdonohue: @mwoodiupui : I'm not sure I understand why the request-a-copy email needs to send a link to the backend, as that seems to imply you'd still need to redirect to the UI at some point.  

Instead, it seems like the flow here could be more similar to the Registration process, which sends you an email after you fill out a registration form.  In the email it sends, it links you back to the *User Interface* (see https://github.com/DSpace/DSpace/blob/main/dspace-api/src/main/java/org/dspace/eperson/AccountServiceImpl.java#L260), which then allows you to complete your registration. 

I *think* you might be able to do the same, if you agree on a URL path with @artlowel .  It could just be a path like `${dspace.ui.url}/request-a-copy?token={token}` ... and then we'd just need to add code to the UI to ensure when that path is called, it uses the `token` to load that request's information via a GET request to the backend.

artlowel: `${dspace.ui.url}/request-a-copy?token={token}` sounds just fine. Let's go with that.

artlowel: On second thought, let's make it `${dspace.ui.url}/request-a-copy/{token}` that way it'll work similar to the registration url, and it'll be slightly easier to parse in the UI

artlowel: So e.g. `https://demo7.dspace.org/request-a-copy/7177942B`

dspace-server-webapp/src/test/java/org/dspace/app/rest/RequestItemRepositoryIT.java
mwoodiupui: Approval can be tested at the API layer.  I've added that, and tests of bad PUTs.

dspace-api/src/main/java/org/dspace/app/requestitem/RequestItemEmailNotifier.java
```diff
@@ -0,0 +1,227 @@
...
+
+        // Which Item?
+        Item item = ri.getItem();
+
+        // Fill the message's placeholders.
+        message.addArgument(bitstreamName);     // {0} bitstream name or "all"
+        message.addArgument(item.getHandle());  // {1} Item handle
+        message.addArgument(ri.getToken());     // {2} Request token
+        message.addArgument(ri.getReqName());   // {3} Requester's name
+        message.addArgument(ri.getReqEmail());  // {4} Requester's address
```
tdonohue, line 203: I could be mistaken, but isn't it a bit odd to have the the request for open access coming from the person who submitted the Item Request?  It seems like this request should be from the *author* (who checked the box in the form) to the submitter or site admin?

dspace-api/src/main/java/org/dspace/app/requestitem/RequestItemServiceImpl.java
tdonohue: I'm honestly still a bit confused by it.  That said, it sounds like it serves a purpose & aligns with JSPUI behavior...so, I'm not against leaving it here, it just seems very oddly named. It's obviously not checking if something "isRestricted".  It seems to be checking if an object has  (any) READ policies that are all still "valid" (unexpired), which is a completely different concept than checking whether the object is access restricted.

tdonohue (changes requested):
@mwoodiupui : Gave this another test today with the UI PR.  Mostly, it looks to be working well & I appreciate the new tests you added. A few small things I ran into inline below...

Beyond that, one other thing I noticed which I found slightly odd.  Should it be possible to *change* the accept/request status of an Item Request?  The reason I ask is that I noticed that I can do this via the UI:

1. As an anonymous user, submit an item request
2. As an Admin click the request link in the email & respond to the request (clicking "accept" for example)
3. Then, as an Admin click the request link in the email a second time & respond to the request again (clicking either option & sending another email).

I guess the question is, should you be allowed to respond multiple times?  Or maybe that's something that is just possible until we implement a way to delete item requests?

(This is an honest question as I'm not sure how this behaved in 6.x.  While this could be annoying if you respond multiple times, it seems like a bug we also could simply solve later on if needed, once we support the ideas of request expiration & deletion.)

Beyond that, this is looking great.  Thanks again for the hard work here @mwoodiupui !

mwoodiupui:
I think that the back end should reject an attempt to grant/deny a request that already has a `decisionDate`.  Separately, the UI could disable such attempts, since `decisionDate` is available.  Thanks for catching this.

Concerning `isRestricted()`:  after it is working as previously implemented, we ought to ask ourselves "how *should* this work?"  The design of item request seems un-polished in some areas.

tdonohue (approved):
👍 Thanks @mwoodiupui !  Retested, and it all looks good now. I can no longer update a previously updated request (though the UI needs to handle this better, but I'll bring that to the UI ticket). All other issues have been resolved.

artlowel (approved):
Thanks @mwoodiupui!

PR #2137, opened 2018-07-26
[DS-3967] 4x - Migrate Dockerfile to DSpace/DSpace

terrywbrady:

Derived from https://github.com/DSpace/DSpace/pull/2136 using dspace-build instead of dspace-installer

### Testing Instructions
Since Flyway DB was not part of DSpace 4x, the database schema does not automatically load in DSpace 4x.  You must manually load it into the database container using the following source file: https://github.com/DSpace/DSpace/blob/dspace-4_x/dspace/etc/postgres/database_schema.sql


ppmdo:
Tested on Fedora Linux 27. Build and compose works without problems. 

The only "drawback" is the need to install the database schema. I did this by opening a shell inside the DB container and then downloading the SQL file and running it with psql.

Used this compose file: https://github.com/DSpace-Labs/DSpace-Docker-Images/blob/master/docker-compose-files/dspace-compose/docker-compose.yml

pnbecker (approved):
:+1:
reviewed, not tested.

tdonohue (approved):
👍 Looks good to me. Merging

PR #2142, opened 2018-08-03
DS-2675: Bugfixing: Jump to value general errors with order (for master)

terrywbrady:
port https://github.com/DSpace/DSpace/pull/1730/ minus xmlui

Changes are untested


tdonohue (approved):
👍 Looks good to me. Changes here are identical to those in #1730 (minus the XMLUI changes, obviously).  Merging.  Thanks @terrywbrady

PR #2143, opened 2018-08-04
[DS-3332] Handle resolver is hardcoded in org.dspace.handle.UpdateHandlePrefix (for master)

terrywbrady:
port https://github.com/DSpace/DSpace/pull/1839 minus xmlui


tdonohue (approved):
👍 Looks good to me. Changes are identical to #1839 (without the XMLUI changes). Thanks!

PR #2144, opened 2018-08-04
[DS-3377] Solr queries too long (change search GET requests to POST) (for master)

terrywbrady:
Port https://github.com/DSpace/DSpace/pull/2045.

This is untested.  I performed a quick migration to start the conversation.


tdonohue (approved):
👍 Looks good to me. The code changes here are not identical to #2045, but these classes have changed a bit on `master`.  I double checked against the master code, and I think this looks to be correct.

PR #2145, opened 2018-08-04
DS-3498 quick fix. Disable full text snippets in search results & add warning (for master)

terrywbrady:
Port https://github.com/DSpace/DSpace/pull/2069.


tdonohue (approved):
👍 Looks good to me. Exact port of #2069

PR #2147, opened 2018-08-04
DS-3616] Fix nested vocabulary search (master)

terrywbrady:
Port https://github.com/DSpace/DSpace/pull/1770/.  Untested


tdonohue (approved):
👍 Looks good to me. Identical changes to #1770. Merging

PR #2149, opened 2018-08-04
DS-3702 & DS-3703: Rebuild the old behavior of bitstreams during vers… (for master)

terrywbrady:
Port https://github.com/DSpace/DSpace/pull/1883 for master.  Not tested


tdonohue (approved):
👍 Looks good to me. Same code changes as in #1883.

PR #1891, opened 2017-11-27
DS-3769 Set the right hibernate property of …

marsaoua:
... org.dspace.eperson.Subscription: ePerson instead of eperson.id

PR #2150, opened 2018-08-04
DS-3769 Set the right hibernate property of … (for master)

terrywbrady:
Port https://github.com/DSpace/DSpace/pull/1891


tdonohue (approved):
👍 LGTM. Same as #1891

PR #2151, opened 2018-08-04
[DS-3822] Don't guess XML structure during ingest (for master)

terrywbrady:
Port https://github.com/DSpace/DSpace/pull/1941/files


tdonohue (approved):
👍 LGTM. Obvious, identical to #1941

PR #2153, opened 2018-08-06
DS-3856 - foreignkey-constraint community2community_child_comm_id_fkey (for master)

terrywbrady:
Port https://github.com/DSpace/DSpace/pull/1973


tdonohue (approved):
👍 Looks good to me. Identical to #1973. Merging

PR #2155, opened 2018-08-06
DS-3768 Fixes the harvest solr parse error by (for master)

terrywbrady:
Port https://github.com/DSpace/DSpace/pull/1890


tdonohue (approved):
👍 Looks good to me.  Identical changes to #1890 (except for the JSPUI fixes)

PR #2160, opened 2018-08-07
DS-3796 Fix missing bitstream detection and reporting for Checksum Checker

aprilherron:
Currently, Checksum Checker fails to correctly process missing bitstreams. If the bitstream is not found, it just uses the historical checksum for comparison instead of marking the bitstream as missing. In addition, the MostRecentChecksumDAOImpl date queries had start and end date parameters reversed.

https://jira.duraspace.org/browse/DS-3796

Fixes #7143


toniprieto (approved):
I have tested it and I can confirm that this solves the two issues. After applying the patch the checksum checker detects the missing files and checker-emailer command reports it correctly.

PR #2164, opened 2018-08-10
Ds 3974

dimitrove:


dimitrove:
Had to be to dspace-6_x and not to master.

PR #2165, opened 2018-08-10
[DS-3974] Orphaned files in assetstore after REST update

dimitrove:
https://jira.duraspace.org/browse/DS-3974

### DIFF
--- a/dspace-api/src/main/java/org/dspace/storage/bitstore/BitstreamStorageServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/storage/bitstore/BitstreamStorageServiceImpl.java
@@ -90,6 +90,22 @@ public class BitstreamStorageServiceImpl implements BitstreamStorageService, Ini
     @Override
     public UUID store(Context context, Bitstream bitstream, InputStream is) throws SQLException, IOException
     {
+		if (bitstream.getInternalId() != null) {
+	        // Clone the bitstream to keep a pointer to the old file in assetstore
+	        Bitstream clonedBitstream = bitstreamService.clone(context, bitstream);
+			clonedBitstream.setDeleted(true);
+	        try {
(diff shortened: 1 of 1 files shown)


abollini:
Please check my comment on the jira issue, my first guess is that this feature did not have to be implemented...
with the proposed fix I guess (I haven't test just inspected the code) that a new bitstream is created and the previous one deleted this mean that the new bitstream will have no metadata and format set (this should be a further bug)...

I suggest to update the (old) rest contract specifying that it is equivalent to a delete followed by a creation and update the code to call delete / create

tdonohue (changes requested):
Thanks @dimitrove for your contribution!  However, as I just [noted in the corresponding ticket](https://jira.duraspace.org/browse/DS-3974?focusedCommentId=60356&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-60356), I recommend fixing this bug in the `BitstreamResource.updateBitstreamData()` method that implements this concept in the DSpace 6.x REST AP:

https://github.com/DSpace/DSpace/blob/dspace-6_x/dspace-rest/src/main/java/org/dspace/rest/BitstreamResource.java#L543

This idea of "replacing" a bitstream should likely be simply a creation of a new bitstream, followed by the deletion of the old one.  That's much safer than attempting to simply replace the "bits".  Replacing the "bits" directly could cause other issues in DSpace, especially since the CheckSum Checker would immediately notice that the file changed checksums.

(As discussed in the ticket, I don't think we'll retain this functionality in the DSpace 7 REST API.  Instead, we'll recommend a bitstream create followed by a delete.  However, as this functionality exists already in DSpace 6, I agree that we should ensure it works properly and doesn't orphan files.)

dimitrove:
Wrong fix.

PR #2168, opened 2018-08-13
[DS-3980] Defend against bad MIME type in Email.addAttachment

mwoodiupui:
https://jira.duraspace.org/browse/DS-3980
Something is passing unparseable MIME types.  This will log it and evade the problem by using the 'application/octet-stream' type instead.
This patch is untested.  An essentially identical version has been tested on 5_x.  If anyone has figured out how to reproduce the problem at will, please note it in the Jira ticket linked above.


tdonohue (approved):
👍 Seems reasonable to me.  Ideally, we might add Unit Tests here, but I'm pretty sure this is an area of the codebase where tests are lacking in general.

pnbecker (approved):
:+1:

PR #2173, opened 2018-08-16
DS-3904 - Add support for startsWith parameters in the browse endpoint

ppmdo:
Support for startsWith implemented for Items and Entries under the Browse endpoint.

Integration Tests pending.


abollini:
@ppmdo  please ping me when the PR is ready to be reviewed. The code changes seem obvious so we only need additional integration tests to approve it. A PR to update the REST contract will be appreciated as well

ppmdo:
@abollini Integration tests written.

abollini:
This PR has been discussed in the 23rd Aug DSpace 7 meeting https://wiki.duraspace.org/display/DSPACE/2018-08-23+DSpace+7+Working+Group+Meeting @ppmdo has kindly accepted to further investigate about the use of the different Browse methods in the existent UIs

ppmdo:
@abollini After further inspection I've found that the problem may reside with the paging implementation and how it relates to browse results., not with the methods of the browse repositories. 

Here's what I've found:
https://wiki.duraspace.org/pages/viewpage.action?pageId=104562967

tomdesair:
The Browse endpoint works via the generic `RestResourceController` which does not take into account any query parameters when creating the pagination links. You'll need to extend that class to also re-use any existing query parameters. 

The help you get started, I think you need to modify the following line https://github.com/DSpace/DSpace/blob/c0f31e889fd1c8eb180159bef354950ae217ace4/dspace-spring-rest/src/main/java/org/dspace/app/rest/RestResourceController.java#L628 with

```
                UriComponentsBuilder uriComponentsBuilder = linkTo(this.getClass(), apiCategory, English.plural(model)).slash(id)
                        .slash(rel).toUriComponentsBuilder();
                
                //TODO we need to strip any existing paging information from the query string
                uriComponentsBuilder.query(request.getQueryString());
                
                Link link = new Link(Link.REL_SELF, uriComponentsBuilder.toUriString());
```

But as the TODO mentions, it's important that we don't re-use the pagination query parameters because then they will be added twice.

ppmdo:
@tomdesair

ppmdo:
Should be ready to merge.

tdonohue (approved):
👍 The code here looks good to me.  I admit, I haven't had a chance to test it, but it has its own integration tests that also look good.

dspace-spring-rest/src/main/java/org/dspace/app/rest/RestResourceController.java
```diff
@@ -678,8 +678,11 @@ public ResourceSupport findRel(HttpServletRequest request, HttpServletResponse r
                     if (Page.class.isAssignableFrom(linkMethod.getReturnType())) {
                         Page<? extends RestModel> pageResult = (Page<? extends RestAddressableModel>) linkMethod
                                 .invoke(linkRepository, request, uuid, page, projection);
-                        Link link = linkTo(this.getClass(), apiCategory, model).slash(uuid).slash(subpath)
-                                .withSelfRel();
+
+                        //TODO we need to strip any existing paging information from the query string
+                        //Even if the paging info is not removed, the params don't get duplicated.
+                        Link link = linkTo(this.getClass(), apiCategory, model).slash(uuid)
+                                .slash(subpath + '?' + request.getQueryString()).withSelfRel();
```
abollini: if the query string is null or empty the question mark should be not added

dspace-spring-rest/src/test/java/org/dspace/app/rest/BrowsesResourceControllerIT.java
```diff
@@ -498,4 +499,226 @@ public void testPaginationBrowseByDateIssuedItems() throws Exception {
...
+                   .andExpect(jsonPath("$.page.number", is(0)))
+
+                   //Verify that the index filters to the "Computing'" items.
+                   .andExpect(jsonPath("$._embedded.browseEntries",
+                                       contains(BrowseEntryResourceMatcher.matchBrowseEntry("Computing", 3)
+                                       )))
+                   //Verify that the startsWith paramater is included in the links
+                    .andExpect(jsonPath("$._links.self.href", containsString("?startsWith=C")));
+
+        // ---- BROWSES BY ITEM ----
```
abollini: it is better to create separate test for that, so that a failure will be more easily to understand

abollini (changes requested):
can you also add a test to document the expected behavior when both the startsWith and the page parameter are present? which one has the precedence? is this a valid request or will be refused?
I'm fine with both solutions so I just recommend to document the implemented behavior

dspace-spring-rest/src/main/java/org/dspace/app/rest/RestResourceController.java
ppmdo: Added new test -- startsWithandPage

dspace-spring-rest/src/test/java/org/dspace/app/rest/BrowsesResourceControllerIT.java
ppmdo: Separated tests for item and entry browses.

ppmdo:
@abollini Fixed the empty or null querystring handling, separated item and entry browses tests and added a new test to verify when both startsWith and Page parameters are submitted by the user.

abollini (approved):

abollini:
code looks good and it comes with the required ITs

PR #2177, opened 2018-08-21
fix for DS-3986 -- 'View' statistics are not logged to solr in dspace 5.9

tantz001:
https://jira.duraspace.org/browse/DS-3986


tantz001:
Whoops, this should be for https://jira.duraspace.org/browse/DS-3986

PR #2179, opened 2018-08-22
DS-3905 Implement a generic uuid lookup endpoint (DSpaceObjectService based)

abollini:
This is a revision of #2176 to use a DSpaceObjectService to lookup for the UUID instead than SOLR merging the approach of #2176 with the original one in #2125

### DIFF
--- a/dspace-spring-rest/src/main/java/org/dspace/app/rest/UUIDLookupRestController.java
+++ b/dspace-spring-rest/src/main/java/org/dspace/app/rest/UUIDLookupRestController.java
@@ -13,6 +13,7 @@ import java.io.IOException;
 import java.net.URI;
 import java.sql.SQLException;
 import java.util.Arrays;
+import java.util.UUID;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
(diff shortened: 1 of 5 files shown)


tomdesair (commented):
Looks good and the functionality is covered by integrations tests.

I only have one small (non-blocking) remark.

tdonohue (approved):
👍 This looks great to me.

dspace-spring-rest/src/main/java/org/dspace/app/rest/UUIDLookupRestController.java
```diff
@@ -0,0 +1,108 @@
...
+            throws IOException, SQLException, SearchServiceException {
+
+        Context context = null;
+        try {
+            context = ContextUtil.obtainContext(request);
+            for (DSpaceObjectService dSpaceObjectService : contentServiceFactory.getDSpaceObjectServices()) {
+                DSpaceObject dso = dSpaceObjectService.find(context, uuid);
+                if (dso != null) {
+                    DSpaceObjectRest dsor = converter.convert(dso);
+                    URI link = linkTo(dsor.getController(), dsor.getCategory(), English.plural(dsor.getType()))
```
tomdesair: Can you use `getTypePlural()` here? https://github.com/DSpace/DSpace/blob/master/dspace-spring-rest/src/main/java/org/dspace/app/rest/model/RestModel.java#L35

abollini: done

tomdesair (approved):
Thx!

PR #2181, opened 2018-08-28
[DS-3990] Curation task runs need per-run parameters.

mwoodiupui:
https://jira.duraspace.org/browse/DS-3990
Lightly tested.

This patch defines "run parameters" which can be set on a Curator instance, and adds a CLI option to set them.  Adding '-p name=value' or '--parameter name=value' to a 'dspace curate' command will make 'value' available under 'name' just as if it were a configured task property.  (See [Task Properties](https://wiki.duraspace.org/display/DSDOC6x/Curation+System#CurationSystem-TaskProperties).)  The option may be specified more than once to define multiple parameters.

Task code need not change -- Curator.taskProperty(name) now searches the run parameters for 'name' before looking up a configured property.  (Thus, if you specify a parameter with the same name as a property, the parameter overrides the property.)

The patch includes CuratorTest for automated testing, and org.dspace.ctask.testing.PropertyParameterTestingTask for manual checkout.  Both produce console output with details of how they were invoked and what property values they found.


mwoodiupui:
Note that DSpace master currently has no GUI for curation, so there are no patches for it here.

mwoodiupui:
There is ugly linkage between CuratorTest and the task it uses for testing.  I had trouble with the way that tasks are instantiated by reflection, which throws an exception when asked to deal with an inner class.

dspace-api/src/test/java/org/dspace/curate/CuratorTest.java
```diff
@@ -0,0 +1,413 @@
...
+        instance.curate(context, dso);
+
+        // Check the result.
+        System.out.format("Task %s result was '%s'%n",
+                TASK_NAME, instance.getResult(TASK_NAME));
+        System.out.format("Task %s status was %d%n",
+                TASK_NAME, instance.getStatus(TASK_NAME));
+        assertEquals("Unexpected task status",
+                Curator.CURATE_SUCCESS, instance.getStatus(TASK_NAME));
+        assertEquals("Wrong run parameter", RUN_PARAMETER_VALUE, runParameter);
```
terrywbrady: How does runParameter get set?  I expected the curation task to be queried for this value.

terrywbrady (approved):

dspace-api/src/main/java/org/dspace/curate/CurationCli.java
```diff
@@ -104,6 +109,19 @@ public static void main(String[] args) throws Exception {
...
+            for (String parameter : line.getOptionValues('p')) {
+                String[] parts = parameter.split("=", 2);
+                String name = parts[0];
+                String value;
+                if (parts.length > 1) {
+                    value = parts[1];
+                } else {
+                    value = "true";
+                }
+                parameters.put(name, value);
```
benbosman: Wouldn't it make sense to trim both `name` and `value` here in case the data is entered as e.g. `'NAME = VALUE'`?

benbosman (approved):
This looks good to me, I included one suggestion inline
I was also wondering why there were so many commented tests in `CuratorTest`.

dspace-api/src/main/java/org/dspace/curate/CurationCli.java
mwoodiupui: That can only happen if the entire expression is quoted:  ``-p 'foo = bar'``, which would be rather silly, but this is still a sensible suggestion.  (Tested:  unquoted spaces are argument delimiters, so ``-p foo = bar`` results in an option 'p' with value 'foo' and two non-option arguments '=' and 'bar'.)

My test classes tend to have commented-out tests because my IDE creates them all but I have developed the bad habit of only implementing the ones I need at the time.  For now I will confirm my habit by removing all of the unused methods.

PR #2183, opened 2018-08-28
DS-3994 Authority ID & Confidence during CSV batch import not respected

### LINKED ISSUE #7341: [DS-3994] Authority ID & Confidence during CSV batch import not respected
<i>Imported from JIRA <a href="https://jira.lyrasis.org/browse/DS-3994">[DS-3994]</a> created by andrewwood</i>
<p>Issue: Providing authority ID and confidence during CSV batch import was not being respected.</p>

<p>Given CSV input: <a href="https://pastebin.com/z4hnGWBC" class="external-link" target="_blank" rel="nofollow noopener">https://pastebin.com/z4hnGWBC</a></p>

<p>We can create cases where authority ID and confidence are not being respected.</p>

<p>See:</p>

<p><a href="https://github.com/DSpace/DSpace/blob/master/dspace-api/src/main/java/org/dspace/app/bulkedit/MetadataImport.java#L1041" class="external-link" target="_blank" rel="nofollow noopener">https://github.com/DSpace/DSpace/blob/master/dspace-api/src/main/java/org/dspace/app/bulkedit/MetadataImport.java#L1041</a></p>

<p>Should be replaced with something more like:</p>

<p>String mdf = md.contains(":") ? StringUtils.substringAfter(md, ":") : md;</p>

<p> </p>

<p>(Credit: to Chris Herron)</p>

<p> </p>

AndrewZachWood:
Fixes #7341 
JIRA Post:

Issue: Providing authority ID and confidence during CSV batch import was not being respected.

Given CSV input: [https://pastebin.com/z4hnGWBC](https://pastebin.com/z4hnGWBC)

We can create cases where authority ID and confidence are not being respected.

See:

[https://github.com/DSpace/DSpace/blob/master/dspace-api/src/main/java/org/dspace/app/bulkedit/MetadataImport.java#L1041](https://github.com/DSpace/DSpace/blob/master/dspace-api/src/main/java/org/dspace/app/bulkedit/MetadataImport.java#L1041
)
Should be replaced with something more like:

`String mdf = md.contains(":") ? StringUtils.substringAfter(md, ":") : md;
`
(Credit: to Chris Herron)

### DIFF
--- a/dspace-api/src/main/java/org/dspace/app/bulkedit/MetadataImport.java
+++ b/dspace-api/src/main/java/org/dspace/app/bulkedit/MetadataImport.java
@@ -1173,7 +1173,7 @@ public class MetadataImport
      */
     private static boolean isAuthorityControlledField(String md)
     {
-        String mdf = StringUtils.substringAfter(md, ":");
+        String mdf = md.contains(":") ? StringUtils.substringAfter(md, ":") : md;
         mdf = StringUtils.substringBefore(mdf, "[");
         return authorityControlled.contains(mdf);
     }


helix84:
Sorry, on my phone. Is the : hardcoded or should this support authorityseparator from bulkedit.cfg?

nwoodward (approved):
👍 Tested and confirmed it works. 

Note: It fixes a logical bug in `isAuthorityControlledField` where `StringUtils.substringAfter()` returns an empty string if ":" is absent, which means the function is always returning false now.  So this fix works if the input CSV contains the "ORCID:" prefix in the column header, as specified in the documentation [1], or if it's absent, as in the CSV provided with the PR.

[1] https://wiki.lyrasis.org/display/DSDOC6x/ORCID+Integration#ORCIDIntegration-EditingexistingitemsusingBatchCSVEditing

uofmsean:
This also resolves the case with compareAndUpdate() that [removes the authority prefix and colon](https://github.com/DSpace/DSpace/blob/7cebc65a313066f0a79019b5c8db77bf1d8e2ac4/dspace-api/src/main/java/org/dspace/app/bulkedit/MetadataImport.java#L662-L664) from md and then later calls [!isAuthorityControlled(md)](https://github.com/DSpace/DSpace/blob/7cebc65a313066f0a79019b5c8db77bf1d8e2ac4/dspace-api/src/main/java/org/dspace/app/bulkedit/MetadataImport.java#L694-L700) that will always then return true causing the authority and confidence values to be ignored.

PR #2190, opened 2018-08-29
DS-1915: Allow to move collections using the CommunityFiliator

AlexanderS:
The community-filiator cli tool currently can only manage the relations between communities. This pull request adds the possibility to use handles or database ids of collections for the child object. With this pull request it is possible to remove and set the parent relations for collections just like for communities before.

This is a fix for the CLI part of [DS-1915](https://jira.duraspace.org/browse/DS-1915).

Note: This makes two additional methods of Collection public, but the appropriate methods of Community are public, too. So I think this should not be a problem.

### DIFF
--- a/dspace-api/src/main/java/org/dspace/administer/CommunityFiliator.java
+++ b/dspace-api/src/main/java/org/dspace/administer/CommunityFiliator.java
@@ -18,8 +18,11 @@ import org.apache.commons.cli.HelpFormatter;
 import org.apache.commons.cli.Options;
 import org.apache.commons.collections4.CollectionUtils;
 import org.dspace.authorize.AuthorizeException;
+import org.dspace.content.Collection;
 import org.dspace.content.Community;
+import org.dspace.content.DSpaceObject;
 import org.dspace.content.factory.ContentServiceFactory;
+import org.dspace.content.service.CollectionService;
 import org.dspace.content.service.CommunityService;
 import org.dspace.core.Constants;
 import org.dspace.core.Context;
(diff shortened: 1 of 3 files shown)


tdonohue:
@AlexanderS : could you provide some basic documentation (for testers) for how to now use this commandline tool to move collections or communities?  Can it move both collections and communities, or just collections?

AlexanderS:
@tdonohue The community-filiator cli tool is not new. It is already documented here: https://wiki.duraspace.org/display/DSDOC6x/Managing+Community+Hierarchy

You already can move communities with this tool. This pull request only adds the possibility to use a collection as a child object. The tool continues to work like documented:

You need to remove the old parent first:

```
dspace community-filiator -r -p oldParentID -c collectionID
```

and then set the new parent:

```
dspace community-filiator -s -p newParentID -c collectionID
```

The workflow is a bit strange, it seems that the tool was designed to support multiple parents of a single community/collection. While this may be supported by the backend, it does not seem to be supported by the UI. But this pull request does not modify the interface, it only adds the possibility to manage collections.

pnbecker (approved):
Not tested, code changes look fine to me. +1

oscar-escire:
@tdonohue @AlexanderS @pnbecker: I know this PR hasn't been updated in quite some time, but we recently received a request to reorganize a repository where, in some cases, moving a collection from one community to another would be the most feasible option rather than creating a new one and transferring all the items (the only option currently available in DSpace to perform this action).

We found this PR, ran the necessary tests, and it works exactly as expected:

1. We had the following structure in DSpace
Captura de pantalla 2026-06-10 a la(s) 17 19 57

2. We ran the commands to defiliate and to afiliate the community to the new parent:
Captura de pantalla 2026-06-10 a la(s) 17 22 40
Captura de pantalla 2026-06-10 a la(s) 17 23 48

3. After reindexing the site (as described in the original documentation), you can see that the collection has been successfully moved to its new home:
Captura de pantalla 2026-06-10 a la(s) 17 24 51

I believe this PR could be very useful for website administrators and during update and restructuring processes, since reorganizations and restructurings have become relatively common since the implementation of DSpace 7.x, 8.x, and 9.x. 

+1

oscar-escire (approved):

tdonohue:
Closing & reopening to trigger a re-run of all automated tests.

@AlexanderS : If you would have time to rebase this against the latest `main` branch, it would also help us to get this merged (especially since it now has two +1 votes).

AlexanderS:
Rebased to main without any conflicts, but I currently do not have a way to test it.

PR #2191, opened 2018-08-29
DS-3996 The "first few letters" search was fixed

santit96:
The problem was that a hidden parameter was added to the browse in order to fix another bug (DS-2675).
That parameter has the same name of the one used in the "first few letters" search input, so the form has the same field name twice which generates an inconsistency.

The whole issue is described here: https://jira.duraspace.org/browse/DS-3996


minurmin:
This fix (essentially identical to #2113 and related to #2464 as well)  seems to fix DS-3760 (=solr syntax errors when quotes and spaces are used within starts_with) - at least in most cases (=I found this while investigating the syntax error, although the starts with hidden field seems to generate other issues as well). So at least DS-3760, DS-3996, and DS-3945 are manifestations of the same issue.

kshepherd:
just temporarily closing and re-opening this PR to force Travis CI to rebuild (on the new service)

J4bbi:
Closing because this is a replication of #2113

PR #2193, opened 2018-09-03
[DS-3997] Set encoding of ORCID search path to UTF-8

qzhao-ulb:
Problem described in DS-3997. The search path for ORCID should use UTF-8 encoding.

https://jira.duraspace.org/browse/DS-3997

Fixes #7344


J4bbi:
Closing and opening again to restart tests

PR #2196, opened 2018-09-03
[DS-3988] InstallItemService should mint all kind of identifiers

pnbecker:
This patch was developed by [The Library Code GmbH](https://www.the-library-code.de) with the support of [Leibniz-Zentrum für Psychologische Information und Dokumentation](https://leibniz-psychology.org).

https://jira.duraspace.org/browse/DS-3988
Fixes #7335

### DIFF
--- a/dspace-api/src/main/java/org/dspace/content/InstallItemServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/content/InstallItemServiceImpl.java
@@ -60,11 +60,16 @@ public class InstallItemServiceImpl implements InstallItemService {
         Item item = is.getItem();
         Collection collection = is.getCollection();
         try {
-            if (suppliedHandle == null) {
-                identifierService.register(c, item);
-            } else {
+            // first register the handle if it was supplied
+            if (suppliedHandle != null) {
                 identifierService.register(c, item, suppliedHandle);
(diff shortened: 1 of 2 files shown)


pnbecker:
To test this you need a SAF containing a handle and a repository that is configured to register DOIs. Without the patch only the handle is registered but no DOI, with the path in place the handle specified in the SAF and a new DOI are registered.

In detail:

- You can create a SAF by exporting an item or a collection from an existing repository. Remember to either delete those before importing the SAF into the same repository or better to import the SAFs in a clean, empty new instance of DSpace. Otherwise the handles will be already in use and the import will fail.
- You can find information about the SAF ex- and import in the DSpace manual: https://wiki.duraspace.org/display/DSDOC6x/Importing+and+Exporting+Items+via+Simple+Archive+Format
- If you do not have repository that registers DOI it may be worth taking a look on how to get one. Read the article in the DSpace manual: https://wiki.duraspace.org/display/DSDOC6x/DOI+Digital+Object+Identifier. There is a test account for DuraSpace that works in the test system only. You can find the information about that here: https://wiki.duraspace.org/display/~pbecker/DOI+support+using+DataCite#DOIsupportusingDataCite-HowtoTest

PR #2197, opened 2018-09-04
DS-4000: REST API (v5.9) does not run

PhilipVis:
This is a pull request for issue https://jira.duraspace.org/browse/DS-4000.

This issue seems to be caused by a conflict in Jersey versions in the DSpace-rest module. 

The DSpace-api module has a dependency on Jersey 2.22.1: 
https://github.com/DSpace/DSpace/blob/dspace-5_x/dspace-api/pom.xml#L659

While the DSpace-rest module has a dependency on Jersey 1.19:
https://github.com/DSpace/DSpace/blob/dspace-5_x/dspace-rest/pom.xml#L39

Since the DSpace-rest module is dependant on the DSpace-api module, both Jersey versions end up in the DSpace-rest module. The solution I suggest here is to exclude the Jersey 2 dependency from the DSpace-rest module.

### DIFF
--- a/dspace-rest/pom.xml
+++ b/dspace-rest/pom.xml
@@ -103,6 +103,12 @@
         <dependency>
             <groupId>org.dspace</groupId>
             <artifactId>dspace-api</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.glassfish.jersey.core</groupId>
+                    <artifactId>jersey-client</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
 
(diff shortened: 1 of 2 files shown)


tdonohue (approved):
👍 Looks obvious to me, though I haven't tested it.

FYI, this issue would not affect `6.x` or `master` as the Jersey dependency version is now managed via a property in the Parent POM, e.g. https://github.com/DSpace/DSpace/blob/dspace-6_x/pom.xml#L29

terrywbrady (approved):
Built and tested on my DSpace 5 instance.  REST is running now.  Thanks @PhilipVis

PR #2203, opened 2018-09-14
DS-4009: Added threading to GoogleAnalyticsEventListener.

anis-moubarik:
Fixes #7356
Our findings showed that sometimes the response times from Google were slow, which meant the response time of DSpace was slow for the client. This creates a worker that does the GA request in its own thread and doesn't block DSpace from loading.

Related: [DS-3504](https://jira.duraspace.org/browse/DS-3504)


juancorr:
Y gg

pierbin:
Hi there,

I have this issue as well.
May I know it's working well? Thanks.

anis-moubarik:
> Hi there,
> 
> I have this issue as well.
> May I know it's working well? Thanks.

Hi, we have used this in production for over a year without problems. I also commited some recent changes.

PR #2213, opened 2018-09-25
DS-4019 Added a check for null when the group of a policy is obtained…

santit96:
… in getMostRecentModificationDate() and willChangeStatus() methods from XOAI.java class


santit96:
The issue is described  here: [DS-4019](https://jira.duraspace.org/browse/DS-4019)

terrywbrady (approved):
@santit96 ,
I just encountered this issue myself.  I was glad to find your fix. Thanks

PR #2214, opened 2018-09-25
[DS-4012] Support Multiple Dockerfiles in one branch

terrywbrady:
https://jira.duraspace.org/browse/DS-4012

See https://hub.docker.com/r/terrywbrady/dspace/builds/ for example build details and to locate published images for testing.

See https://wiki.duraspace.org/display/DSPACE/DSpace+and+Docker#DSpaceandDocker-Proposal:SupportMultipleJDKVersionsfor a description of the strategy to be implemented.

The dockerhub automated build seems to be more fragile once I began referencing Dockerfiles that exist outside of the root directory.  You will see many commits attempting to properly overlay the local.cfg file and other resources.

### DIFF
--- a/Dockerfile.jdk8
+++ b/Dockerfile.jdk8
@@ -2,7 +2,7 @@
 # See https://dspace-labs.github.io/DSpace-Docker-Images/ for usage details
 # 
 # This version is JDK8 compatible
-# - tomcat:8
+# - tomcat:8-jre8
 # - ANT 1.10.5
 # - maven:latest
 # - note: 
@@ -19,7 +19,7 @@ COPY dspace/src/main/docker/local.cfg /app/local.cfg
 RUN mvn package
 
 # Step 2 - Run Ant Deploy
-FROM tomcat:8 as ant_build
+FROM tomcat:8-jre8 as ant_build
 ARG TARGET_DIR=dspace-installer
 COPY --from=build /app /dspace-src
 WORKDIR /dspace-src/dspace/target/${TARGET_DIR}
@@ -36,7 +36,7 @@ RUN ant update_configs update_code update_webapps update_solr_indexes
 
 # Step 3 - Run tomcat
 # Create a new tomcat image that does not retain the the build directory contents
-FROM tomcat:8
+FROM tomcat:8-jre8
 COPY --from=ant_build /dspace /dspace
 EXPOSE 8080 8009
 
--- a/Dockerfile.jdk8
+++ b/Dockerfile.jdk8
@@ -1,5 +1,10 @@
 # This image will be published as dspace/dspace
 # See https://dspace-labs.github.io/DSpace-Docker-Images/ for usage details
+# 
+# This version is JDK8 compatible
+# - tomcat:8
+# - ANT 1.10.5
+# - maven:latest
 
 # Step 1 - Run Maven Build
 FROM maven as build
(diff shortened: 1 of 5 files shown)


terrywbrady:
My attempts to override the README.md file that Docker Hub places in the image description field have failed.  Therefore, I will add a Dockerfile section to the main DSpace README.

terrywbrady:
The following script exists to facilitate the checkout of the published images: https://wiki.duraspace.org/display/DSPACE/DSpace+and+Docker#DSpaceandDocker-ImageVerificationScript

Dockerfile.jdk8
```diff
@@ -1,5 +1,12 @@
 # This image will be published as dspace/dspace
 # See https://dspace-labs.github.io/DSpace-Docker-Images/ for usage details
+# 
+# This version is JDK8 compatible
+# - tomcat:8
```
tdonohue: As this is the `jdk8` Docker file, should this be using `tomcat:8-jre8`?  I'm not sure what JRE/JDK is in use by the default `tomcat:8` image.

tdonohue: As this is the `jdk8` Docker file, should this be using `tomcat:8-jre8`?  I'm not sure what JRE/JDK is in use by the default `tomcat:8` image.

Dockerfile.jdk8-test
```diff
@@ -0,0 +1,64 @@
+# This image will be published as dspace/dspace
+# See https://dspace-labs.github.io/DSpace-Docker-Images/ for usage details
+# 
+# This version is JDK8 compatible
+# - tomcat:8
```
tdonohue: As this is the `jdk8` Docker file, should this be using `tomcat:8-jre8`?  I'm not sure what JRE/JDK is in use by the default `tomcat:8` image.

dspace/src/main/docker/test/rest_web.xml
```diff
@@ -0,0 +1,114 @@
...
+<!--
+
+    The contents of this file are subject to the license and copyright
+    detailed in the LICENSE and NOTICE files at the root of the source
+    tree and available online at
+
+    http://www.dspace.org/license/
+
+-->
+<web-app xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
tdonohue: Could we add a comment to the top of this file that describes its purpose?  I'm just worried we'll forget why we are overriding this REST web.xml with a new one

dspace/src/main/docker/test/solr_web.xml
```diff
@@ -0,0 +1,205 @@
...
+
+    The contents of this file are subject to the license and copyright
+    detailed in the LICENSE and NOTICE files at the root of the source
+    tree and available online at
+
+    http://www.dspace.org/license/
+
+-->
+
+<web-app>
```
tdonohue: Same here, I'd recommend we add a comment to describe the purpose of overriding this file.

tdonohue (commented):
@terrywbrady : This looks good overall.  But, I did have a few inline comments/questions.

I glanced at the various Ports of this PR too, and I think they have similar questions (at least any that use Tomcat 8).  But, overall, I like the direction here. Just a few minor things to cleanup before merger.

Dockerfile.jdk8
terrywbrady: @tdonohue , I think that tomcat:8 and tomcat:8-jre8 are equivalent, but I like your suggestion to make that very explicit.

image

dspace/src/main/docker/test/solr_web.xml
terrywbrady: Commit added based on review comments.

pnbecker (approved):
:+1:
thank you!

tdonohue (approved):
👍 Looks good to me too, thanks @terrywbrady

PR #2217, opened 2018-09-28
[DS-4012] port to 5x for Jdk7 and 8

terrywbrady:
Port of https://github.com/DSpace/DSpace/pull/2214

### DIFF
--- a/Dockerfile.jdk7
+++ b/Dockerfile.jdk7
@@ -2,7 +2,7 @@
 # See https://dspace-labs.github.io/DSpace-Docker-Images/ for usage details
 # 
 # This version is JDK7 compatible
-# - tomcat:7.0.90-jre7
+# - tomcat:7-jre7
 # - ANT 1.9.13
 # - maven:latest
 # - note: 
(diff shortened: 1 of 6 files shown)


terrywbrady:
@tdonohue  and @terrywbrady discussed this over Slack on 10/1/2018 and we decided to revise the tomcat versions in use.

Tim Donohue [10:07 AM]
I'd say for 5.x we should use Tomcat 7 + JRE7 (those are the versions in install guide).  6.x should probably use Tomcat 8.  master should stick with Tomcat 8 for now (as I'm not sure we've tried it elsewhere) (edited)

To be clearer, based on this page: https://wiki.duraspace.org/display/DSPACE/DSpace+and+Docker.. I'd recommend:
• 7.x-jdk8 = Tomcat 8 + JRE8
• 6.x-jdk8 = Tomcat 8 + JRE8
• 5.x-jdk8 = Tomcat 7 + JRE8
• 5.x-jdk7 = Tomcat 7 + JRE7
• 4.x-jdk7 = Tomcat 7 + JRE7
So, I think the only one that needs "fixing" is 5.x-jdk8.... it should use Tomcat 7

Terry Brady [10:11 AM]
I see that our tomcat:8 is version 8.5.34.0 (the latest version 8
No change is needed there.

Tim Donohue [10:12 AM]
agreed. I think that's fine.  I think we just want to update 5.x-jdk8 to use a Tomcat 7 image that runs "jre8".  Everything else looks good to me

Terry Brady [10:13 AM]
Looking at the tags page, I see we can choose tomcat:7-jre7 and tomcat:7-jre8.  Those are probably better choices than hard-coding to a specific point version.

Tim Donohue [10:13 AM]
sounds reasonable to me

Terry Brady [10:14 AM]
I will add a note to the PR.  Thanks for thinking this through with me.

pnbecker (approved):
:+1:

tdonohue (approved):

PR #2222, opened 2018-10-02
DS-3453 OAI-PMH verb Identify doesn't work on Oracle

rruizibai:
Changes to detect if db type is oracle


rruizibai:
There is a mistake on imports, I close pull request

PR #2224, opened 2018-10-02
[DS-4032] RequestItem.isAccept_request() can NPE if setAccept_request() was never called.

mwoodiupui:
https://jira.duraspace.org/browse/DS-4032


mwoodiupui:
I have assumed that we don't really need a 3-state value (accepted, rejected, undecided) since the method returns a boolean primitive.

tdonohue (approved):
👍 I haven't tested this, but the change looks small/obvious.

kshepherd (approved):
+1 by inspection, boolean primitive is safer here and ensures null won't be returned

PR #2238, opened 2018-10-10
/api/config/submissiondefinitions/traditional/collections should return empty list rather than 204 [was#2189]

tantz001:
Here is an initial attempt following the initial simper suggestion as a solution.

### DIFF
--- a/dspace-spring-rest/src/main/java/org/dspace/app/rest/RestResourceController.java
+++ b/dspace-spring-rest/src/main/java/org/dspace/app/rest/RestResourceController.java
@@ -757,10 +757,12 @@ public class RestResourceController implements InitializingBean {
 
         } else {
             if (resource.getEmbeddedResources().get(rel) == null) {
-                response.setStatus(HttpServletResponse.SC_NO_CONTENT);
+	    PageImpl<Resource>emptyPage = new PageImpl(new ArrayList<Resource>(), page, 0);
+	    return assembler.toResource ( emptyPage );
             }
-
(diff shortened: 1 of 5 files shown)


abollini:
I played a bit more with this issue locally... we should only change the return code to 200 and provide an empty page for entpoint that should return a collection of resources not for single resource (like /bitstreams/:id/logo) that should continue to return 204 no content.

It seems to me that changing the code around line 783 in RestResourceController
```
             if (fullList == null || fullList.size() == 0) {
-                response.setStatus(HttpServletResponse.SC_NO_CONTENT);
-                return null;
+//                response.setStatus(HttpServletResponse.SC_NO_CONTENT);
+                PageImpl<RestAddressableModel> pageResult = new PageImpl(fullList, page,
+                        0);
+                result = assembler.toResource(pageResult);
+                return result;
             }

```

and in the DSpaceResource around line 129
```
                                         } else {
-                                            wrapObject = null;
+                                            PageImpl<RestAddressableModel> page = new PageImpl(linkedRMList);
+                                            wrapObject = new EmbeddedPage(linkToSubResource.getHref(), page,
+                                                    linkedRMList, name);
                                         }
```

we should eventually check that all the linkrepository return an empty page instead than `null`

tdonohue (approved):
👍 This code looks correct to me, based on all the discussions/feedback from @abollini .  Thanks again @tantz001 for sticking with this ticket!

I'll let @abollini give a final, quick review when he gets a chance, but I think this is ready to merge.

abollini (approved):
it looks good and ITs changes are in line with the expectation

PR #2240, opened 2018-10-12
[DS-3992] Added multiple custom exceptions and removed most direct uses of RuntimeException.

mspalti:


mspalti:
This is a work in progress.

The best practice for exceptions is to _refrain from using "RuntimeException" directly, and instead look to use Spring Data Rest Exceptions (like "DataAccessException") which extend RuntimeException.  If we find we need a new exception type, we should create it._

For the most part I have created new custom exceptions.

In most cases, how to apply a Spring NestedRuntimeException and specifically a DataAccessException is opaque (to me). So advice is welcome.

Here is how I have done the mapping.  In cases where an SQLException is thrown while attempting to retrieve items (e.g. findOne and findAll) I have thrown DataRetrievalFailureExceptions. https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/dao/DataRetrievalFailureException.html

Since the cause of other SQLExections is less clear, I created a custom exception that extends DataAccessException:  https://github.com/DSpace/DSpace/blob/214eefb8470cef3620d9f2836730a79d9abbe295/dspace-spring-rest/src/main/java/org/dspace/app/rest/exception/RESTSQLException.java#L17

Spring Data has a UncategorizedSQLException with a single 3-parameter constructor that didn't seem suited to our needs, but perhaps I am wrong and this is an alternative. https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/jdbc/UncategorizedSQLException.html

For other exceptions, I created custom exceptions that extend RuntimeException.

I made a lot of changes in exception catch blocks to throw the new exceptions. This too is a work in progress since we need to rebase with master after other PRs have been merged.

mspalti:
Still a work in progress.  The last commit includes the following changes:

- Implemented more custom exceptions.
- Where it makes sense, I am rethrowing the custom exceptions in DSpaceRestRepository to reduce the number of try/catch blocks that need to be maintained elsewhere in the code.
- Added only a few calls to Spring NextedRuntimeExceptions (so far only using DataRetrieveException as described earlier).
- Updated controller advice to handle the new custom exceptions.

mspalti:
Quick note. I see that https://jira.duraspace.org/browse/DS-3825 is a related issue claimed by @lap82. 

I saw there were unhandled exceptions in the submission workflow (and might have started work on a few of these). I'm happy to wait on this PR until that work is completed, then see what is left for me to do.

PR #2247, opened 2018-10-25
[DS-3433] Don't scare the admin. when an email Session is not found in JNDI

mwoodiupui:
https://jira.duraspace.org/browse/DS-3433
It's quite reasonable that the mail Session is not defined in the JNDI initial Context, or that there is no initial Context at all.  Don't issue WARN messages to the log in these cases as they are not failures -- it is expected that a new Session will be created from the DSpace configuration.

Other name service exceptions are still logged, as they may indicate real problems.

Also, the name being looked up is now logged at DEBUG level.

Briefly tested on the command line without a Context; with a Context but no name defined; with a Context and a name defined.  Currently the logger is not set up by webapp.s until after the Session has been retrieved or created, so it's not possible to test in a webapp.

See https://github.com/mwoodiupui/ephemeral-jndi-context for how I rigged up the Context while testing.

### DIFF
--- a/dspace-services/src/main/java/org/dspace/services/email/EmailServiceImpl.java
+++ b/dspace-services/src/main/java/org/dspace/services/email/EmailServiceImpl.java
@@ -12,7 +12,9 @@ import javax.mail.Authenticator;
 import javax.mail.PasswordAuthentication;
 import javax.mail.Session;
 import javax.naming.InitialContext;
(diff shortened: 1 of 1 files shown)


tdonohue (approved):
👍 Code looks good / obvious to me.

terrywbrady (approved):

PR #2250, opened 2018-10-29
DS-4067 Allow custom arg to ImageMagick

cwilper:
https://jira.duraspace.org/browse/DS-4067
Fixes #7414


terrywbrady (approved):
Code looks good

PR #2251, opened 2018-10-29
DS-4068 Configure ability to send Google Analytics as either "event" or "pageview"

aprilherron:
Currently, DSpace reports bitstream downloads to Google Analytics as "events". However, users may prefer to see these recorded as "pageviews" in the Google Analytics dashboard.

This PR allows the user to configure if GA should report as "event" or "pageview"

This PR is based on [PR-2248](https://github.com/DSpace/DSpace/pull/2248)

Fixes #7415

### DIFF
--- a/dspace-api/src/main/java/org/dspace/google/GoogleAnalyticsEvent.java
+++ b/dspace-api/src/main/java/org/dspace/google/GoogleAnalyticsEvent.java
@@ -1,0 +1,94 @@
+/**
+ * The contents of this file are subject to the license and copyright
+ * detailed in the LICENSE and NOTICE files at the root of the source
+ * tree and available online at
+ * <p>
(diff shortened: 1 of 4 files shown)


bram-atmire:
Closing — this PR was written against the legacy Universal Analytics backend code. Universal Analytics itself was [sunset by Google in July 2024](https://support.google.com/analytics/answer/11583528), so the underlying ability to configure UA reporting as "event" vs "pageview" no longer has a target to send to.

Current `main` has migrated to GA4, which uses a fundamentally different event-only model (via `GoogleAnalytics4ClientRequestBuilder`) — the event/pageview distinction doesn't exist there in the same way. Closing as the feature is no longer meaningful under the GA4 architecture. Thanks @aprilherron for the original contribution.

PR #2252, opened 2018-10-29
DS-4069 Use Google Analytics cookie for more consistent session tracking

aprilherron:
Currently, Google Analytics looks for the "Client ID" in the request's session. If none exists, it simply creates a new random UUID. This can result to poor tracking of a user's session. To improve on this, we can check the "_ga" cookie for the session ID. This greatly improves user session tracking.

This PR is based on [PR-2248](https://github.com/DSpace/DSpace/pull/2248)

Fixes #7416

### DIFF
--- a/dspace-api/src/main/java/org/dspace/google/GoogleAnalyticsEvent.java
+++ b/dspace-api/src/main/java/org/dspace/google/GoogleAnalyticsEvent.java
@@ -2,7 +2,7 @@
  * The contents of this file are subject to the license and copyright
  * detailed in the LICENSE and NOTICE files at the root of the source
  * tree and available online at
- * <p>
+ *
  * http://www.dspace.org/license/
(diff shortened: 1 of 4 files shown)


bram-atmire:
Closing — this PR was written against the legacy Universal Analytics backend code. Universal Analytics itself was [sunset by Google in July 2024](https://support.google.com/analytics/answer/11583528), and the `_ga` Client ID format / session handling it relied on are specific to UA.

Current `main` has migrated to GA4 (via `GoogleAnalytics4ClientRequestBuilder`) and a split architecture where page-view tracking happens in the Angular UI; the UA-style server-side session reconstruction via `_ga` cookie is no longer the relevant layer. Closing as the feature doesn't translate to the GA4 model. Thanks @aprilherron for the original contribution.

PR #2253, opened 2018-11-01
DS-4050 Search method returns 204 NoContent or 200 for empty page if no result.

tantz001:


benbosman (approved):
Looks good to me. 204 is more sensible than 404 for when the query didn't contain any results

abollini (approved):
Code looks good

PR #2254, opened 2018-11-01
fix slow discovery

kardeiz:
Fixes issue reported at https://groups.google.com/forum/#!topic/dspace-tech/PbKbDfQlhok.

`DSpaceValidity`, which is used to cache search results in `AbstractSearch`, looks up all bundles, and all bitstreams of those bundles, for each item in a search results page. When an item has lots of bitstreams, this can lead to thousands upon thousands of SQL queries per page load.

This PR only looks at ORIGINAL bundles when determining the cache validity key.


AlexanderS:
There may be some more locations for this fix, f.e the list of recent submissions on a collection/community home page.

PR #2257, opened 2018-11-03
[DS-4062] Endpoint to allow logged in EPerson to change password or other profile information.

mspalti:
This PR adds functionality for updates to the eperson password, name, phone and language by authenticated users. Includes: 

- New authn controller patch endpoint.
- New eperson repository patch method for authenticated users.
- New eperson patch operations.
- Integration tests.

### DIFF
--- a/dspace-spring-rest/src/main/java/org/dspace/app/rest/AuthenticationRestController.java
+++ b/dspace-spring-rest/src/main/java/org/dspace/app/rest/AuthenticationRestController.java
@@ -117,7 +117,6 @@ public class AuthenticationRestController implements InitializingBean {
 
         context = ContextUtil.obtainContext(request);
 
-
         if (context == null || context.getCurrentUser() == null) {
             // Note that the actual HTTP status in this case is set by
             // org.dspace.app.rest.security.StatelessLoginFilter.unsuccessfulAuthentication()
@@ -128,4 +127,5 @@ public class AuthenticationRestController implements InitializingBean {
             return ResponseEntity.ok().build();
         }
     }
+
 }
--- a/dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/EPersonRestRepository.java
+++ b/dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/EPersonRestRepository.java
@@ -186,7 +186,7 @@ public class EPersonRestRepository extends DSpaceObjectRestRepository<EPerson, E
     }
 
     @Override
-    @PreAuthorize("hasAuthority('ADMIN')")
+    @PreAuthorize("hasPermission(#uuid, 'EPERSON', #patch)")
     protected void patch(Context context, HttpServletRequest request, String apiCategory, String model, UUID uuid,
                          Patch patch) throws AuthorizeException, SQLException {
         patchDSpaceObject(apiCategory, model, uuid, patch);
@@ -206,6 +206,9 @@ public class EPersonRestRepository extends DSpaceObjectRestRepository<EPerson, E
         }
         if (ePersonRest.isCanLogIn() != ePerson.canLogIn()) {
             ePerson.setCanLogIn(ePersonRest.isCanLogIn());
+        }
+        if (!Objects.equals(ePersonRest.getEmail(), ePerson.getEmail())) {
+            ePerson.setEmail(ePersonRest.getEmail());
         }
         if (!Objects.equals(ePersonRest.getNetid(), ePerson.getNetid())) {
             ePerson.setNetid(ePersonRest.getNetid());
--- a/dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/patch/ItemPatch.java
+++ b/dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/patch/ItemPatch.java
@@ -18,6 +18,8 @@ import org.springframework.stereotype.Component;
 
 /**
  * Provides PATCH operations for item updates.
+ *
+ * @author Michael Spalti
  */
 @Component
 public class ItemPatch extends DSpaceObjectPatch<ItemRest> {
@@ -26,7 +28,7 @@ public class ItemPatch extends DSpaceObjectPatch<ItemRest> {
     ItemOperationFactory patchFactory;
 
     /**
-     * Peforms the replace operation.
+     * Performs the replace operation.
      * @param item the rest representation of the item
      * @param operation the replace operation
      * @throws UnprocessableEntityException
--- a/dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/patch/factories/impl/PatchOperation.java
+++ b/dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/patch/factories/impl/PatchOperation.java
@@ -10,7 +10,6 @@ package org.dspace.app.rest.repository.patch.factories.impl;
 import org.apache.commons.lang3.BooleanUtils;
 import org.dspace.app.rest.exception.PatchBadRequestException;
 import org.dspace.app.rest.model.RestModel;
-import org.dspace.app.rest.model.patch.LateObjectEvaluator;
 import org.dspace.app.rest.model.patch.Operation;
 
 /**
--- a/dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/patch/factories/impl/PatchOperation.java
+++ b/dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/patch/factories/impl/PatchOperation.java
@@ -7,9 +7,6 @@
  */
 package org.dspace.app.rest.repository.patch.factories.impl;
 
-import java.util.ArrayList;
-import java.util.List;
-
 import org.apache.commons.lang3.BooleanUtils;
 import org.dspace.app.rest.exception.PatchBadRequestException;
 import org.dspace.app.rest.model.RestModel;
@@ -58,42 +55,12 @@ public abstract class PatchOperation<R extends RestModel, T>
         if (value instanceof String) {
             bool = BooleanUtils.toBooleanObject((String) value);
             if (bool == null) {
-                // make sure the string was converted to boolean.
                 throw new PatchBadRequestException("Boolean value not provided.");
             }
         } else {
             bool = (Boolean) value;
         }
         return bool;
-    }
-
-    // This is duplicated code (see org.dspace.app.rest.submit.factory.impl.PatchOperation)
-    // If it stays here, it should be DRY. Current patch resource patch operations do not
-    // use these methods since operation values are either strings or booleans.
-    // These methods handle JsonValueEvaluator instances for json objects and arrays,
-    // as returned by the JsonPatchConverter. A complete implementation of the PatchOperation
-    // class will need these methods.
-    public List<T> evaluateArrayObject(LateObjectEvaluator value) {
-        List<T> results = new ArrayList<T>();
-        T[] list = null;
-        if (value != null) {
-            LateObjectEvaluator object = (LateObjectEvaluator) value;
-            list = (T[]) object.evaluate(getArrayClassForEvaluation());
-        }
-
-        for (T t : list) {
-            results.add(t);
-        }
-        return results;
-    }
-
-    public T evaluateSingleObject(LateObjectEvaluator value) {
-        T single = null;
-        if (value != null) {
-            LateObjectEvaluator object = (LateObjectEvaluator) value;
-            single = (T) object.evaluate(getClassForEvaluation());
-        }
-        return single;
     }
 
     /**
--- a/dspace-spring-rest/src/main/java/org/dspace/app/rest/security/EPersonRestPermissionEvaluatorPlugin.java
+++ b/dspace-spring-rest/src/main/java/org/dspace/app/rest/security/EPersonRestPermissionEvaluatorPlugin.java
@@ -9,9 +9,14 @@ package org.dspace.app.rest.security;
 
 import java.io.Serializable;
 import java.sql.SQLException;
+import java.util.List;
 import java.util.UUID;
 
+import org.dspace.app.rest.model.patch.Operation;
+import org.dspace.app.rest.model.patch.Patch;
+import org.dspace.app.rest.repository.patch.factories.EPersonOperationFactory;
 import org.dspace.app.rest.utils.ContextUtil;
+import org.dspace.authorize.service.AuthorizeService;
 import org.dspace.core.Constants;
 import org.dspace.core.Context;
 import org.dspace.eperson.EPerson;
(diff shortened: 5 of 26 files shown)


mspalti:
Note that this PR includes the Spring LateObjectEvaluator.  This will need to be updated. See:

https://jira.duraspace.org/browse/DS-3802?focusedCommentId=61380

dspace-spring-rest/src/main/java/org/dspace/app/rest/AuthenticationRestController.java
```diff
@@ -111,6 +133,39 @@ public ResponseEntity logout() {
         return ResponseEntity.noContent().build();
     }
 
+    @RequestMapping(value = "/patch", method = {RequestMethod.PATCH})
```
abollini: we don't need to offer a separate endpoint for patching done by the current user on his profile. It should be managed transparently by the REST implementation on the current endpoint

abollini (dismissed):
Hi @mspalti thanks for your PR, sorry to be unable to provide a review earlier.
You should take a look to the documentation provided by @tomdesair in his original PR where the authorization framework was introduced https://github.com/DSpace/DSpace/pull/1976#issuecomment-406543859
I expect that the functionality can be implemeted without major change to the existent EPersonRestRepository code except for the introduction of a custom "permission evaluator" maybe starting from https://github.com/DSpace/DSpace/blob/66b89a786d7f55318ea665ca6ebd5699d4daf1fc/dspace-spring-rest/src/main/java/org/dspace/app/rest/security/EPersonRestPermissionEvaluatorPlugin.java

dspace-spring-rest/src/main/java/org/dspace/app/rest/AuthenticationRestController.java
mspalti: Hi @abollini 
Sorry for the delay in getting back to this. I did have some doubts about the new endpoint and happy to take a look at the PR provided by @tomdesair. I believe I understand what you are suggesting here, so shouldn't take long to update this PR.

mspalti:
@abollini, I finally got back to this PR.  I believe that using the current eperson PreAuthorize plugin is sufficient to allow both administrators and the currently logged in user to patch the record.  However, it seems to me that some patch operations should be restricted to administrators.  I added a check in the repository method to do this.  That check can be removed or expanded if I'm mistaken.

mspalti:
Removed the metadata operations since all DSO metadata patches are now handled with a separate endpoint.

mspalti:
Hi @abollini, in the rest contract PR you mentioned the need for an email patch operation (I assume for administrators only). I added new code for an email patch to this PR.  Perhaps it would have been better to submit it in a different PR, and that can still be done. But I don't think there is a separate Jira issue yet.

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/EPersonRestRepository.java
```diff
@@ -201,6 +217,29 @@ public void patch(Context context, HttpServletRequest request, String apiCategor
...
+            for (Operation op: operations) {
+                if (op.getPath().contentEquals("/canLogin")
+                        || op.getPath().contentEquals("/certificate")
+                        || op.getPath().contentEquals("/netid")
+                        || op.getPath().contentEquals("/email")) {
+                    if (!isAdmin) {
+                        throw new AuthorizeException(op.getPath() + " requires administrator.");
+                    }
+                }
+            }
```
tomdesair: Hi @mspalti,

Ideally this code should move to a Permission evaluator so that this method can [just focus on doing the patch operation](https://en.wikipedia.org/wiki/Single_responsibility_principle). In addition, we might encounter similar scenario's for other DSpace objects.

This might seem like a challenge, but I think it's doable:

1. Change the `hasPermission(#uuid, 'EPERSON', 'WRITE')` at the top of this method to `hasPermission(#uuid, 'EPERSON', #patch)`.

1. In the `DSpaceObjectRestPermissionEvaluator` super class, implement the `boolean hasPermission(Authentication authentication, Serializable targetId, String targetType, Object permission)` method as follows: Check if `permission instanceof Patch`. If that is the case, call a new method `hasPatchPermission(authentication, targetId, targetType, (Patch) permission)`. Otherwise convert the permission to a `DSpaceRestPermission` (`DSpaceRestPermission restPermission = DSpaceRestPermission.convert(permission);`) and call a new abstract method `hasDSpacePermission(authentication, targetId, targetType, restPermission)`.

1. Add a default implementation for `hasPatchPermission` that just does `return hasPermission(authentication, targetId, targetType, "WRITE");`

1. Implement the `hasDSpacePermission` in all subclasses by just renaming the currently present `hasPermission` method and changing the type of the last parameter.

1. In the `EPersonRestPermissionEvaluatorPlugin`, override the `hasPatchPermission` method. In that implementation, first check that `hasPermission(authentication, targetId, targetType, "WRITE")` is true. If that is the case, then execute the custom admin checks you wrote here.

Then you should be able to remove these admin checks from the `EPersonRestRepository` class.

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/EPersonRestRepository.java
```diff
@@ -201,6 +217,29 @@ public void patch(Context context, HttpServletRequest request, String apiCategor
...
+            try {
+                isAdmin = authorizeService.isAdmin(context);
+            } catch (SQLException e) {
+                throw new RuntimeException(e.getMessage(), e);
+            }
+            for (Operation op: operations) {
+                if (op.getPath().contentEquals("/canLogin")
+                        || op.getPath().contentEquals("/certificate")
+                        || op.getPath().contentEquals("/netid")
+                        || op.getPath().contentEquals("/email")) {
```
tomdesair: Shouldn't it be possible for a user to change his e-mail address?

tomdesair (commented):
I added some ideas on how to extend the `EPersonRestPermissionEvaluatorPlugin`. Let me know what you think.

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/EPersonRestRepository.java
abollini: this will be a new feature compared to what we have in the current dspace.
There are also some additional complexity related to the change of the email such as
- the risk that a spammer will register in the system using a valid email and after will change it email (of course we will have log to track that)
- the risk to put a wrong email that will not go through the same validation procedure

I recommend to do not include such feature in this PR but eventually open a JIRA ticket to discuss it

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/EPersonRestRepository.java
abollini: I agree with @tomdesair proposal. If it looks to complex I'm also fine with log it as a necessary improvement and deal with it in a followup PR

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/patch/ItemPatch.java
```diff
@@ -18,6 +18,8 @@
 
 /**
  * Provides PATCH operations for item updates.
+ *
+ * @author Michael Spalti
```
abollini: I don't think this is an intentional change

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/patch/factories/impl/PatchOperation.java
```diff
@@ -68,33 +64,35 @@ Boolean getBooleanOperationValue(Object value) {
...
-            list = (T[]) object.evaluate(getArrayClassForEvaluation());
-        }
+    // as returned by the JsonPatchConverter.
 
-        for (T t : list) {
-            results.add(t);
-        }
-        return results;
-    }
+    // UPDATE: Because metadata patches for all DSOs is now handled separately,
```
abollini: we should remove the commented out code

dspace-spring-rest/src/test/java/org/dspace/app/rest/EPersonRestRepositoryIT.java
```diff
@@ -48,6 +48,8 @@
 
 public class EPersonRestRepositoryIT extends AbstractControllerIntegrationTest {
 
+    String defaultPassword = "7Testpass";
```
abollini: there is already a password attribute in the https://github.com/DSpace/DSpace/blob/master/dspace-spring-rest/src/test/java/org/dspace/app/rest/test/AbstractIntegrationTestWithDatabase.java#L62

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/EPersonRestRepository.java
mspalti: Since this will be a new feature that probably needs more analysis, I agree with the idea of a new JIRA ticket.

I also agree that the user should be able to change their email address.

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/EPersonRestRepository.java
mspalti: I will work on @tomdesair suggestion.

mspalti: Hi @tomdesair 

I finally got to work on your proposal and it appears to be successful. The check for patch permission on the eperson is now handled by the `EPersonRestPermissionEvaluatorPlugin`. All tests are passing...

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/patch/ItemPatch.java
mspalti: Hi @abollini ,
This was intentional (I had left off the author).  BTW, I don't need the attribution, but I assume the author tag is recommended as a way of providing work history.  Let me know if that's not the case and I'll make less use of it. :-)

dspace-spring-rest/src/main/java/org/dspace/app/rest/security/DSpaceObjectPermissionEvaluatorPlugin.java
```diff
@@ -34,5 +37,42 @@ public boolean hasPermission(Authentication authentication, Object targetDomainO
         return hasPermission(authentication, dSpaceObject.getId(), dSpaceObject.getType(), permission);
     }
 
+    public boolean hasPermission(Authentication authentication, Serializable targetId, String targetType,
```
tomdesair: Please add the `@Override` annotation

dspace-spring-rest/src/main/java/org/dspace/app/rest/security/EPersonRestPermissionEvaluatorPlugin.java
```diff
@@ -70,4 +85,31 @@ public boolean hasPermission(Authentication authentication, Serializable targetI
...
+        }
+
+        List<Operation> operations = patch.getOperations();
+
+        /**
+         * Authenticated users are allowed to update their
+         * own password. Other operations are restricted to administrators.
+         */
+        for (Operation op: operations) {
+            if (op.getPath().contentEquals(EPersonOperationFactory.OPERATION_PASSWORD_CHANGE)) {
```
tomdesair: We should be careful here. This says "if any of the operations is PASSWORD_CHANGED, allow the complete patch operation".

What it should be I think: "if any of the operations is NOT PASSWORD_CHANGED, then refuse the complete patch operation".

dspace-spring-rest/src/main/java/org/dspace/app/rest/security/EPersonRestPermissionEvaluatorPlugin.java
```diff
@@ -70,4 +85,31 @@ public boolean hasPermission(Authentication authentication, Serializable targetI
...
+                return true;
+            }
+        }
+        return false;
```
tomdesair: In that case, then this should be `return true;`

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/EPersonRestRepository.java
```diff
@@ -201,6 +217,29 @@ public void patch(Context context, HttpServletRequest request, String apiCategor
...
+//            for (Operation op: operations) {
+//                if (op.getPath().contentEquals("/canLogin")
+//                        || op.getPath().contentEquals("/certificate")
+//                        || op.getPath().contentEquals("/netid")
+//                        || op.getPath().contentEquals("/email")) {
+//                    if (!isAdmin) {
+//                        throw new AuthorizeException(op.getPath() + " requires administrator.");
+//                    }
+//                }
+//            }
```
tomdesair: Please remove the commented out code

tomdesair (commented):
Thanks a lot for the `hasPatchPermission` implementation @mspalti !

I only have a few remarks left.

dspace-spring-rest/src/main/java/org/dspace/app/rest/security/EPersonRestPermissionEvaluatorPlugin.java
mspalti: Yes, I agree. Thanks!  I will have time to update the PR tomorrow.

mspalti:
I updated the permissions plugin as suggested. Since there seems to be no satisfying way to test this part via integration tests (401 is always returned) I added a unit test.

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/patch/ItemPatch.java
mspalti: and...now that I've rebased I see that @cwilper has been working here as well. I'm still new to the process so please advise on what is most useful to everyone vis-a-vis author annotations!

mwoodiupui: https://wiki.duraspace.org/display/DSPACE/Code+Style+Guide#CodeStyleGuide-DSpaceJavaStyleGuide(Adopted2018,forDSpace7.xandabove) may be useful.

I don't think there is a guideline for this, but personally I tend to add @author only when I was involved in creating the file -- probably because my IDE does it automagically and I have no reason to take it out.  I'm too lazy to add one manually to existing code that I'm updating, unless I've been asked to.

``git`` keeps track of our contributions whether we tag the code or not.

mspalti: Thanks @mwoodiupui that's helpful to know.  I think the path of least resistance makes sense here!

tomdesair:
Thanks @mspalti, the code looks good! Using a unit test is not a problem in these cases.

I have one final request, could you add (or extend) an integration in which one non-admin eperson tries to update the password of another non-admin eperson. I think this is an important security use case that we should automatically validate in a test.

mspalti:
@tomdesair, I added the integration test you requested.

tomdesair (approved):
Perfect! Thanks a lot @mspalti !

@abollini, if you don't have any remaining remarks, then I think this PR can be merged.

mspalti:
Hi @abollini  I resolved the merge conflicts on this PR.

tdonohue (approved):
👍 This looks good to me. I gave this a code review today.

I'm going to merge this as it's now at +2, and I've verified that all of @abollini's prior requests have been met.  If further updates are necessary, they can come in a future PR.  Thanks @mspalti!

PR #2273, opened 2018-11-26
Item: Remove duplicated getBundles(name) method

AlexanderS:
The `getBundles(String name)` method is already provided by the [ItemService](https://github.com/DSpace/DSpace/blob/7f4b4e9/dspace-api/src/main/java/org/dspace/content/ItemServiceImpl.java#L291-L301). This was introduced in #1910 to fix DS-3310. I think it was simply missed, that this method already exists in the ItemService.

### DIFF
--- a/dspace-api/src/main/java/org/dspace/content/Item.java
+++ b/dspace-api/src/main/java/org/dspace/content/Item.java
@@ -291,26 +291,6 @@ public class Item extends DSpaceObject implements DSpaceObjectLegacySupport {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


tdonohue (approved):
👍 Looks correct to me. This (identical) method obviously need not exist in two places. Thanks for the cleanup, @AlexanderS!

alanorth:
Thanks @AlexanderS. I rebased this on top of latest `main` and force pushed so the tests would run. It looks like a simple and obvious fix to me.

PR #2279, opened 2018-12-04
[DS-4094] DiscoveryRestController and DiscoveryRestRepository log as ScopeResolver

mwoodiupui:
https://jira.duraspace.org/browse/DS-4094
While fixing this, I also noticed that we were logging the DSO type as the query in several places.
After patching, all dspace-spring-rest unit and integration tests run without complaint.


tdonohue (approved):
👍 Obviously correct. Haven't tested, but these are just logging fixes, and they look good to me.

terrywbrady (approved):
code looks good

PR #2289, opened 2018-12-12
DS-4119: Enable ENTER key in textareas and add suitable rendering of running text (e.g. abstracts, etc.)

### LINKED ISSUE #7466: [DS-4119] Enable ENTER key in textareas and suitable rendering of running text
<i>Imported from JIRA <a href="https://jira.lyrasis.org/browse/DS-4119">[DS-4119]</a> created by r3r57</i>
<h2><a name="Problem"></a>Problem</h2>

<p>Currently, DSpace prevents users from inserting line breaks within <em>textareas</em> by pressing the ENTER key .</p>

<p>This is due to <a href="https://github.com/DSpace/DSpace/blob/9c16711dbef079e9eee55b660997b52adf487354/dspace-jspui/src/main/webapp/utils.js#L143," class="external-link" target="_blank" rel="nofollow noopener">https://github.com/DSpace/DSpace/blob/9c16711dbef079e9eee55b660997b52adf487354/dspace-jspui/src/main/webapp/utils.js#L143,</a> which disables the ENTER key in forms in general.</p>

<p>But this  simple method to prevent "accidentally submitting a form", hinders the user to write a well structured description or abstract when submitting a publication. Indeed, the user can copy and paste an abstract in the corresponding form field, but when it comes to manual adjustments, there is no convenient way to structure the text with 'headings' and paragraphs.</p>

<p>Furthermore, when reviewing the submission in the "Verify" step, even previously structured text is displayed as continuous text without any structure. The same display error occurs when viewing a publication. The issue here is, that DSpace escapes the input to prevent XSS, but the HTML encoded line feed (&#x0A) is not rendered as new line in the front end.</p>
<h2><a name="Solution"></a>Solution</h2>

<p>The solution might be quite simple. DSpace must only disable the ENTER key in input fields other than textareas, since the default event will insert a line break in textareas and never submit the form by accident.</p>

<p>To enable suitable rendering of running text in the "Verify" step and in a publication's details view, replacing the encoded line feed with it's HTML equivalent (<br/>) solves the problem.</p>

<p>I've already created a pull request on Github (<a href="https://github.com/DSpace/DSpace/pull/2289)," class="external-link" target="_blank" rel="nofollow noopener">https://github.com/DSpace/DSpace/pull/2289),</a> which addresses the mentioned problem.</p>

cmtjk:
Fixes #7466

## Rationale (for DSpace 6_x, JSPUI)
This pull requests addresses DSpace's issue with `textareas`, not allowing the user to insert a line break in a suitable way.
By default DSpace disables the 'ENTER' key in general to prevent "accidentally submitting a form when the 'Enter' key is pressed".

Running text, e.g. abstracts or descriptions, can only be inserted as continuous text, without sensible structure through paragraphs.

## Insert line breaks in `textarea`
This pull request restricts the 'disabled ENTER key' functionality to `inputs` other than `textarea`.

textarea_form

## Render running text in a submission's "Verify" step and when displaying an item in a suitable way
This pull request also addresses an issue with running text being rendered as one continuous block of text in the submission's "Verify" step, because it turned out, that the user may be confused by this representation.

screenshot from 2018-12-10 12 47 08

*I'll create a corresponding issue on https://jira.duraspace.org/projects/DS/issues/ as soon as I get my account.*

### DIFF
--- a/dspace-jspui/src/main/java/org/dspace/app/webui/jsptag/ItemTag.java
+++ b/dspace-jspui/src/main/java/org/dspace/app/webui/jsptag/ItemTag.java
@@ -672,7 +672,7 @@ public class ItemTag extends TagSupport
 	                    }
                         else
                         {
-                            out.print(Utils.addEntities(val.getValue()));
+                            out.print(Utils.addEntities(val.getValue().replace("&#x0A;", "<br/>")));
                         }
                     }
                 }
--- a/dspace-jspui/src/main/webapp/submit/review-metadata.jsp
+++ b/dspace-jspui/src/main/webapp/submit/review-metadata.jsp
@@ -194,7 +194,7 @@
                     }
                     else
                     {
-                        row.append(Utils.addEntities(values.get(i).getValue()));
+                        row.append(Utils.addEntities(values.get(i).getValue()).replace("&#x0A;", "<br/>"));
                     }
                     if (isAuthorityControlled)
                     {
--- a/dspace-jspui/src/main/webapp/utils.js
+++ b/dspace-jspui/src/main/webapp/utils.js
@@ -149,7 +149,7 @@ function disableEnterKey(e)
      else
           key = e.which;     //Firefox & Netscape
 
-     if(key == 13)  //if "Enter" pressed, then disable!
+     if(key == 13 && event.target.nodeName!='TEXTAREA')
           return false;
      else
           return true;


r3r57:
Please note that I reviewed my PR and included an much easier solution from 4science to enable the enter key for textareas.

PR #2294, opened 2018-12-13
DS-4104: Avoid crosswalk of incorrect publication dates for google scholar

bram-atmire:
Fixes #7451


tdonohue (approved):
👍 Looks correct to me. I agree that this should only be set to `dc.date.issued` by default.

kshepherd (approved):
approving as with 5.x version
merging
needs porting to master

PR #2295, opened 2018-12-13
DS-4104 Avoid crosswalking wrong dates for GS

bram-atmire:
Fixes #7451

6.x version: https://github.com/DSpace/DSpace/pull/2294


tdonohue (approved):
👍 Looks correct to me. I agree that this should only be set to `dc.date.issued` by default.

kshepherd (approved):
+1 looks good to me, merging now
thanks @bram-atmire !

PR #2298, opened 2018-12-13
[DS-4113] Configure clean plugin in parent pom to also clean sub-modules

kshepherd:
https://jira.duraspace.org/browse/DS-4113

This change will get the 'clean' maven task to also clean target directories from sub-modules. Currently, they never get touched by a 'clean' from the parent project.

It does refer to 'dspace/modules' as the path for submodules so if that ever changes, this plugin configuration will need to be updated (but so will a lot of pom configuration).

I'd like to port this to 6.x as well if others agree.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -233,6 +233,22 @@
                         </execution>
                     </executions>
                 </plugin>
+                <!-- Used to clean all 'target' directories from parent project -->
+                <!-- This additional configuration also cleans sub-modules -->
+                <plugin>
+                    <artifactId>maven-clean-plugin</artifactId>
+                    <version>2.5</version>
+                    <configuration>
+                        <filesets>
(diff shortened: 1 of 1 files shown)


hardyoyo (approved):

terrywbrady (approved):
Code looks good, have not tested this.  I have wished this functionality was in place.

toniprieto:
Hello @kshepherd,  it has been applied to master branch but it hasn't been ported to 6.x. I'm using it on 6.x and it is really useful. Could it be cherry-picked?

Let me know if I can help submitting the 6.x port. Thanks!

kshepherd:
@toniprieto thanks! and tanks for the reminder about porting - i thought i'd done it already
if you have something ready to go feel free to send it up!

PR #2303, opened 2018-12-17
CST-420 Added ecommerce.tag.id to build.properties

fcadili:
A new property ecommerce.tag.id was introduced to localize the identifier of Retired tag used in wwocommerce plugin to remove a published item.


fcadili:
No action was performed. Closed issue.

PR #2307, opened 2018-12-21
[DS-4126] Optimize docker builds

terrywbrady:
## Jira Issues

- https://jira.duraspace.org/browse/DS-4126
- https://jira.duraspace.org/browse/DS-4146

## Purpose of this PR

The goal of this PR is to make the DSpace docker build process more efficient.
- Make the build faster
- Reduce the size of the final docker image
- Reduce the size of the interim docker images created by the build

### Summary of Changes

- A new docker image "dspace/dspace-dependencies" will be created for each branch of DSpace.  This image will contain a download of the baseline maven dependencies needed for that branch of DSpace.
- Subsequent dockerfile builds will only need to download maven dependencies that are not already present in the dspace-dependencies image.
- The DSpace dockerfile build runs in 3 steps (1)maven (2) ant (3) copy to tomcat.  By running a multi-stage docker build, the final image is smaller in size.
- This pull request will delete temporary build artifacts after each step of the build process.  This change significantly decreases the size of the interim images generated by the docker build process.
- Add "ant init_installation" to ensure all working directories are created in the container
- Remove the ant download from the final image since images should be rebuilt

## Testing Results based on this PR

Baseline Stats
- Workstation build time: 12 min
- Dockerhub build time: ~40-50 min
- Intermediate images left on build machine: 10G
  - These intermediate images are designed to facilitate a build re-run.
  - The docker command offers no option to clear these.  The user must explicitly delete them with "docker image prune"

Purpose
- Create a base image dspace/dspace-dependencies containing a populated maven cache
- Minimize the residual content recorded to intermediate images on Docker 

Based on some experimentation, I believe that we can
- Reduce the size of the intermediate images by 50%
- Reduce build time by 60%

Testing results
- Workstation build time: 4:42 min (-60%)
- Dockerhub build time: 29 min (-25%-40%)
- Intermediate images left on build machine: 5G (-50%)

## Future Goals for DSpace and Docker

This pull request is part of a larger set of goals for DSpace and Docker

- Make DSpace Docker Images the recommended approach for developing and testing changes to DSpace.  By creating a default development and testing environment for DSpace, the process of on-boarding and supporting new developers could be simplified.
- Publish docker images for each pull request submitted to DSpace.  Because this would eliminate the need to run a local build for a given PR, the review cycle for PR's should be improved.
- Document instructions for running published DSpace docker images on each of the major cloud providers.


.dockerignore
```diff
@@ -3,3 +3,4 @@
 .settings/
 */target/
 dspace/modules/*/target/
+Dockerfile.*
```
PTrottier:
```suggestion
Dockerfile.*

```
is there a preference to add a new line at the end of a file?

Dockerfile.dependencies
```diff
@@ -0,0 +1,25 @@
...
+ADD --chown=dspace . /app/
+COPY dspace/src/main/docker/local.cfg /app/local.cfg
+
+# Trigger the installation of all maven dependencies including the Mirage2 dependencies
+# Clean up the built artifacts in the same step to keep the docker image small
+RUN mvn package -Dmirage2.on=true && mvn clean
+
+# clear the contents of the /app directory so no artifacts are left when dspace:dspace is built
+USER root
+RUN rm -rf /app/*
```
PTrottier:
```suggestion
RUN rm -rf /app/*

```

See previous

PTrottier (commented):
@terrywbrady I quickly glanced over the code.

.dockerignore
terrywbrady: There is no preference either way.

terrywbrady:
Thanks @PTrottier .  If you had a chance to run the build with and without these changes, it would be good to get a confirmation that the updated build is faster.

Dockerfile.dependencies
```diff
@@ -0,0 +1,25 @@
+# This image will be published as dspace/dspace-dependencies
+# The purpose of this image is to make the build for dspace/dspace run faster 
+
+# Step 1 - Run Maven Build
+FROM maven:3-jdk-8 as build
```
pnbecker: Does that mean we always build on JDK8, even if we want to test with JDK7? (Yes, I know its eol)

pnbecker (approved):
Reviewed, not tested. 👍

Dockerfile.dependencies
```diff
@@ -0,0 +1,25 @@
...
+
+# Copy the DSpace source code into the workdir (excluding .dockerignore contents)
+ADD --chown=dspace . /app/
+COPY dspace/src/main/docker/local.cfg /app/local.cfg
+
+# Trigger the installation of all maven dependencies including the Mirage2 dependencies
+# Clean up the built artifacts in the same step to keep the docker image small
+RUN mvn package -Dmirage2.on=true && mvn clean
+
+# clear the contents of the /app directory so no artifacts are left when dspace:dspace is built
```
PTrottier:
```suggestion
# Clear the contents of the /app directory so no artifacts are left when dspace:dspace is built
```
Just to keep things consistent

PTrottier (approved):
@terrywbrady Thank you for your dedication to improving the DSpace workflow.
Here are the stats from my laptop when running `docker build -f Dockerfile.jdk8 .` (and for `.jdk8-test` Dockerfile)

Before #2307:
real 17m23.300s (JDK8)
real 16m43.603s (JDK8-test)

After #2307 
real	11m17.834s (JDK8)
real	11m20.954s (JDK8-test)

I think we are ready to the benefits of this PR, I added some soft suggestions.

Dockerfile.dependencies
terrywbrady: @pnbecker , the DSpace 5 Dockerfiles offer a jdk7 and jdk8 options since those jdks are referenced in the installation notes.  For DSpace 6, only the jdk8 Dockerfiles are offered.  New Dockerfiles would need to be added if that support is necessary.

PR #2319, opened 2019-01-08
[DS-4136] Resolve hibernate cache size issue when running large batch process

terrywbrady:
https://jira.duraspace.org/browse/DS-4136


terrywbrady:
Closing, this change did not clear the cache

PR #2322, opened 2019-01-11
DS-4142 updated maven docker build to specify jdk8

J4bbi:
Docker build is failing for DSpace 6 because maven latest now (since Nov) uses jdk11.

### DIFF
--- a/Dockerfile.jdk8
+++ b/Dockerfile.jdk8
@@ -9,7 +9,7 @@
 # - default tag for branch: dspace/dspace: dspace/dspace:dspace-6_x-jdk8
 
 # Step 1 - Run Maven Build
-FROM maven as build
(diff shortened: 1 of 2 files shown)


terrywbrady:
@J4bbi , on line 7 of each of these files there is a comment about the maven version.  It would be good to note that version there as well.

terrywbrady (approved):
Reviewed and tested.  This issue is due to the jdk version assoicated with maven:latest.  Our Docker builds will start failing without this change.

This will be needed on master as well.  For 5x and 4x it will be good to verify that we are using an optimal maven version tag.

tdonohue (approved):
👍 Looks like an obvious win to me. We definitely need to "pin" the version of Maven & JDK, as DSpace isn't always tested on the latest and greatest.

PR #2326, opened 2019-01-14
[DS-4142] dspace-5x: update maven jdks for docker

J4bbi:


terrywbrady (approved):
These changes look like a correct port of https://github.com/DSpace/DSpace/pull/2322.

The build failure seems to be due to unrelated mirage2 build issues on dspace-5_x.

kshepherd (approved):
+1 looks like good backport of #2322 
if we want to kick Travis back off, we should rebase this PR against 5.x since #2330 was just merged to fix that issue

PR #2331, opened 2019-01-15
Mixing entities and plain-text values

benbosman:
This PR is still work in progress, but has been created to allow for an early review.
The goal is to support using both entities and plain text for a metadata field and the approach is explained in https://docs.google.com/document/d/1wEmHirFzrY3qgGtRr2YBQwGOvH1IuTVGmxDIdnqvwxM/edit#heading=h.mw35favtpt8z

The order is visible in the virtual metadata, e.g.:
```json
{
  "metadata": [
    {
      "key": "dc.contributor.author",
      "value": "Simmons, Cameron",
      "language": "*",
      "place": 0,
      "authority": "virtual::732",
      "confidence": -1
    },
    {
      "key": "dc.contributor.author",
      "value": "Plain text author",
      "language": null,
      "place": 1,
      "authority": null,
      "confidence": -1
    },
    {
      "key": "dc.contributor.author",
      "value": "Plain text author 2",
      "language": null,
      "place": 2,
      "authority": null,
      "confidence": -1
    },
    {
      "key": "relation.isAuthorOfPublication",
      "value": "0ffbee3f-e7ea-42bc-92fe-2fbef1a52c0f",
      "language": "*",
      "place": 1,
      "authority": "virtual::889",
      "confidence": -1
    },
    {
      "key": "dc.contributor.author",
      "value": "Van de Velde, Kevin",
      "language": "*",
      "place": 4,
      "authority": "virtual::890",
      "confidence": -1
    }
  ]
}
```

The metadata values with an authority containing virtual and the ID of the relationship are the virtual metadata values.

Tasks to be completed are:
* Merge the latest configurable_entities branch into this branch
* JavaDoc, ITs, license headers
* There's still a small bug in calculating the places in some use cases

There's overlap with https://github.com/DSpace/DSpace/pull/2332 because CRUD on relationships helps to verify the functionality for determining the place when creating relationships.
The PR has already been created to facilitate an early review

Diff between this PR and #2332: https://github.com/atmire/DSpace/compare/w2p-57159_permission-to-create-relations...atmire:w2p-58898_place-column-calculation-error


tdonohue:
@benbosman : Based on the description here, I'm seeing some possible parallels between this PR's effort and @cwilper's work on representing DSpaceObject metadata as a map...see:
* Ticket: https://jira.duraspace.org/browse/DS-4107
* REST Contract: https://github.com/DSpace/Rest7Contract/pull/39/
* REST Implementation: https://github.com/DSpace/DSpace/pull/2287

I wanted to point out these parallels since this PR is still a work-in-progress.  I wonder whether you feel @cwilper's work would help make this implementation *easier*?  If so, we can try and get Chris's work moved along more rapidly & note it as a dependency here.

benbosman:
@tdonohue : We are aware of the overlap, which is why we kept that part very lightweight here.
Once Chris' work has been approved and merged, I would prefer to remove the changes here and use Chris' work instead.
The only change to revert in that case should be `MetadataEntryRest.java`.
I would prefer to avoid "waiting" for Chris' work to continue this work if we can continue (also in Angular) using just a few lines of code

benbosman:
The latest updates to this PR include:
* Merged the latest configurable_entities branch into this branch
* Merged the latest https://github.com/DSpace/DSpace/pull/2332 code into this branch
* ITs (they are all succeeding), license headers
* Fixed the bug in calculating the places in the few broken use cases

JavaDoc still needs to be completed

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/DSpaceRestRepository.java
```diff
@@ -407,4 +408,21 @@ protected void patch(Context context, HttpServletRequest request, String apiCate
         throw new RepositoryMethodNotImplementedException("No implementation found; Method not allowed!", "");
     }
 
+    public T put(HttpServletRequest request, String apiCategory, String model, ID id, JsonNode jsonNode) {
```
AlexanderS: Missing JavaDoc for new public method.

dspace-api/src/main/java/org/dspace/content/RelationshipServiceImpl.java
```diff
@@ -51,45 +56,101 @@ public Relationship create(Context context) throws SQLException, AuthorizeExcept
...
+        context.turnOffAuthorisationSystem();
+        if (!virtualMetadataPopulator.isUseForPlaceTrueForRelationshipType(relationship.getRelationshipType(), true)) {
+            if (!leftRelationships.isEmpty()) {
+                leftRelationships.sort((o1, o2) -> o2.getLeftPlace() - o1.getLeftPlace());
+                for (int i = 0; i < leftRelationships.size(); i++) {
+                    leftRelationships.get(leftRelationships.size() - 1 - i).setLeftPlace(i);
+                }
+                relationship.setLeftPlace(leftRelationships.get(0).getLeftPlace() + 1);
+            } else {
+                relationship.setLeftPlace(0);
```
AlexanderS: So this is this?
```
relationship.setLeftPlace(leftRelationships.size());
```
I think this would be more readable then `leftRelationships.get(0).getLeftPlace() + 1` and it would even work for the case if the relationships are empty.

And is there any reason, why the sorting and reordering is done descending and from the back of the list to the front? Should not something like this also work?

```
leftRelationships.sort((o1, o2) -> o1.getLeftPlace() - o2.getLeftPlace());
for (int i = 0; i < leftRelationships.size(); i++) {
    leftRelationships.get(i).setLeftPlace(i);
}
```

dspace/config/spring/api/core-services.xml
```diff
@@ -170,13 +170,14 @@
         <property name="separator">
             <value>, </value>
         </property>
+        <property name="useForPlace" value="true"/>
```
AlexanderS: Is there any reason why a virtual metadata should not be *used for place*? Or is this just for performance if there are a lot of relations?

AlexanderS (commented):
I just made a code review and added a few comments. Overall the code looks good.

It is a bit pity, that we need to duplicate some code because of the left and right side needs to be handled separately. But that cannot be changed.

dspace-api/src/main/java/org/dspace/content/DSpaceObjectServiceImpl.java
```diff
@@ -546,10 +557,42 @@ public void update(Context context, T dso) throws SQLException, AuthorizeExcepti
              */
             // A map created to store the latest place for each metadata field
             Map<MetadataField, Integer> fieldToLastPlace = new HashMap<>();
-            List<MetadataValue> metadataValues = dso.getMetadata();
+            List<MetadataValue> metadataValues = new LinkedList<>();
+            if (dso.getType() == Constants.ITEM) {
+                metadataValues = itemService.getMetadata((Item) dso, Item.ANY, Item.ANY, Item.ANY, Item.ANY);
+            } else {
+                metadataValues = dso.getMetadata();
+            }
```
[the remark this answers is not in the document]
benbosman: This has been updated to `metadataValues = getMetadata(dso, Item.ANY, Item.ANY, Item.ANY, Item.ANY);`
It removed the dependency on `ItemService`

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/DSpaceRestRepository.java
benbosman: The JavaDoc has been written now

dspace/config/spring/api/core-services.xml
benbosman: This is both for performance and because a single relationship can map multiple metadata fields at once while calculation can only happen on one field (it is a place stored in the relationship)

If you have a relationship between an author and a publication, there's no need to start calculating the publication's place in the author using a metadata field in the author

dspace-api/src/main/java/org/dspace/content/RelationshipServiceImpl.java
benbosman: @AlexanderS thanks for the great suggestion. It does indeed result in the same output but is more readable

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/DSpaceRestRepository.java
AlexanderS: > Method to support updating a DSpace instance.

*DSpace instance*? Maybe *DSpaceObject instance*?

benbosman:
All feedback has been processed now, and it's ready for the final review

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/DSpaceRestRepository.java
benbosman: It's actually not limited to a DSpaceObject.
I've adjusted it to be similar to the patch comment:
`Apply an update to the REST object via JSON PUT`

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/DSpaceRestRepository.java
```diff
@@ -407,4 +408,45 @@ protected void patch(Context context, HttpServletRequest request, String apiCate
...
+            thisRepository.put(context, request, apiCategory, model, id, jsonNode);
+            context.commit();
+        } catch (SQLException | AuthorizeException e) {
+            throw new RuntimeException(e.getMessage(), e);
+        }
+        return findOne(id);
+    }
+
+    /**
+     * Implement this method in the subclass to support updating a DSpace instance.
```
AlexanderS: `DSpace instance` should be replace with `REST object` here, too.

AlexanderS (approved):
Looks good to me now. Just one little wording issue is left.

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/DSpaceRestRepository.java
benbosman: Thanks, I pushed this wording change as well

AlexanderS:
Note: This PR has merge conflicts, now.

benbosman:
I was still running the Integration Tests locally, but the branch has now been updated with the latest configurable_entities branch

dspace-api/src/main/java/org/dspace/content/ItemServiceImpl.java
```diff
@@ -1388,10 +1431,13 @@ public boolean canCreateNewVersion(Context context, Item item) throws SQLExcepti
...
-        MetadataValue metadataValue = constructMetadataValue(
-                MetadataSchemaEnum.RELATION.getName() + "." + relationName);
-        metadataValue.setAuthority("virtual");
+    private RelationshipMetadataValue getRelationMetadataFromOtherItem(Context context, Item otherItem,
+                                                                       String relationName,
+                                                                       Integer relationshipId) {
+        RelationshipMetadataValue metadataValue = constructMetadataValue(context,
+                                                                         MetadataSchemaEnum.RELATION
+                                                                             .getName() + "." + relationName);
+        metadataValue.setAuthority(Constants.VIRTUAL_AUTHORITY_PREFIX + relationshipId);
```
paulo-graca: Please consider that metadataValue can be null because constructMetadataValue might return null.

benbosman: Thanks @paulo-graca for identifying this. I've updated the branch with a check for null values from this method and some other small changes

dspace-spring-rest/src/test/java/org/dspace/app/rest/RelationshipRestRepositoryIT.java
```diff
@@ -190,4 +216,1141 @@ public void findAllRelationshipTest() throws Exception {
...
+                                      .withTitle("Publication1")
+                                      .withAuthor("Testy, TEst")
+                                      .withIssueDate("2015-01-01")
+                                      .withRelationshipType("Publication")
+                                      .build();
+
+        RelationshipType isAuthorOfPublicationRelationshipType = relationshipTypeService
+            .findbyTypesAndLabels(context, entityTypeService.findByEntityType(context, "Publication"),
+                                  entityTypeService.findByEntityType(context, "Person"),
+                                  "isAuthorOfPublication", "isPublicationOfAuthor");
```
tdonohue: Turn authorization back on

dspace-spring-rest/src/test/java/org/dspace/app/rest/RelationshipRestRepositoryIT.java
```diff
@@ -190,4 +216,1141 @@ public void findAllRelationshipTest() throws Exception {
...
+
+        authorizeService.addPolicy(context, publication, Constants.WRITE, user);
+
+        String token = getAuthToken(user.getEmail(), password);
+
+        MvcResult mvcResult = getClient(token).perform(post("/api/core/relationships")
+                                                           .param("relationshipType",
+                                                                  isAuthorOfPublicationRelationshipType.getID()
+                                                                                                       .toString())
+                                                           .contentType(MediaType.parseMediaType("text/uri-list"))
```
tdonohue: All of the hardcoded `text/uri-list` strings in your Tests should likely be replaced with `org.springframework.data.rest.webmvc.RestMediaTypes.TEXT_URI_LIST_VALUE`   We don't need to be hardcoding this over and over again, just like we don't hardcode `application/json` in our tests.

tdonohue (dismissed):
@benbosman : I gave this PR a more thorough review today.  Overall, I think the code is looking good, but I think the tests could use some (minor) work/improvements. There's also some areas of code that I feel could be better commented / documented, as the code logic is not always self-explanatory or straightforward.

dspace-api/src/main/java/org/dspace/content/RelationshipServiceImpl.java
```diff
@@ -52,45 +58,103 @@ public Relationship create(Context context) throws SQLException, AuthorizeExcept
...
-                                                                              relationship.getRightItem(),
+                                                                              rightItem,
                                                                               relationship.getRelationshipType(),
                                                                               false);
 
-        if (!leftRelationships.isEmpty()) {
-            leftRelationships.sort((o1, o2) -> o2.getLeftPlace() - o1.getLeftPlace());
-            for (int i = 0; i < leftRelationships.size(); i++) {
-                leftRelationships.get(leftRelationships.size() - 1 - i).setLeftPlace(i + 1);
+        context.turnOffAuthorisationSystem();
```
tdonohue: Why are we turning off authorization here?  I'm a little nervous anytime I see us turn off authorization *without* an inline comment to describe it.  Shouldn't the user have proper permissions to do these updates?

benbosman (commented):
Thanks @tdonohue for your review.

Raf has committed improvements and I've answered your comments inline

dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/RelationshipRestRepository.java
```diff
@@ -60,4 +91,89 @@ public RelationshipRest findOne(Context context, Integer integer) {
...
+                .find(context, Integer.parseInt(req.getParameter("relationshipType")));
+
+            EPerson ePerson = context.getCurrentUser();
+            if (authorizeService.authorizeActionBoolean(context, leftItem, Constants.WRITE) ||
+                authorizeService.authorizeActionBoolean(context, rightItem, Constants.WRITE)) {
+                relationship.setLeftItem(leftItem);
+                relationship.setRightItem(rightItem);
+                relationship.setRelationshipType(relationshipType);
+                relationship = relationshipService.create(context, relationship);
+                context.turnOffAuthorisationSystem();
```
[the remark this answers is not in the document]
tdonohue: @benbosman : I think we should add a comment to this `context.turnOffAuthorizationSystem();` call that says what you just stated. We should make it clear why we are turning off the authorization.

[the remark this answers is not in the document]
benbosman: We'll add a comment here

dspace-spring-rest/src/test/java/org/dspace/app/rest/RelationshipRestRepositoryIT.java
```diff
@@ -190,4 +216,1141 @@ public void findAllRelationshipTest() throws Exception {
...
+
+        Item publication = ItemBuilder.createItem(context, col3)
+                                      .withTitle("Publication1")
+                                      .withIssueDate("2015-01-01")
+                                      .withRelationshipType("Publication")
+                                      .build();
+        RelationshipType isAuthorOfPublicationRelationshipType = relationshipTypeService
+            .findbyTypesAndLabels(context, entityTypeService.findByEntityType(context, "Publication"),
+                                  entityTypeService.findByEntityType(context, "Person"),
+                                  "isAuthorOfPublication", "isPublicationOfAuthor");
```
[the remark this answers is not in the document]
tdonohue: This test method leaves the authorization system *off* for all tests.  Again, as I comment frequently, we need to ONLY turn off authorization *briefly* in tests. It is **not acceptable** to have tests which leave the authorization system turned off during the entirety of the test.

This method creates/updates  test content in several places...so it needs to be updated to surround all sections with `context.turnOffAuthorisationSystem();` and `context.restoreAuthSystemState();`

dspace-spring-rest/src/test/java/org/dspace/app/rest/RelationshipRestRepositoryIT.java
```diff
@@ -190,4 +216,1141 @@ public void findAllRelationshipTest() throws Exception {
...
```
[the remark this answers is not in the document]
tdonohue: Just like above comment, this method is leaving authorization system turned off for the entire method. Please correct it.

dspace-spring-rest/src/test/java/org/dspace/app/rest/RelationshipRestRepositoryIT.java
```diff
@@ -190,4 +216,1141 @@ public void findAllRelationshipTest() throws Exception {
...
```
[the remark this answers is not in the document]
tdonohue: Again, authorization system is left turned off for the entire test method

dspace-spring-rest/src/test/java/org/dspace/app/rest/RelationshipRestRepositoryIT.java
```diff
@@ -190,4 +216,1141 @@ public void findAllRelationshipTest() throws Exception {
...
+        Item orgUnit1 = ItemBuilder.createItem(context, col3)
+                                   .withTitle("OrgUnit1")
+                                   .withIssueDate("2015-01-01")
+                                   .withRelationshipType("OrgUnit")
+                                   .build();
+
+        RelationshipType isOrgUnitOfPersonRelationshipType = relationshipTypeService
+            .findbyTypesAndLabels(context, entityTypeService.findByEntityType(context, "Person"),
+                                  entityTypeService.findByEntityType(context, "OrgUnit"),
+                                  "isOrgUnitOfPerson", "isPersonOfOrgUnit");
```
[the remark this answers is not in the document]
tdonohue: And again, another new test method with authorization left turned off for the entire method.

dspace-spring-rest/src/test/java/org/dspace/app/rest/RelationshipRestRepositoryIT.java
```diff
@@ -190,4 +216,1141 @@ public void findAllRelationshipTest() throws Exception {
...
```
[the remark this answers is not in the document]
tdonohue: And, again, authorization left turned off here.

tdonohue (commented):
@benbosman : I've reviewed this again today by doing a [diff between this PR and #2332](https://github.com/atmire/DSpace/compare/w2p-57159_permission-to-create-relations...atmire:w2p-58898_place-column-calculation-error).

The code itself looks good. But, there are still several (new) test methods that are incorrectly leaving all Authorization checks turned off.  We need to be very strict about not allowing this behavior (it's potentially dangerous -- as the tests may not always be fully trustworthy).  Once these tests are corrected, then I'll gladly add a +1 here.

dspace-api/src/main/java/org/dspace/content/ItemServiceImpl.java
paulo-graca: 👍

paulo-graca: @benbosman I noticed the added verifications, but the one I pointed out is still unverified. The `metadataValue.setAuthority` on line 1440.

benbosman: @paulo-graca if I'm not mistaken, you're referring to https://github.com/DSpace/DSpace/blob/970b66e/dspace-api/src/main/java/org/dspace/content/ItemServiceImpl.java#L1458 in the current codebase?
This is verifying whether metadata value is null.

paulo-graca: My fault! Thank you! 👍

paulo-graca (approved):
Thank you @benbosman !

dspace-spring-rest/src/test/java/org/dspace/app/rest/RelationshipRestRepositoryIT.java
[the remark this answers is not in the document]
benbosman: This was resolved by https://github.com/DSpace/DSpace/pull/2331/commits/970b66edb2f9d8cd9a9824f0a1be266073c2a139#diff-61a3266941bcb44b91094e4d3f330ba6R790 and below

dspace-spring-rest/src/test/java/org/dspace/app/rest/RelationshipRestRepositoryIT.java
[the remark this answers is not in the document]
benbosman: This was resolved by https://github.com/DSpace/DSpace/pull/2331/commits/970b66edb2f9d8cd9a9824f0a1be266073c2a139#diff-61a3266941bcb44b91094e4d3f330ba6R997 and below

dspace-spring-rest/src/test/java/org/dspace/app/rest/RelationshipRestRepositoryIT.java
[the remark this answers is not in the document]
benbosman: This was resolved by https://github.com/DSpace/DSpace/pull/2331/commits/970b66edb2f9d8cd9a9824f0a1be266073c2a139#diff-61a3266941bcb44b91094e4d3f330ba6R1208 and below

dspace-spring-rest/src/test/java/org/dspace/app/rest/RelationshipRestRepositoryIT.java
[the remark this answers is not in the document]
benbosman: This was resolved by https://github.com/DSpace/DSpace/pull/2331/commits/970b66edb2f9d8cd9a9824f0a1be266073c2a139#diff-61a3266941bcb44b91094e4d3f330ba6R1331 and below

benbosman (commented):
@tdonohue Raf had implemented these changes a few days ago, and has now also completed the merges to make sure this branch is compatible with configurable_entities.
It has been verified using both the ITs and manually using in the HAL browser

tdonohue (approved):
👍 This looks good to me now.  Thanks for the updates @benbosman !   As this has already received approval from @paulo-graca and @AlexanderS , I'll merge it shortly.

PR #2337, opened 2019-01-28
[DS-4150] Ensure dspace.restUrl is supported out of the box for DSpace 7 Docker

terrywbrady:
https://jira.duraspace.org/browse/DS-4150


tdonohue (approved):
👍 Looks good. The RESTv7 webapp is currently created named `spring-rest`, so we should make this configuration default to the same name.

tdonohue (approved):

abollini (approved):

PR #2341, opened 2019-02-01
DS-4159

AndrewZachWood:
For Jira: [DS-4159](https://jira.duraspace.org/browse/DS-4159)
Fixes #7501


nwoodward (approved):
:+1: Looks good.

PR #2342, opened 2019-02-06
[DS-4163] XOAI.java skip empty list and dont add it to solr

ssolim:
https://jira.duraspace.org/browse/DS-4163

### DIFF
--- a/dspace-oai/src/main/java/org/dspace/xoai/app/XOAI.java
+++ b/dspace-oai/src/main/java/org/dspace/xoai/app/XOAI.java
@@ -300,9 +300,11 @@ public class XOAI {
                 }
             }
             System.out.println("Total: " + i + " items");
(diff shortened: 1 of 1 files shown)


mwoodiupui (approved):
+1 by inspection.  Obviously correct.

tdonohue (approved):
👍 Looks correct to me too.  This needs to be cherry-picked (or ported) to `master` as well, since it fixes code that was applied to both 6.x and master in https://jira.duraspace.org/browse/DS-4136

ulb-sa-schmilj:
I don't understand the fix. It is also possible to add an empty list to solr.

For example - the batchSize is 50, we have exactly 100 items in the repository. We add the 100 items inside the while loop. After the loop "i > 0" is true, but the list is also empty...

PR #2343, opened 2019-02-06
[DS-4164] fixed the bug in the error handling

Raf-atmire:
This fixes the issue as described in Jira: https://jira.duraspace.org/browse/DS-4164


tdonohue (approved):
👍  This looks correct to me. I'm not sure why this controller is even implementing `ErrorController`.  This also passes [existing integration tests for OpenSearch,](https://github.com/DSpace/DSpace/blob/master/dspace-spring-rest/src/test/java/org/dspace/app/opensearch/OpenSearchControllerIT.java) so there seems to be no loss in functionality.

abollini (approved):

PR #2347, opened 2019-02-14
[DS-4163] OAI exception on empty list

mwoodiupui:
https://jira.duraspace.org/browse/DS-4163
Cherry-picked from #2342 .


tdonohue (approved):
👍 Change looks correct to me. Assuming Travis CI also approves, I think this is good to merge.  Thanks @mwoodiupui for porting this to master.

PR #2350, opened 2019-02-18
DS-4169: Treat empty language as null in edit item

AlexanderS:
The metadata editor in the admin edit item view always submits an empty string
if there is no language (the HTML input fields cannot submit null values). To
preserve the null values of the language attributes all empty strings in the
language input fields are treated as null values, now.

https://jira.duraspace.org/browse/DS-4169

### DIFF
--- a/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/administrative/FlowItemUtils.java
+++ b/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/administrative/FlowItemUtils.java
@@ -14,6 +14,7 @@ import java.util.*;
 
 import org.apache.cocoon.environment.Request;
 import org.apache.cocoon.servlet.multipart.Part;
+import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.time.DateUtils;
 import org.dspace.app.util.Util;
 import org.dspace.app.xmlui.utils.UIException;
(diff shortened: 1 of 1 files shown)


dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/administrative/FlowItemUtils.java
```diff
@@ -211,6 +212,12 @@ public static FlowResult processEditItem(Context context, UUID itemID, Request r
 			String lang = request.getParameter("language_"+index);
 			String remove = request.getParameter("remove_"+index);
 			
+			// Empty lang is null lang
+			if (StringUtils.isEmpty(lang))
```
nwoodward: Since isEmpty() doesn't trim the input string, would isBlank() work better here? I wonder about the off chance that the user leaves a space in the field.  https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/StringUtils.html

mwoodiupui: Well spotted.  I agree.

terrywbrady: This change seems like a good idea.  I do not have a repository available to test the change.

AlexanderS: Thanks, fixed this.

nwoodward (approved):
:+1:  I tested this on a local deployment and confirmed it works as expected. The dspace-6_x code changes null text_lang instances to empty strings whereas this code leaves them as null, including when there's a space :slightly_smiling_face:.

mwoodiupui (approved):
+1 by inspection.

PR #2358, opened 2019-03-06
DS-3444 JSPUI must keep shibboleth attributes on session renewal

pnbecker:
Fixes #6797

When a user logs into DSpace, JSPUI creates a new session to prevent session hijacking. While doing so, it must carry over to the information shibboleth stored to the old session. These information are:
- that a the actual user was authorized using shibboleth, and
- the special groups that were assigned to the user by shibboleth attributes, if any.

This fix was developed by The Library Code GmbH with the support of the Universitätsbibliothek Mainz and the Fraunhofer IRB, Research Services & Open Science.

### DIFF
--- a/dspace-jspui/src/main/java/org/dspace/app/webui/util/Authenticate.java
+++ b/dspace-jspui/src/main/java/org/dspace/app/webui/util/Authenticate.java
@@ -12,6 +12,7 @@ import java.sql.SQLException;
 import java.util.Iterator;
 import java.util.Locale;
 import java.util.UUID;
+import java.util.List;
 
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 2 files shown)


pnbecker:
This must **not** be cherry-picked to master. We might consider bringing this back to DSpace 5 and/or 4.

To test this: Setup an installation of DSpace 6 with Shibboleth (you can use https://samltest.id). Create a Group, create a collection and add the group to the submitters. Configure shibboleth to use a shibboleth attribute to assign the group you created to a user as special groups. Login as this user and try to create a new submission. You will not be able to select the collection as the special group got lost. Apply the patch and try again.

lap82 (approved):
+1 cherry-picked in 5_x and tested (not yet in DSpace 6.x but I guess that it works)

kshepherd:
fwiw i am +1 on this too but as I worked with @pnbecker on some related LDAP work i'm not sure if i should formally approve it. any takers?

kshepherd (approved):
we've had another +1 (and a 5.x port), merging now!

PR #2369, opened 2019-03-12
[DS-3894] Fixed facet’s "view more" link throwing exception when results are sorted

Atmire-Kristof:
Changed the "order" parameter included in facet’s "view more" links to "sort_order" to avoid conflicts with the order parameter used in discovery.
This fixes the exception being thrown when clicking a search facet while being on a sorted discovery page.

Related Jira issue: https://jira.duraspace.org/browse/DS-3894

[The King Abdullah University of Science and Technology (KAUST) Repository](https://repository.kaust.edu.sa/) commissioned Atmire to fix and contribute this issue as an effort to continually improve the DSpace codebase.


kaladay:
I failed to reproduce this and discovered that `order` is now being converted to `filterorder`.
After some investigation, it seems that this was resolved in a similar way and committed via this PR: #2026 .

alanorth:
@kaladay is correct that this issue is not present in DSpace 6.3. I went to reproduce it and found that #2026 had already fixed it.

This pull request should be closed. It *does not need to be merged for DSpace 6.4*.

PR #2371, opened 2019-03-12
[DS-3881] Fixed discovery "view more" facet displaying incorrect totals

Atmire-Kristof:
Because the "view more" page for a discovery facet shows the number of items returned by the search rather than the number of values in the relevant field, the total amount being displayed is incorrect.
To fix this, we can simply hide the total amount, so it will show "Now showing items x-y" instead of "Now showing items x-y of z", where "z" was incorrect.

Related Jira issue: https://jira.duraspace.org/browse/DS-3881
Fixes #7228

[The King Abdullah University of Science and Technology (KAUST) Repository](https://repository.kaust.edu.sa/) commissioned Atmire to fix and contribute this issue as an effort to continually improve the DSpace codebase.

### DIFF
--- a/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/discovery/SearchFacetFilter.java
+++ b/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/discovery/SearchFacetFilter.java
@@ -355,7 +355,7 @@ public class SearchFacetFilter extends AbstractDSpaceTransformer implements Cach
 
                     // We put our total results to -1 so this doesn't get shown in the results (will be hidden by the xsl)
                     // The reason why we do this is because solr 1.4 can't retrieve the total number of facets found
-                    results.setSimplePagination((int) queryResults.getTotalSearchResults(), offSet + 1,
+                    results.setSimplePagination(-1, offSet + 1,
                                                     shownItemsMax, getPreviousPageURL(browseParams, request), nextPageUrl);
 
                     Table singleTable = results.addTable("browse-by-" + facetField + "-results", (int) (queryResults.getDspaceObjects().size() + 1), 1);


toniprieto:
I have just tested it. It may be a problem related with my test instance, but to make it works I had to do two extra changes on file  `dspace-xmlui-mirage2/src/main/webapp/xsl/core/attribute-handlers.xsl`

This file has some `<i18n:text>`  with a line break that should be removed:

```
<i18n:text>xmlui.dri2xhtml.structural.pagination-info.nototal
</i18n:text>
```
instead of

`<i18n:text>xmlui.dri2xhtml.structural.pagination-info.nototal</i18n:text>`

On lines:

https://github.com/DSpace/DSpace/blob/dspace-6_x/dspace-xmlui-mirage2/src/main/webapp/xsl/core/attribute-handlers.xsl#L125-L126

https://github.com/DSpace/DSpace/blob/dspace-6_x/dspace-xmlui-mirage2/src/main/webapp/xsl/core/attribute-handlers.xsl#L226-L227

On the other hand, I think that the proposed change is correct. In fact, the comment previous to the line of code explains why the value should by -1.

alanorth:
This is a welcome fix for an old and annoying bug. Tested and confirmed working on DSpace 6.3.

@toniprieto is correct, though, there is an extra newline in the two i18n:text tags here that need to be fixed:

```diff
diff --git a/dspace-xmlui-mirage2/src/main/webapp/xsl/core/attribute-handlers.xsl b/dspace-xmlui-mirage2/src/main/webapp/xsl/core/attribute-handlers.xsl
index d122c018f1..8c66a92584 100644
--- a/dspace-xmlui-mirage2/src/main/webapp/xsl/core/attribute-handlers.xsl
+++ b/dspace-xmlui-mirage2/src/main/webapp/xsl/core/attribute-handlers.xsl
@@ -122,8 +122,7 @@
                                         <i18n:translate>
                                             <xsl:choose>
                                                 <xsl:when test="parent::node()/@itemsTotal = -1">
-                                                    <i18n:text>xmlui.dri2xhtml.structural.pagination-info.nototal
-                                                    </i18n:text>
+                                                    <i18n:text>xmlui.dri2xhtml.structural.pagination-info.nototal</i18n:text>
                                                 </xsl:when>
                                                 <xsl:otherwise>
                                                     <i18n:text>xmlui.dri2xhtml.structural.pagination-info</i18n:text>
@@ -223,8 +222,7 @@
                                         <i18n:translate>
                                             <xsl:choose>
                                                 <xsl:when test="parent::node()/@itemsTotal = -1">
-                                                    <i18n:text>xmlui.dri2xhtml.structural.pagination-info.nototal
-                                                    </i18n:text>
+                                                    <i18n:text>xmlui.dri2xhtml.structural.pagination-info.nototal</i18n:text>
                                                 </xsl:when>
                                                 <xsl:otherwise>
                                                     <i18n:text>xmlui.dri2xhtml.structural.pagination-info</i18n:text>
```

I recommend someone (either @Atmire-Kristof or perhaps a maintainer like @J4bbi) merges this and pushes the fix above as well.

alanorth:
Here is a before and after of the "view more" for countries in our repository (with the additional XSL fix from @toniprieto).

Before:

Screenshot 2021-07-24 at 09-28-01 Filter by Country-fs8

After:

Screenshot 2021-07-24 at 09-28-18 Filter by Country-fs8

tdonohue:
@Atmire-Kristof : This looks to have approval for 6.4...but two small changes were requested above: https://github.com/DSpace/DSpace/pull/2371#issuecomment-885887580  Any chance you could make those changes part of this PR so that we can get this into 6.4?

alanorth:
I tested this on top of current `dspace-6_x` and it still works, with the newline fixes mentioned by @toniprieto and I above. I'm going to update this manually and merge it since the original PR author has not responded and this is a welcome fix.

PR #2374, opened 2019-03-15
DS-4190. Word-break CSS class breaks words without hyphens in Firefox…

J4bbi:
… and Opera mini
This CSS class change creates hyphens in Firefox / Opera mini when breaking words.


kshepherd (changes requested):
this looks great, thanks :) 
i'm +1, but just a note: the attribution for the CSS is a nice touch, but to keep things consistent with how other CSS is managed in dspace, removing the comment might keep things cleaner. do you mind making that additional commit to this PR?

J4bbi:
DS-2963 Word wrap and hyphenation in Item Full View
is the same issue

kshepherd (approved):

nwoodward (approved):
:+1: Tested and this looks good.

kshepherd:
merged, thanks @J4bbi

PR #2378, opened 2019-03-19
NPE by uploading unregistered format file with embargo because of conflicting error code number

qzhao-ulb:
By uploading a file with a unregistered format in submission it will throws a NullPointException, if I use the uploading with the embargo feature. After processing of upload it returns an error code of unknown format with digit 10, but the class UploadWithEmbargo in XMLUI checks it with the error code for edit item policy as well, the same digit, then it calls AccessStep later and throws an NullPointException.

### DIFF
--- a/dspace-api/src/main/java/org/dspace/submit/step/UploadStep.java
+++ b/dspace-api/src/main/java/org/dspace/submit/step/UploadStep.java
@@ -78,7 +78,7 @@ public class UploadStep extends AbstractProcessingStep
     public static final int STATUS_NO_FILES_ERROR = 5;
 
     // format of uploaded file is unknown
-    public static final int STATUS_UNKNOWN_FORMAT = 10;
+    public static final int STATUS_UNKNOWN_FORMAT = 11;
 
     // virus checker unavailable ?
     public static final int STATUS_VIRUS_CHECKER_UNAVAILABLE = 14;


nwoodward:
I'm having trouble reproducing the null pointer exception with an unregistered file format. Can you provide a stack trace or point to where in the UploadWithEmbargo class in the XMLUI the incorrect error code causes the problem?

qzhao-ulb:
I didn't get the error stack trace once again. But I copied the error description from our internal report.

"when UploadWithEmbargoStep is active and uploading a file with an unknown format, the errorFlag will be 10 = org.dspace.submit.step.UploadStep.STATUS_UNKNOWN_FORMAT. 
It runs in the if condition "else if(this.errorFlag==org.dspace.submit.step.AccessStep.STATUS_EDIT_POLICY"  (line 146 in org.dspace.app.xmlui.aspect.submission.submit.UploadWithEmbargoStep) according to  STATUS_EDIT_POLICY has value of 10.
Then object named editPolicy (org.dspace.app.xmlui.aspect.submission.submit.UploadWithEmbargoStep: line 149) is not null, but  line 182 calls editPolicy.addBody(body) and throws a NPE (NullPointException), because the object resourcePolicy in EditPolicyStep (line 76) is null."

PR #2380, opened 2019-03-22
[DS-4187] Tool to dump and restore the 'authority' Solr core

mwoodiupui:
WIP because I need to find a way to generate some authority records for testing.  At this moment the tool has been tested only on an empty authority core (and it succeeds).

'dspace authority -d' dumps the core to standard out.
'dspace authority -r' loads a dump from standard in.  Add '-c' to remove all existing authority records first.


mwoodiupui:

PR #2382, opened 2019-03-24
[DS-4187] Tool to dump and restore the 'authority' Solr core (6_x)

mwoodiupui:


mwoodiupui:
Now tested with a single record.  Corrected much silliness.

mwoodiupui:
Now tested with enough records to exhaust the dumper's Solr response window and trigger a refill.

mwoodiupui:

PR #2384, opened 2019-03-25
[DS-4197] StatisticsDataVisits: Use dso length value

### LINKED ISSUE #7537: [DS-4197] Item names are truncated in statistics
<i>Imported from JIRA <a href="https://jira.lyrasis.org/browse/DS-4197">[DS-4197]</a> created by sulfrian</i>
<p>Item names in the statistics are truncated at the first space. The length is not set and so initialized with 0 and the code is looking for the first space after the desired length.</p>

<p> </p>

<p>This bug is introduced in this commit: <a href="https://github.com/DSpace/DSpace/commit/54222f3c1d3c63bd03af263703a950e9cfad7439#diff-f0ba6e58dd68fe26a75c4c23ed214379L717" class="external-link" target="_blank" rel="nofollow noopener">https://github.com/DSpace/DSpace/commit/54222f3c1d3c63bd03af263703a950e9cfad7439#diff-f0ba6e58dd68fe26a75c4c23ed214379L717</a></p>

AlexanderS:
The supplied length was unused before and always got initialized with 0.

https://jira.duraspace.org/browse/DS-4197
Fixes #7537 

_(I will backport this fix to dspace-6_x once merged in master. A backport to dspace-5_x is not required because this bug was introduced in dspace-6_x.)_

### DIFF
--- a/dspace-api/src/main/java/org/dspace/statistics/content/StatisticsDataVisits.java
+++ b/dspace-api/src/main/java/org/dspace/statistics/content/StatisticsDataVisits.java
@@ -781,6 +781,7 @@ public class StatisticsDataVisits extends StatisticsData {
         public void setDso(DSpaceObject dso, int dsoType, int length) {
             this.dsoType = dsoType;
             this.dso = dso;
+            this.dsoLength = length;
         }
 
         public void setDsoType(int dsoType) {


tdonohue (approved):
👍 I haven't tested this, but it looks like an obvious bug.

alanorth:
Hi @AlexanderS. This patch looks obviously correct, but I'm wondering where the bug is. I don't see any filenames in my Solr statistics core. I only see bitstream IDs...?

bram-atmire:
On closer review this is still applicable — reopening.

The one-line fix in this PR (`this.dsoLength = length;`) has not been applied to current `main`: the three-argument `setDso(DSpaceObject dso, int dsoType, int length)` at [`StatisticsDataVisits.java:784`](https://github.com/DSpace/DSpace/blob/main/dspace-api/src/main/java/org/dspace/statistics/content/StatisticsDataVisits.java#L784) still receives `length` and silently drops it. In practice the separate `setDsoLength(...)` call elsewhere masks the bug, but the 3-arg method remains a latent trap.

Happy to keep this in the queue as a small defensive fix worth porting. Thanks @AlexanderS.

PR #2385, opened 2019-03-25
DS-4198: Create On-Demand build of Docker Image for a PR

terrywbrady:
### Background Links
- https://github.com/DSpace-Labs/DSpace-Docker-Images/issues/103
- https://jira.duraspace.org/browse/DS-4198

The automated builds could be configured as follows

image

A committer or contributor with DockerHub build rights could trigger the build the of image.

When the tag name is in the form prNNNN, the appropriate PR code will be merged before the maven build.  See the following sample output.

```
Step 9/28 : RUN PRNUM=`echo ${DOCKER_TAG}| egrep "^pr([0-9]+)$" | sed -e s/pr//` && if [ -n "$PRNUM" ]; then echo "Merging $PRNUM"; curl -o /tmp/pr.patch -L https://github.com/DSpace/DSpace/pull/$PRNUM.diff; git apply /tmp/pr.patch; fi
---> Running in 326e043de993
Merging 2379
% Total % Received % Xferd Average Speed Time Time Time Current
Dload Upload Total Spent Left Speed
0 0 0 0 0 0 0 0 --:--:-- --:--:-- --:--:-- 0 100 139 0 139 0 0 583 0 --:--:-- --:--:-- --:--:-- 584
100 1593 0 1593 0 0 3520 0 --:--:-- --:--:-- --:--:-- 3520
Removing intermediate container 326e043de993
```
## Note that a dockerhub build hook is needed to pass the tag to the build

- The build hook must be relative to the dockerfile
- ~~Is it time to move all the Dockerfiles to dspace/src/main/docker?~~
  - The old dockerhub build system did not permit this

### DIFF
--- a/Dockerfile.jdk8-test
+++ b/Dockerfile.jdk8-test
@@ -11,6 +11,8 @@
 # Step 1 - Run Maven Build
 FROM dspace/dspace-dependencies:dspace-7_x as build
 ARG TARGET_DIR=dspace-installer
+# The docker hub build will set the following environment variable
+ARG DOCKER_TAG=${DOCKER_TAG}
 WORKDIR /app
 
 # The dspace-install directory will be written to /install
@@ -23,6 +25,14 @@ USER dspace
 # Copy the DSpace source code into the workdir (excluding .dockerignore contents)
 ADD --chown=dspace . /app/
 COPY dspace/src/main/docker/local.cfg /app/local.cfg
+
+# if DOCKER_TAG is in the format prNNNN then merge code for that PR on top of the current branch
+RUN PRNUM=`echo ${DOCKER_TAG}| egrep "^pr([0-9]+)$" | sed -e s/pr//` && \
+    if [ -n $PRNUM ]; \ 
+    then echo "Merging $PRNUM"; \
+      curl -o /tmp/pr.patch -L https://github.com/DSpace/DSpace/pull/$PRNUM.diff; \
+      git apply /tmp/pr.patch; \
+    fi
 
 # Build DSpace.  Copy the dspace-install directory to /install.  Clean up the build to keep the docker image small
 RUN mvn package && \
--- a/Dockerfile.jdk8-test
+++ b/Dockerfile.jdk8-test
@@ -26,7 +26,7 @@ USER dspace
 ADD --chown=dspace . /app/
 COPY dspace/src/main/docker/local.cfg /app/local.cfg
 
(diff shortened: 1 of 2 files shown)


terrywbrady:
The DOCKER_TAG variable is not available as I had hoped.  I have submitted the following questions for help.
- https://stackoverflow.com/questions/55347752/dockerhub-autobuild-for-a-github-pull-request-using-the-docker-tag-variable
- https://forums.docker.com/t/passing-variables-to-an-automated-build-docker-tag/71875

terrywbrady:
Adding a build hook to attempt to pass a variable to the build.

Build hooks must be relative to the Dockerfile.

Note: now that the new automated build system is in place for Dockerhub, it would be possible to move the Dockerfiles to dspace/src/main/docker.

terrywbrady:
## Closing on Recommendation from the 2019-03-27 Dev Meeting

Terry Brady [1:52 PM]
I would like to have a mechanism to build "feature" images for priority PR's.
I created a PR that could accomplish this with a dockerhub build.
https://github.com/DSpace/DSpace/pull/2385
But I do not necessarily love this approach.
Another option would be for us to build "feature" branches in DSpace/DSpace as needed to illustrate changes that need user input.
If that was done, we could simply auto-build an image each time a feature branch is updated.
Do you all have thoughts on this?

Tim Donohue [1:56 PM]
I guess I'd first ask about the use case here.  If the primary goal is to make testing larger PRs *easier*, then moving towards a centralized "feature" branch approach (like with `configurable_entities`) makes a lot of sense.  Large efforts likely should be central branches anyhow (and they are not always).
If the goal is to make testing *all* PRs easier, a centralized "feature" branch approach won't work so well...as only Committers can create a centralized "feature" branch...and most PRs don't come from Committers

Terry Brady [1:57 PM]
I think the feature branch approach is cleaner.

Tim Donohue [1:57 PM]
I agree it sounds cleaner

Terry Brady [1:57 PM]
Perhaps we establish a naming convention like feature_...
I will close PR 2385 and add this discussion.

Tim Donohue [1:58 PM]
I'm just pointing out that would mean that we'd only have these Docker builds for *some* PRs...those that have feature branches

Terry Brady [1:59 PM]
We could revive my PR 2308 (I did a strikethrough on it in the notes) if we want every PR built as an image

Tim Donohue [1:59 PM]
A naming convention for feature branches sounds good to me too

PR #2389, opened 2019-03-28
DS-4201 Do not pass starts_with if browsing an item index.

J4bbi:
Fixes [DS-4192](https://jira.duraspace.org/browse/DS-4192).


terrywbrady:
I tested this PR to see if it could also resolve https://jira.duraspace.org/browse/DS-4208.  The issue is similar.  Unfortunately, this PR does not also resolve that bug.  If I find a solution to DS-4208, I will add a comment to this PR.

terrywbrady (approved):
Tested and verified the fix

nwoodward (approved):
:+1: Tested and confirmed this works.

kshepherd:
merged, thanks @J4bbi

PR #2395, opened 2019-04-04
[DS-1892] Honour controlled vocab store, delimiter settings

kaladay:
Replace PR #460 rebased against a more recent master and attempt to follow new syntax guidelines.
I made no functional changes from the original authors work.


bram-atmire:
Closing — the class this PR modifies, `org.dspace.vocabulary.ControlledVocabulary`, is no longer referenced anywhere in the codebase (grepping `dspace-api/` + `dspace-server-webapp/` for imports turns up nothing on current `main`).

The DSpace 7+ REST API uses `org.dspace.content.authority.DSpaceControlledVocabulary` instead, which already supports the configuration options this PR adds — `vocabulary.plugin.<name>.hierarchy.store` and `vocabulary.plugin.<name>.delimiter` are both read and honored in that class.

Thank you for the original rebase; the feature itself is available in modern DSpace via the newer vocabulary implementation.

PR #2396, opened 2019-04-05
DS-4211: Set Item iterator to emptyIterator() so that it's never null (6.x)

nwoodward:
Fixes https://jira.duraspace.org/browse/DS-4211

### DIFF
--- a/dspace-api/src/main/java/org/dspace/app/bulkedit/MetadataExport.java
+++ b/dspace-api/src/main/java/org/dspace/app/bulkedit/MetadataExport.java
@@ -122,7 +122,9 @@ public class MetadataExport
                 System.out.print(" ");
             }
             Iterator<Item> items = buildFromCommunity(context, subCommunity, indent + 1);
-            result = addItemsToResult(result,items);
+            if (items != null) {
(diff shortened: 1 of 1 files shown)


mwoodiupui:
This looks like it will work.  But would it not be better to instead initialize `result` to `Collections.emptyIterator<Item>()` rather than `null`?  Then `addItemsToResult()` will *always* return an `Iterator<Item>` and no null check is necessary here or elsewhere.

It might be even better to change the `Iterator` to a `List<Item>`, add `Item`s to it in the normal way, and return the value of its `#iterator()` method.

nwoodward:
Yeah, an empty iterator would be better. I assume you're talking about the Collections from commons-collections? I don't see an emtpyIterator class that takes a type in version 3.2.2, so I thought I'd ask.

mwoodiupui:
No, I'm talking about java.util.Collections.emptyIterator().

nwoodward:
How does it look now?

kshepherd (approved):
+1 looks like an easy win, thanks @nwoodward !

mwoodiupui (approved):
+1 by inspection.  Yes, that should work.

PR #2403, opened 2019-04-12
[DS-4217] applied the name refactor and rewrote BadRequestException t…

Raf-atmire:
Addressed Jira: https://jira.duraspace.org/browse/DS-4217

This refactor will allow us to use this DSpaceBadRequestException in a wide variety of places and fixed a few cases where the BadRequestException would cause a response with error code 500 instead of 400


dspace-spring-rest/src/main/java/org/dspace/app/rest/repository/CommunityRestRepository.java
```diff
@@ -14,7 +14,6 @@
 import java.util.UUID;
 import javax.servlet.ServletInputStream;
 import javax.servlet.http.HttpServletRequest;
-import javax.ws.rs.BadRequestException;
```
abollini: this was the real issue (same for collection and item)

abollini (changes requested):
The reason why the BadRequestException doesn't raise the right error code in some cases is because it was not the right BadRequestException :)
https://github.com/DSpace/DSpace/pull/2403/files#diff-e80ccd8ce091b9966d7106e4deafae92L17

note the "wrong" package.

As importing from the wrong package can be an easy mistake I agree to rename the exception class (also to make it suitable outside the Patch verb).
I would ask to introduce test here to demonstrate that the identified issue on the community, collection and item endpoint have been solved

KevinVdV:
Tests have been added, ready for review.

dspace-spring-rest/src/test/java/org/dspace/app/rest/ItemRestRepositoryIT.java
```diff
@@ -1768,4 +1768,80 @@ private void runPatchMetadataTests(EPerson asUser, int expectedStatus) throws Ex
         new MetadataPatchSuite().runWith(getClient(token), "/api/core/items/" + item.getID(), expectedStatus);
     }
 
+    @Test
+    public void testCreateItemInArchiveFalseBadRequestException() throws Exception {
```
abollini: it is not immediate to understand why this test expect a bad request, it will be nice to add a javadoc for the test or a comment before the itemRest.setInArchive(false); to highlight that this is the issue (we cannot create a not final item on this endpoint)

abollini (approved):
Test looks good, I have only added the suggestion to add a comment, this can be eventually done by the repo maintainer before merge if you agree

paulo-graca (commented):
Thank you @Raf-atmire! This is a relative small PR, the majority of the changes are related with naming refactoring. 
But I think the chosen name might be ambiguous because you already have a similar signature:
org.dspace.app.rest.exception.InvalidRequestException

Perhaps it would be better to also replace all occurrences of org.dspace.app.rest.exception.InvalidRequestException with org.dspace.app.rest.exception.DSpaceBadRequestException and keep a single Exception for bad and invalid requests.

Raf-atmire:
I added a comment as request by @abollini and replaced all the occurrences of InvalidRequestExceptions with the DSpaceBadRequestException as suggested by @paulo-graca

paulo-graca (approved):
Thank you @KevinVdV and @Raf-atmire for considering my suggestions !

tdonohue (approved):
👍 Looks good to me, and all prior requests have now been addressed.  Merging, as this is a +3

PR #2412, opened 2019-04-25
[DS-4228] Fix colspan values in EditItemBitstreamsForm

### LINKED ISSUE #7568: [DS-4228] Wrong colspan values in EditItemBitstreamsForm
<i>Imported from JIRA <a href="https://jira.lyrasis.org/browse/DS-4228">[DS-4228]</a> created by sulfrian</i>
<p>The table was extended in <a href="https://github.com/DSpace/DSpace/commit/3e49b04e325fca595cdde93f02143dc3d6006fa0" class="external-link" target="_blank" rel="nofollow noopener">3e49b04e325fca595cdde93f02143dc3d6006fa0</a> but the colspan values were not adjusted. It may be relevant for the styling of the table, because some rows will not have a right border.</p>

AlexanderS:
The table of the EditItemBitstreamsForm was extended in 3e49b04e325fca595cdde93f02143dc3d6006fa0, but the colspan values were not adjusted.

It may be relevant for the styling of the table, because some rows will not have a right border.

_I can create a backport for dspace-5_x, if this is merged._

Fixes #7568

### DIFF
--- a/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/administrative/item/EditItemBitstreamsForm.java
+++ b/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/administrative/item/EditItemBitstreamsForm.java
@@ -122,7 +122,7 @@ public class EditItemBitstreamsForm extends AbstractDSpaceTransformer {
 		for (Bundle bundle : bundles)
 		{
 
-			Cell bundleCell = files.addRow("bundle_head_" + bundle.getID(), Row.ROLE_DATA, "").addCell(1, 5);
+			Cell bundleCell = files.addRow("bundle_head_" + bundle.getID(), Row.ROLE_DATA, "").addCell(1, 6);
 			bundleCell.addContent(T_bundle_label.parameterize(bundle.getName()));
 
 			java.util.List<Bitstream> bitstreams = bundle.getBitstreams();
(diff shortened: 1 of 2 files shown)


nwoodward (approved):
:+1: Tested and confirmed it works. The column number now matches the span.

alanorth:
@AlexanderS pardon my ignorance. I applied the patch but I still see a seemingly mismatched table in the Edit Item → Item Bitstreams screen:

Screenshot 2021-07-24 at 16-51-54 Item Bitstreams-fs8

Is this what the patch attempts to fix? Thank you.

tdonohue:
@AlexanderS or @nwoodward : Looks like more recent tests by @alanorth show this not to be working correctly.  Any ideas of what might be going on & whether this needs updating?

nwoodward:
I'm not sure. I recall when I tested this it got rid of a superfluous column in the table. @alanorth When I look at your screenshot I can't tell how the table columns are mismatched.

alanorth:
@nwoodward this was awhile ago, but looking at the screenshot again it does look like some table columns are messed up. Circled here:

126871108-46fc4dd3-6087-4a27-8cfb-990e4f6d08af

nwoodward:
Ah. Those buttons are enclosed in <p> tags, which is why there's a line break in between them. I had to inspect this page in Developer Tools and look at the HTML to see that the colspan didn't match the number of columns before the PR.

PR #2414, opened 2019-04-29
feedback on configurable entities

benbosman:
feedback on configurable entities

### DIFF
--- a/dspace-api/src/main/java/org/dspace/content/ItemServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/content/ItemServiceImpl.java
@@ -1450,7 +1450,7 @@ prevent the generation of resource policy entry values with null dspace_object a
                                                                                     relationship.getID(), place));
         }
         RelationshipMetadataValue relationMetadataFromOtherItem =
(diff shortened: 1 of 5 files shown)


dspace-api/src/main/java/org/dspace/content/ItemServiceImpl.java
```diff
@@ -1450,7 +1450,7 @@ public boolean canCreateNewVersion(Context context, Item item) throws SQLExcepti
                                                                                     relationship.getID(), place));
         }
         RelationshipMetadataValue relationMetadataFromOtherItem =
-            getRelationMetadataFromOtherItem(context, otherItem, relationName, relationship.getID());
+            getRelationMetadataFromOtherItem(context, otherItem, relationName, relationship.getID(), place);
```
abollini: I haven't found feedback about that on the linked PR so I cannot judge about it

abollini (approved):
except for the changes noted inline on which I haven't an opinion, the other changes reflect some of the changes that I asked for. Thanks!

paulo-graca (commented):
Just wanted to give my 👍

PR #2421, opened 2019-05-06
[DS-1443] Export structure with structure-builder

mwoodiupui:
https://jira.duraspace.org/browse/DS-1443
Recreate #1257.

Export the current Community/Collection structure in a form suitable for re-importation with the command-line structure-builder.


tdonohue (approved):
👍 Looks good to me. Thanks for the ITs!

PR #2430, opened 2019-05-16
DS-4253 Avoid oom during discovery reindex

cwilper:
Tested locally, this resolves https://jira.duraspace.org/browse/DS-4253

### DIFF
--- a/dspace-api/src/main/java/org/dspace/discovery/SolrServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/discovery/SolrServiceImpl.java
@@ -34,8 +34,6 @@ import java.util.TimeZone;
 import java.util.UUID;
 import java.util.Vector;
 
-import com.google.common.collect.ImmutableList;
-
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.collections4.MapUtils;
 import org.apache.commons.collections4.Transformer;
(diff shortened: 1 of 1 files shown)


benbosman (approved):
I haven't tested this on my machine, but what I think is:
* ImmutableList.copyOf will indeed cause memory issues
* Using the normal iterator similar to the code prior to the MyDSpace merge will solve this issue
* A race condition may occur with the normal iterator, but I don't think it's relevant to catch here because the indexing process doesn't create/delete items

So this looks good to merge for me (without having deployed and tested it anywhere)

abollini (approved):
it seems quite obvious to me by code inspection and it is also aligned with the previous existing code, so +1

tdonohue (approved):
+1. Agreed this is obvious.  Thanks @cwilper for catching this quickly!

mwoodiupui:
Making a `List` out of the content of an `Iterator` is clearly not a good idea unless we *know* that the content is not (potentially) huge.  But it is unfortunate that we have `Iterator`s over potentially long content streams which do not also implement `Iterable`.  We are stuck with the uncomfortable, error-prone `Iterator` interface when looping instead of writing a simple `for` loop and letting the compiler do the Right Thing.  It's easy to implement `Iterable` if `Iterator` is also implemented:  usually `public Iterator<Something> iterator() { return this; }` will do.

PR #2432, opened 2019-05-19
DS-4252 Discovery re-index doesn't work without -b

abollini:
This is a fix for the issue described here
https://jira.duraspace.org/browse/DS-4252

The resource_type and resource_id are now required to retrieve objects from a solr document and the clean method was not updated to reflect that. I have also improved the javadoc to make this more evident


benbosman:
Although the cleanIndex(boolean force) method is not modified in this PR, it seems to be part of the reason why it's failing when the "-f" option is used.
The deleteByQuery in that method builds an AND query instead of an OR query:
```
getSolr().deleteByQuery(
                    "search.resourcetype:[" + Constants.ITEM + " TO " + Constants.COMMUNITY + "]" +
                    " AND " +
                    "search.resourcetype:[" + Constants.WORKSPACEITEM + " TO " + Constants.CLAIMEDTASK + "]");
```

The cleanIndex(boolean force, int type) has indeed been improved now to remove the items without the "-f" option, but it's only called with force set to false (so the `if(force)` is never executed).

abollini:
Hi @artlowel , thanks for the review. You are right about the cleanIndex method so I have included the fix for the query in the case both the -b and -f options are used

KevinVdV (approved):
Made a review of the code & tested this locally, it works as expected now, +1 from me.

tdonohue (approved):
👍 Looks good to me too. I didn't test, but the code looks fine, and I trust @KevinVdV's tests.

PR #2437, opened 2019-05-24
[DS-4259] HAL login form does not support special characters

AlexanderS:
This fixes the login with special characters in the username or password. The issue can be reproduced on the 4science demo with the official demo accounts. The "+" sign in the username is a problem and will be replaced with a space in the request.

The second commit fixes a little usability issue and uses a submit input field as login button. This allows the browsers to automatically submit the form, if you press the return key.

https://jira.duraspace.org/browse/DS-4259

### DIFF
--- a/dspace-spring-rest/src/main/webapp/login.html
+++ b/dspace-spring-rest/src/main/webapp/login.html
@@ -55,11 +55,11 @@
 </head>
 <body>
 <div class="container">
-    <form class="form-signin">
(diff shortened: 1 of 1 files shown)


abollini (approved):
Thanks! tested it works.
To simplify the conduction of the workshop at OR2019 where we will demo the HAL browser, I have already manually applied the fix to the official demo

tdonohue (approved):
👍

PR #2438, opened 2019-05-25
DS-4260: Remove non-existing command from oai harvester's help

pnbecker:
https://jira.duraspace.org/browse/DS-4260


tdonohue (approved):
👍 Tiny, harmless fix. Looks good. Thanks!

PR #2452, opened 2019-06-20
[DS-4032] RequestItem.isAccept_request() can NPE if setAccept_request() was never called.

mwoodiupui:
https://jira.duraspace.org/browse/DS-4032
Ported #2224 to v6.


tdonohue (approved):
👍 Exact copy of #2224.  This is a simple backport. Merging immediately.

PR #2457, opened 2019-06-27
Delete relationship Integration Test

benbosman:
The delete relationship REST API is used in some of the ITs, but only indirectly verified.
This is an explicit IT which verifies the delete relationship REST API
It will create relationships and delete them again to verify they are no longer present


AlexanderS (approved):
:+1: Code looks good and the test is verifying, that the deletion of relationships works.

tdonohue (approved):
👍 Looks good to me.  Thanks!

PR #2463, opened 2019-07-03
DS-4291 Properly show results for 0-9 link in browse

aprilherron:
When browsing by Author, if you click the "0-9" link at the top, it is only showing results for 0 instead of 0-9.

Currently, only a facet query prefix for "0" is being used, so no other results are shown. Because facet query prefixes don't support ranges, the simplest solution is to perform separate queries and combine the results for 0-9. This fix performs the additional queries and combines their results and totals.

This is the 6.x version of #2696

### DIFF
--- a/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/artifactbrowser/ConfigurableBrowse.java
+++ b/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/artifactbrowser/ConfigurableBrowse.java
@@ -24,9 +24,11 @@ import org.dspace.app.xmlui.utils.*;
 import org.dspace.app.xmlui.wing.Message;
 import org.dspace.app.xmlui.wing.WingException;
 import org.dspace.app.xmlui.wing.element.*;
+import org.dspace.app.xmlui.wing.element.List;
 import org.dspace.authorize.AuthorizeException;
 import org.dspace.browse.*;
 import org.dspace.content.*;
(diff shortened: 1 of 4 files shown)


kshepherd (approved):
+1 by inspection, looks good

dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/artifactbrowser/ConfigurableBrowse.java
```diff
@@ -856,7 +855,21 @@ private BrowseInfo getBrowseInfo() throws SQLException, UIException
         {
             // Create a new browse engine, and perform the browse
             BrowseEngine be = new BrowseEngine(context);
-            this.browseInfo = be.browse(params.scope);
+            if ("0".equals(params.scope.getStartsWith())) {
+                BrowseInfo info = be.browse(params.scope);
+                for (String num : new String[]{"1","2","3","4","5","6","7","8","9"}) {
+                    // Append our results to info and upate total
```
nwoodward: upate => update

nwoodward (approved):
:+1: Tested and it works.

KingKrimmson:
My previous implementations was missing some fixes we had made for a client. This new implementation supports both asc/desc orders and reports proper counts as well

nwoodward (approved):
:+1: Tested the new implementation, and it works, including ordering and counts.

PR #2464, opened 2019-07-03
DS-4292  Fix Browse pagination with starts-with

aprilherron:
Currently, if you're browsing by Author, select a starts-with link and then click the next page, it brings you to the 2nd page of all results rather than the 2nd page of those _starts_with results.

This fix adds the starts-with parameter to `getControlParameters()` so that pagination forward _and_ backwards work with starts-with

### DIFF
--- a/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/artifactbrowser/ConfigurableBrowse.java
+++ b/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/artifactbrowser/ConfigurableBrowse.java
@@ -1099,6 +1099,7 @@ class BrowseParams
         paramMap.put(BrowseParams.RESULTS_PER_PAGE, Integer
                 .toString(this.scope.getResultsPerPage()));
         paramMap.put(BrowseParams.ETAL, Integer.toString(this.etAl));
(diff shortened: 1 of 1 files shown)


kshepherd:
@KingKrimmson this has a conflict that makes me think this issue was already fixed (along with fixing some other starts by issues) in #1730 - if that is not the case, please let me know and we can look at rebasing/resolving?

J4bbi:
I couldn't replicate so I suggest we close @kshepherd

ssolim:
> 
> 
> I couldn't replicate so I suggest we close @kshepherd

We have found this bug on our live. You can see it here: 

https://edoc.hu-berlin.de/browse?rpp=20&sort_by=-1&type=author&etal=-1&starts_with=P&order=ASC

I didnt test this fix, but i can confirm its still existing. The fix #1730 is applied so i assume it must be still there.

kshepherd:
@ssolim I just tested on the link, and other starts-with browsing on your site and it looks ok to me.. clicking next page does keep me within that browse scope, not the full browse results. Am I doing something wrong with testing?

ssolim:
Yeah you are right. I mixed this up. My fault!!!
https://jira.lyrasis.org/browse/DS-3996
This is what is i meant and what you can see at the link. I am sorry for the confusion. Somehow i thought this is related.

But this seems also fixed to me. Sorry for the confusion.

PR #2465, opened 2019-07-03
DS-4293 Fix Known/Supported labels in UploadStep/UploadWithEmbargoStep

aprilherron:
Currently in UploadStep/UploadWithEmbargoStep, "Supported" file types are mislabeled as "Known" and "Known" file types are mislabeled as "Supported". Support levels are shown correctly in the Review Step.

This fix fixes the switch statement to choose the correct labels (and use enum values)

### DIFF
--- a/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/submission/submit/UploadStep.java
+++ b/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/submission/submit/UploadStep.java
@@ -332,13 +332,13 @@ public class UploadStep extends AbstractSubmissionStep
                     cell.addContent(" ");
                     switch (support)
                     {
-                        case 1:
-                            cell.addContent(T_supported);
-                            break;
-                        case 2:
+                        case BitstreamFormat.KNOWN:
                             cell.addContent(T_known);
                             break;
-                        case 3:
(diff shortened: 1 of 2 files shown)


nwoodward (approved):

alanorth:
I don't understand the difference between "known" and "supported" in the upload step, but this pull request does address the issue. Before, when uploading a PDF it would say "known", but now it says "supported". So looks good to me? :+1:

Nitpick: this patch adds unnecessary whitespace...

J4bbi (approved):

J4bbi:
Thanks for the PR @KingKrimmson

mwoodiupui:
I think these labels were just carried over from the original MIT instance where they had locally-defined meanings.  But I've always taken them to mean:

- known:  repository staff recognize this format and can help with it.
- supported:  we undertake to maintain the content of bitstreams in this format if it becomes obsolescent.  We will make a best effort to transform the content to a similar and better-supported format if necessary.

But it's really left to each instance to define support-level labels for its users.

J4bbi:
@mwoodiupui , was I too hasty? should we revert?

mwoodiupui:
@J4bbi No, I was just responding to @alanorth 's comment on the difference between "known" and "supported".  Clearly the new code is more correct.

PR #2466, opened 2019-07-03
[DS-1443] Export Community/Collection structure with structure-builder

mwoodiupui:
Somehow Travis didn't catch a conflicting patch, so fix it.


mwoodiupui:
Merged without review to un-break master.  Sorry about the mess.

PR #2479, opened 2019-08-06
[DS-4110] fix issue in legacy id cleanup of stats records

mwoodiupui:
https://jira.duraspace.org/browse/DS-4110
Port #2292 to master.


tdonohue (approved):
👍 Looks good to me.  This is an exact copy of #2292, which was well tested & reviewed.  Merging

PR #2480, opened 2019-08-06
[DS-4321] master: docker image dspace/dspace-cli:dspace-7_x

terrywbrady:
Create a container for running DSpace CLI tasks.
- https://jira.duraspace.org/browse/DS-4321
- https://wiki.duraspace.org/display/DSPACE/DSpace+Docker+and+Cloud+Deployment+Goals
- https://github.com/DSpace-Labs/DSpace-Docker-Images/pull/140
- https://github.com/DSpace-Labs/DSpace-Docker-Images/pull/142

### DIFF
--- a/Dockerfile.cli.jdk8
+++ b/Dockerfile.cli.jdk8
@@ -25,7 +25,7 @@ ADD --chown=dspace . /app/
 COPY dspace/src/main/docker/local.cfg /app/local.cfg
 
 # Build DSpace.  Copy the dspace-install directory to /install.  Clean up the build to keep the docker image small
-RUN mvn package -P'!dspace-rest,!dspace-server-webapp,!dspace-rdf,!dspace-sword,!dspace-swordv2,!dspace-oai' && \
+RUN mvn package -P'!dspace-rest' && \
   mv /app/dspace/target/${TARGET_DIR}/* /install && \
   mvn clean
 
--- a/Dockerfile.cli.jdk8
+++ b/Dockerfile.cli.jdk8
@@ -3,7 +3,7 @@
 #
 # This version is JDK8 compatible
 # - openjdk:8-jdk
-# - ANT 1.10.5
+# - ANT 1.10.7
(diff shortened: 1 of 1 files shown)


Dockerfile.cli.jdk8
```diff
@@ -0,0 +1,54 @@
...
+    && chown -Rv dspace: /app
+
+USER dspace
+
+# Copy the DSpace source code into the workdir (excluding .dockerignore contents)
+ADD --chown=dspace . /app/
+COPY dspace/src/main/docker/local.cfg /app/local.cfg
+
+# Build DSpace.  Copy the dspace-install directory to /install.  Clean up the build to keep the docker image small
+RUN mvn package && \
```
kshepherd: as with the 6.x PR, though we have different profiles now with mvn in 7+, can we exclude the webapp to keep compile time faster since we shouldn't need it for the special CLI image?

Dockerfile.cli.jdk8
```diff
@@ -0,0 +1,54 @@
...
+ENV ANT_VERSION 1.10.5
+ENV ANT_HOME /tmp/ant-$ANT_VERSION
+ENV PATH $ANT_HOME/bin:$PATH
+
+RUN mkdir $ANT_HOME && \
+    wget -qO- "https://www.apache.org/dist/ant/binaries/apache-ant-$ANT_VERSION-bin.tar.gz" | tar -zx --strip-components=1 -C $ANT_HOME
+
+RUN ant init_installation update_configs update_code
+
+# Step 3 - Run tomcat
```
pnbecker: Same as for DSpace 6: we do not run tomcat.

pnbecker (approved):
Not tested. Looks good to me.

kshepherd (approved):
+1, changes are working - will be good to find out how to exclude the unnecessary (webapp) profiles from the mvn build in DSpace 7. Presumably the single webapp is creating some extra dependencies between the core project and other modules?

PR #2482, opened 2019-08-06
Travis CI: Continue to use Ubuntu Trusty 14.04

mwoodiupui:
Porting #2477 to 5_x.


tdonohue (approved):
👍 Looks good! Obviously we should have been pinning this OS version a long time ago

PR #2485, opened 2019-08-08
[DS-3422] Remove XMLUI and JSPUI from the official distribution

mwoodiupui:
https://jira.duraspace.org/browse/DS-3422

A few more no-longer-used files removed.


tdonohue (approved):
👍 Obvious changes.  Merging immediately as these 3 files were specific to XMLUI

PR #2495, opened 2019-08-23
[DS-3695] Upgrade to Solr 7:  support sharded statistics simply

mwoodiupui:

This patch takes the simple approach to sharding support:  keep our "manual" sharding code in DSpace and rework it to query Solr for existing shards, rather than assume that we know where Solr's files are.  (Solr might not even be running on the same host, so its index storage might not be directly accessible.)

As I write this, the patch includes commits from #2393 which supports upgrading DSpace to v7 from v1-v6.  These should drop off when that PR is merged.  New work begins with commit 8878b3a.


cwilper (approved):
@mwoodiupui this looks good to me. The main code change that this adds (beyond the common solr fixes that are already part of the other PR) is in how sharded statistics are discovered and loaded. Code looks right and it built fine for me. I tested by manually creating two new cores, named statistics-YYYY in solr, and then exercising this code by requesting a bitstream's content from REST. This triggered the method that lazy-inits the sharded stats core, and I confirmed it correctly discovered and loaded them.

terrywbrady (approved):
Untested, but the changes look good to me.

PR #2499, opened 2019-09-03
Removal of filtered-discovery-page in REST

benbosman:
In the initial entities implementation, a REST feature for filtered-discovery-page has been created.
This was later-on noticed as a feature which is not necessary.

This PR removes this functionality


paulo-graca (approved):
I've done some tests and all seems to work properly. Thank you @benbosman !

mwoodiupui (approved):
+1 by inspection.  Seems harmless.  I could find no references to these classes elsewhere.

PR #2501, opened 2019-09-04
[DS-3849] REST API items resource returns items in non-deterministic order (further 6.x changes)

PhilipVis:
Continuation of the work in https://github.com/DSpace/DSpace/pull/2460.

This PR updates the ItemDAOImpl class to make sure all item lists returned by this class will now be in a deterministic order.

### DIFF
--- a/dspace-api/src/main/java/org/dspace/content/dao/impl/ItemDAOImpl.java
+++ b/dspace-api/src/main/java/org/dspace/content/dao/impl/ItemDAOImpl.java
@@ -19,6 +19,7 @@ import org.dspace.eperson.EPerson;
 import org.hibernate.Criteria;
 import org.hibernate.Query;
 import org.hibernate.criterion.DetachedCriteria;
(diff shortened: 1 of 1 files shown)


tdonohue (approved):
👍  Changes seem fine to me.  However, both this PR and the original #2460 PR *must* also be applied to `master`.  Otherwise, these fixes will not be carried forward (which is obviously not ideal).  Please also create a PR to apply both of these to our `master` branch. Thanks!

kshepherd (approved):
+1
Merging this so it makes it in for 6.4 - the master port #2515 is still failing some tests (seems to be seeing Public Item 3 where it expects Public Item 2 -- and I assume the fault is with the tests?) but this code looks good and behaves well in 6.x. Thanks @PhilipVis

kshepherd:
Actually - first, I'm just going to Close and re-open this PR so we can trigger the Travis CI build again which seemed to stall, and is too old for me to find in the Travis build list

PR #2502, opened 2019-09-05
Endpoints to collect Statistics

Raf-atmire:
We've created two endpoints on the REST API that will allow us to create view events and search events through the REST API.
These endpoints are `/api/statistics/viewevents` and `/api/statistics/searchevents` respectively. These have been configured to only allow POST requests to be handled and will block off any GET requests. These endpoints will make it possible to store those events in SOLR similar to how it was handled in DSpace 6.
The REST Contract for these endpoints can be found here: https://github.com/DSpace/Rest7Contract/blob/master/statistics.md

An additional endpoint has been added namely Statistics under `/api/statistics` which will serve as an intermediate page between the root of the rest api and the event links. This endpoint will serve the purpose of containing the links to the event endpoints.

This PR contains both the implementation for this feature and integration tests to go with it.


dspace-server-webapp/src/main/java/org/dspace/app/rest/repository/SearchEventRestRepository.java
```diff
@@ -0,0 +1,90 @@
...
+    private EventService eventService;
+
+    @Autowired
+    private SearchEventConverter searchEventConverter;
+
+    public SearchEventRest createSearchEvent() throws AuthorizeException, SQLException {
+
+        Context context = obtainContext();
+        HttpServletRequest req = getRequestService().getCurrentRequest().getHttpServletRequest();
+        List<SearchFilter> result = new LinkedList<>();
```
mwoodiupui: Unused?

dspace-api/src/main/java/org/dspace/usage/RestUsageSearchEvent.java
```diff
@@ -0,0 +1,208 @@
...
+ */
+package org.dspace.usage;
+
+import java.util.List;
+import javax.servlet.http.HttpServletRequest;
+
+import org.dspace.content.DSpaceObject;
+import org.dspace.core.Context;
+
+public class RestUsageSearchEvent extends UsageEvent {
```
benbosman: This class is missing JavaDoc, but it can probably be removed

dspace-api/src/main/java/org/dspace/statistics/SolrLoggerUsageEventListener.java
```diff
@@ -53,14 +57,27 @@ public void receiveEvent(Event event) {
...
                 } else if (UsageEvent.Action.SEARCH == ue.getAction()) {
-                    UsageSearchEvent usageSearchEvent = (UsageSearchEvent) ue;
-                    //Only log if the user has already filled in a query !
-                    if (!CollectionUtils.isEmpty(((UsageSearchEvent) ue).getQueries())) {
-                        solrLoggerService.postSearch(ue.getObject(), ue.getRequest(), currentUser,
-                                                     usageSearchEvent.getQueries(), usageSearchEvent.getRpp(),
-                                                     usageSearchEvent.getSortBy(),
-                                                     usageSearchEvent.getSortOrder(), usageSearchEvent.getPage(),
-                                                     usageSearchEvent.getScope());
+                    if (ue instanceof UsageSearchEvent) {
```
benbosman: This seems to be an instanceof implementation because the old UsageSearchEvent can contain multiple queries (which is weird). I would recommend to use UsageSearchEvent instead, remove the instanceof and fix UsageSearchEvent to contain a single query

dspace-server-webapp/src/main/java/org/dspace/app/rest/model/PageRest.java
```diff
@@ -0,0 +1,58 @@
+/**
+ * The contents of this file are subject to the license and copyright
+ * detailed in the LICENSE and NOTICE files at the root of the source
+ * tree and available online at
+ *
+ * http://www.dspace.org/license/
+ */
+package org.dspace.app.rest.model;
+
+public class PageRest {
```
benbosman: This class is missing JavaDoc

dspace-server-webapp/src/main/java/org/dspace/app/rest/model/SearchEventRest.java
```diff
@@ -0,0 +1,110 @@
...
+ */
+package org.dspace.app.rest.model;
+
+import java.util.List;
+import java.util.UUID;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.dspace.app.rest.RestResourceController;
+
+public class SearchEventRest extends BaseObjectRest<UUID> {
```
benbosman: This class is missing JavaDoc

dspace-server-webapp/src/main/java/org/dspace/app/rest/model/ViewEventRest.java
```diff
@@ -0,0 +1,51 @@
...
+ * http://www.dspace.org/license/
+ */
+package org.dspace.app.rest.model;
+
+import java.util.UUID;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.dspace.app.rest.RestResourceController;
+
+public class ViewEventRest extends BaseObjectRest<UUID> {
```
benbosman: This class is missing JavaDoc

benbosman (changes requested):
I was able to test the code locally without any problems.

I've verified:
* creating view events of bitstream, item, collection, community, site
* creating a view event while authenticated
* creating a search event

For each of the use cases, solr contained the relevant statistics (after the commit)

What I am missing in this PR:

Can you extend ViewEventRestRepositoryIT to also test creating a view event of (I verified they work, but is missing ITs):
* bitstream
* collection
* community
* site
* authenticated user (for one of the types)

Is it possible to verify the view event is in fact created (e.g. using a mock Usage Event Listener or the TabFileUsageEventListener) apart from the return status? I was able to verify it worked, but it would be interesting to automate the verification.

I noticed a few file were only changed with whitespace changes. It's not a large amount so it was still readable, but it would be nicer to exclude them

I've also included some inline feedback

dspace-server-webapp/src/main/java/org/dspace/app/rest/repository/ViewEventRestRepository.java
```diff
@@ -0,0 +1,68 @@
...
+        ObjectMapper mapper = new ObjectMapper();
+        ViewEventRest viewEventRest = null;
+        try {
+            ServletInputStream input = req.getInputStream();
+            viewEventRest = mapper.readValue(input, ViewEventRest.class);
+        } catch (IOException e1) {
+            throw new UnprocessableEntityException("Error parsing request body", e1);
+        }
+        if (viewEventRest.getTargetId() == null || StringUtils.isBlank(viewEventRest.getTargetType()) ||
+            !Arrays.asList(Constants.typeText).contains(viewEventRest.getTargetType().toUpperCase())) {
```
mwoodiupui: We will be creating this array quite often, from constant input.  Perhaps it should be done once, statically?

mwoodiupui (approved):
+1 by inspection

benbosman (approved):
All my feedback has been fixed.
I did test it locally, and it worked correctly.

This looks good to me, thanks

PR #2505, opened 2019-09-05
[DS-4336] Update ant version in Docker Build (6x)

terrywbrady:
https://jira.duraspace.org/browse/DS-4336

PR #2507, opened 2019-09-05
[DS-4336] Update ant version in Docker Build (4x) 

terrywbrady:
https://jira.duraspace.org/browse/DS-4336


tdonohue (approved):
👍 Tiny update. Looks good to me.  Port of #2505 (which was already approved/tested & merged)

PR #2508, opened 2019-09-05
[DS-4338] Remove final smidgeon of Solr source code.

mwoodiupui:
https://jira.duraspace.org/browse/DS-4338
Instead of keeping a copy of others' code that we have to maintain, just get it from a dependency.


tdonohue (approved):
👍 Looks right to me. I agree this source file should be removed, and the dependency tweaks all seem reasonable.

kshepherd (approved):
+1 merging now, thanks!

PR #2513, opened 2019-09-09
DS-4340 Duplicate Headers when bitstream has a comma in the title

abollini:
Fixes #7680
The original issue DS-1422 (#4791) was fixed only in the XMLUI


tdonohue (approved):
👍 Looks good to me. Haven't tested, but fix seems obvious

kshepherd (approved):
+1 thanks @abollini !

PR #2516, opened 2019-09-12
DS-4342 improve performance when retrieving items from a collection

PhilipVis:
Fixes #7682
This PR updates the CollectionsResource.getCollectionItems method to include the limit and offset parameters when retrieving items for a collection, instead of retrieving al items for a collection.

This should improve the performance of the collections/collection_id/items REST endpoint.


lap82:
@PhilipVis see my message reported at #2517

tdonohue (commented):
👍 Code looks fine to me. I didn't test, but trust @lap82's tests.  This may also need to be ported to `master`.

J4bbi:
just a note that #2275 would enable itemservice.findAuthorizedByCollection to be created.

PR #2517, opened 2019-09-16
DS-4342 improve performance when retrieving items from a collection

PhilipVis:
Fixes #7682
This PR updates the CollectionsResource.getCollectionItems method to include the limit and offset parameters when retrieving items for a collection, instead of retrieving al items for a collection.

This should improve the performance of the collections/collection_id/items REST endpoint.


lap82:
@PhilipVis The works looks good. Thank you. But I note the commit message is not quite usefull, please can you amend the commit inserting an usefull description?  My suggestion is "[DS-4342] improve the performance of the collections/collection_id/items REST endpoint"

tdonohue (commented):
👍 Code looks fine to me too. Didn't test, but I trust @lap82's tests.  This may need to be ported to `6_x` and `master` branches.

PR #2518, opened 2019-09-17
[DS-4144] Update node and npm versions for mirage 2 (5.x)

kshepherd:
JIRA: https://jira.duraspace.org/browse/DS-4144
This small change brings node and npm versions up to date in DSpace 5.x
This fixes an issue where mirage2 is failing to build in some versions of DSpace 5.


tdonohue (approved):
👍 Haven't tested, but this is a simple version update. Seems fine to me (assuming others have tested this and found it to work)

terrywbrady (approved):
Untested, but looks good

mwoodiupui (approved):
Wow, the range of those changes is breath-taking, but all the more reason to do it if we are *that* far behind.

PR #2519, opened 2019-09-17
[DS-4144] Update node and npm versions for mirage 2 (6.x)

kshepherd:
JIRA: https://jira.duraspace.org/browse/DS-4144
This small change brings node and npm versions up to date as per the fix for DSpace 5.x -- the 6.x mirage2 builds aren't actually failing currently, but we might as well bring the versions up to date and get them matched up.


tdonohue (approved):
👍 Haven't tested, but this is a simple version update. Seems fine to me.

terrywbrady (approved):
Untested, but change looks good

mwoodiupui (approved):

PR #2520, opened 2019-09-17
[DS-3849] REST API items resource returns items in non-deterministic order (5.x)

PhilipVis:
This PR updates methods in the Collection and Item classes to make sure all item lists returned will now be in a deterministic order.


kshepherd:
thanks @PhilipVis -- will this PR bring 5.x into line with 6.x in terms of the REST ordering, or did the #2501 PR bring in some further changes that also need porting back to 5.x?

PR #2521, opened 2019-09-17
Fix config v1

ryscher:


ryscher:
oops, sorry for the crazy PR to the wrong repository

PR #2524, opened 2019-09-23
[DS-4347] remove solr from ant step in docker build

terrywbrady:
https://jira.duraspace.org/browse/DS-4347


kshepherd (approved):
+1 untested but easy and obvious change since this task is gone now

PR #2543, opened 2019-10-10
DS-4271: Replaced brackets with double quotes in SolrServiceImpl.

### LINKED ISSUE #7611: [DS-4271] Discovery "contains" filter does not perform a phrase query
<i>Imported from JIRA <a href="https://jira.lyrasis.org/browse/DS-4271">[DS-4271]</a> created by cwilper</i>
<p>When doing a "contains" filter query, if the value entered for the field contains spaces, the behavior is to return documents where the individual words are present in any order the field.</p>

<p>A typical user would expect "contains" to do a phrase query, where the entire phrase occurs somewhere in the field's value.</p>

<p>This is a long-standing issue that stems from the following line in SolrServiceImpl, which has remained untouched since the original Discovery implementation, and is present in versions 4.x through 7.0 Preview 1:</p>

<p><a href="https://github.com/DSpace/DSpace/blob/dspace-6_x/dspace-api/src/main/java/org/dspace/discovery/SolrServiceImpl.java#L2110" class="external-link" target="_blank" rel="nofollow noopener">https://github.com/DSpace/DSpace/blob/dspace-6_x/dspace-api/src/main/java/org/dspace/discovery/SolrServiceImpl.java#L2110</a></p>

<p>Notice how it surrounds the text sent to solr with "(" and ")". <em>This has specific meaning in solr, which differs from surrounding the text with double quotes.</em> Surrounding with double quotes would have the expected effect of doing a phrase query. Surrounding with parentheses is equivalent to entering multiple fielded search tearms, one for each word. For example, a the solr query:</p>

<p><tt>title:(introduction to dspace)</tt></p>

<p>...is the same as the solr query:</p>

<p><tt>title:introduction title:to title:dspace</tt></p>

<p>...which is not the expected effect.</p>

<p>The solution, which we have tested, is to replace the surrounding parentheses in the code referenced above, with double quotes.</p>

jonas-atmire:
DSpace 6 versions of the fix provided in:
https://jira.duraspace.org/browse/DS-4271 
Fixes #7611 

Quite a small fix, but it has been open for a little while now, so took the liberty to create a quick PR
(More information is present in the JIRA ticket itself)

### DIFF
--- a/dspace-api/src/main/java/org/dspace/discovery/SolrServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/discovery/SolrServiceImpl.java
@@ -2107,7 +2107,7 @@ public class SolrServiceImpl implements SearchService, IndexingService {
                 if(!value.matches("\\[.*TO.*\\]"))
                 {
                     value = ClientUtils.escapeQueryChars(value);
-                    filterQuery.append("(").append(value).append(")");
+                    filterQuery.append("\"").append(value).append("\"");
                 }
                 else
                 {


AlexanderS (approved):
👍 Looks good.

nwoodward (approved):

alanorth:
Agree this is a good fix for a problem I didn't even realize we had. Tested with a _contains_ filter with spaces on a subject term in our staging repository.

Before, 293 results, of which many are not what the user would expect:

ds-4271-before-fs8

After, 162 results, of which all are exactly matching the search term:

ds-4271-after-fs8

Looks good to me for DSpace 6.4. Merge!

J4bbi:
I haven't tested this but it's got +3 @kshepherd

alanorth:
@J4bbi I think this is a good one for 6.4. It's at +3 for some months now.

alanorth (approved):

alanorth:
Ported to `main` for DSpace 7.x here: https://github.com/DSpace/DSpace/pull/8284

PR #2545, opened 2019-10-10
[DS-4359] added nullcheck in EntityServiceImpl#getAllRelationshipTypes

Raf-atmire:
Fixed the issue described at: https://jira.duraspace.org/browse/DS-4359

This PR simply adds a nullcheck around the EntityType and returns an empty list if it's null. This solves the issue as a nullcheck was needed.
This does not cause any issue since it's not possible to get a RelationshipType that is defined by a null EntityType; therefor it'll always need to return an empty list if the EntityType is null


benbosman (approved):
This seems to be a straightforward catch to avoid exceptions

PR #2546, opened 2019-10-10
[DS-4360] fixed construction of the SearchEvents link on the statistics endpoint

Raf-atmire:
This PR fixes the issue described in: https://jira.duraspace.org/browse/DS-4360

The construction of the SearchEvents link has been altered to the correct method call


benbosman (approved):
Seems very straightforward

PR #2550, opened 2019-10-13
Patch for DS-4362

### LINKED ISSUE #7701: [DS-4362] Feedback link breaks on sites without sub-domain
<i>Imported from JIRA <a href="https://jira.lyrasis.org/browse/DS-4362">[DS-4362]</a> created by saiful</i>
<p>On the DSpace 6.4-snapshot the feedback page breaks if the <em>dspace.hostname</em> in config has a single dot (say 'example.com'). The error in the stacktrace is:</p>
<pre class="code-java">
Java stacktrace: java.lang.StringIndexOutOfBoundsException: <span class="code-object">String</span> index out of range: -1
 at java.lang.<span class="code-object">String</span>.substring(<span class="code-object">String</span>.java:1927)
 at org.dspace.app.xmlui.aspect.artifactbrowser.SendFeedbackAction.act(SendFeedbackAction.java:67)</pre>
</div></div>
<p>The line 67 of <em>SendFeedbackAction.java</em> tries to extract the sub-domain even when it doesn't exist and it breaks the page.</p>

saiful-semantic:
Fixes breaking of feedback link on sites without sub-domain.
https://jira.duraspace.org/browse/DS-4362

Fixes #7701

### DIFF
--- a/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/artifactbrowser/SendFeedbackAction.java
+++ b/dspace-xmlui/src/main/java/org/dspace/app/xmlui/aspect/artifactbrowser/SendFeedbackAction.java
@@ -64,7 +64,8 @@ public class SendFeedbackAction extends AbstractAction
             // cut off all but the hostname, to cover cases where more than one URL
             // arrives at the installation; e.g. presence or absence of "www"
             int lastDot = host.lastIndexOf('.');
-            basicHost = host.substring(host.substring(0, lastDot).lastIndexOf('.'));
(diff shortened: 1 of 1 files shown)


nwoodward (changes requested):
Please add braces to the if statement.

nwoodward (approved):
:+1: Thanks! Tested and confirmed it works.

tdonohue (approved):
👍 Looks good to me. Already approved...and passes recent build.  Thanks @saiful-semantic !

PR #2551, opened 2019-10-15
Entities bug-fixes in place management

benbosman:
We've noticed a few use cases where the place of a new relationship is not defined correctly. These problems have been solved and JUnit tests have been created to automatically verify the functionality.

The previously existing tests didn't cause any problems because:
* All tests added the relationship as the last element
* There were no tests to change the place of a relationship
* The item.update was always called somewhere implicitly to fix the places

All of the previous tests still succeed, and the additional tests succeed as well.
Inline comments have been included to explain the more complicated reasons for the approach in determining the place
The JavaDocs have been updated as well

### DIFF
--- a/dspace-api/src/main/java/org/dspace/app/bulkedit/MetadataImport.java
+++ b/dspace-api/src/main/java/org/dspace/app/bulkedit/MetadataImport.java
@@ -812,8 +812,8 @@ public class MetadataImport {
             leftItem = otherItem;
         }
         RelationshipType relationshipType = acceptedRelationshipType;
-        int leftPlace = relationshipService.findLeftPlaceByLeftItem(c, leftItem) + 1;
-        int rightPlace = relationshipService.findRightPlaceByRightItem(c, rightItem) + 1;
+        int leftPlace = relationshipService.findNextLeftPlaceByLeftItem(c, leftItem) ;
+        int rightPlace = relationshipService.findNextRightPlaceByRightItem(c, rightItem);
         Relationship persistedRelationship = relationshipService.create(c, leftItem, rightItem,
                                                                         relationshipType, leftPlace, rightPlace);
         relationshipService.update(c, persistedRelationship);
--- a/dspace-api/src/main/java/org/dspace/app/bulkedit/MetadataImport.java
+++ b/dspace-api/src/main/java/org/dspace/app/bulkedit/MetadataImport.java
@@ -193,7 +193,7 @@ public class MetadataImport {
      * @param workflowNotify If the workflows should be used, whether to send notifications or not
      * @param useTemplate    Use collection template if create new item
      * @return An array of BulkEditChange elements representing the items that have changed
-     * @throws MetadataImportException if something goes wrong
+     * @throws MetadataImportException  if something goes wrong
      */
     public List<BulkEditChange> runImport(boolean change,
                                           boolean useWorkflow,
--- a/dspace-api/src/main/java/org/dspace/app/bulkedit/MetadataImport.java
+++ b/dspace-api/src/main/java/org/dspace/app/bulkedit/MetadataImport.java
@@ -779,7 +779,7 @@ public class MetadataImport {
 
         // Create the relationship
         int leftPlace = relationshipService.findNextLeftPlaceByLeftItem(c, leftItem);
-        int rightPlace = relationshipService.findNextRightPlaceByRightItem(c, rightItem)1;
+        int rightPlace = relationshipService.findNextRightPlaceByRightItem(c, rightItem);
         Relationship persistedRelationship = relationshipService.create(c, leftItem, rightItem,
                 foundRelationshipType, leftPlace, rightPlace);
         relationshipService.update(c, persistedRelationship);
--- a/dspace-api/src/main/java/org/dspace/content/RelationshipServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/content/RelationshipServiceImpl.java
@@ -82,6 +82,8 @@ public class RelationshipServiceImpl implements RelationshipService {
         if (isRelationshipValidToCreate(context, relationship)) {
             if (authorizeService.authorizeActionBoolean(context, relationship.getLeftItem(), Constants.WRITE) ||
                 authorizeService.authorizeActionBoolean(context, relationship.getRightItem(), Constants.WRITE)) {
+                // This order of execution should be handled in the creation (create, updateplace, update relationship)
+                // for a proper place allocation
                 Relationship relationshipToReturn = relationshipDAO.create(context, relationship);
                 updatePlaceInRelationship(context, relationshipToReturn);
                 update(context, relationshipToReturn);
@@ -100,6 +102,8 @@ public class RelationshipServiceImpl implements RelationshipService {
     public void updatePlaceInRelationship(Context context, Relationship relationship)
         throws SQLException, AuthorizeException {
         Item leftItem = relationship.getLeftItem();
+        // Max value is used to ensure that these will get added to the back of the list and thus receive the highest
+        // (last) place as it's set to a -1 for creation
         if (relationship.getLeftPlace() == -1) {
             relationship.setLeftPlace(Integer.MAX_VALUE);
         }
--- a/dspace-api/src/main/java/org/dspace/content/RelationshipServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/content/RelationshipServiceImpl.java
@@ -73,8 +73,10 @@ public class RelationshipServiceImpl implements RelationshipService {
         if (isRelationshipValidToCreate(context, relationship)) {
             if (authorizeService.authorizeActionBoolean(context, relationship.getLeftItem(), Constants.WRITE) ||
                 authorizeService.authorizeActionBoolean(context, relationship.getRightItem(), Constants.WRITE)) {
-                updatePlaceInRelationship(context, relationship, true);
-                return relationshipDAO.create(context, relationship);
+                Relationship relationshipToReturn = relationshipDAO.create(context, relationship);
+                updatePlaceInRelationship(context, relationshipToReturn);
+                update(context, relationshipToReturn);
+                return relationshipToReturn;
             } else {
                 throw new AuthorizeException(
                     "You do not have write rights on this relationship's items");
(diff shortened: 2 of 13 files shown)


dspace-api/src/main/java/org/dspace/app/bulkedit/MetadataImport.java
```diff
@@ -778,8 +778,8 @@ private void addRelationship(Context c, Item item, String typeName, String value
         }
 
         // Create the relationship
-        int leftPlace = relationshipService.findLeftPlaceByLeftItem(c, leftItem) + 1;
-        int rightPlace = relationshipService.findRightPlaceByRightItem(c, rightItem) + 1;
+        int leftPlace = relationshipService.findNextLeftPlaceByLeftItem(c, leftItem);
```
paulo-graca: The first time I executed an CSV import, I got an error. On the second try, without any changes, the record was registered. But I noticed a duplication of isAuthorOfPublication on the Metadatafield with the relationship.

`metadata_field_id: 134 (relation.isAuthorOfPublication)`
2019-10-29 09_52_48-Valentina Studio

Should a specific relation.isAuthorOfPublication should be present in this table?

benbosman: Hi @paulo-graca,

I was able to reproduce the problem, but also on master. It seems you've found a bug in one of the CSV branches already merged in master.
Can you move this to a separate ticket so it's not lost?

I was able to reproduce this with the CSV https://www.dropbox.com/s/vvugb1tuknjhp9a/test-records-2.csv?dl=1 which created relation.isAuthorOfPublication with value 0ffbee3f-e7ea-42bc-92fe-2fbef1a52c0f twice using master

dspace-api/src/main/java/org/dspace/content/RelationshipServiceImpl.java
```diff
@@ -94,18 +98,32 @@ public Relationship create(Context context, Relationship relationship) throws SQ
...
         throws SQLException, AuthorizeException {
         Item leftItem = relationship.getLeftItem();
+        // Max value is used to ensure that these will get added to the back of the list and thus receive the highest
+        // (last) place as it's set to a -1 for creation
+        if (relationship.getLeftPlace() == -1) {
+            relationship.setLeftPlace(Integer.MAX_VALUE);
+        }
+        Item rightItem = relationship.getRightItem();
+        if (relationship.getRightPlace() == -1) {
+            relationship.setRightPlace(Integer.MAX_VALUE);
```
tdonohue: This method's logic  is increasingly confusing.  Can you better explain why we have to temporarily set the place to `MAX_VALUE`?  Why not just set it to the length of `leftRelationships` or `rightRelationships` on the next lines?

dspace-api/src/main/java/org/dspace/content/RelationshipServiceImpl.java
```diff
@@ -94,18 +98,32 @@ public Relationship create(Context context, Relationship relationship) throws SQ
...
         List<Relationship> rightRelationships = findByItemAndRelationshipType(context,
                                                                               rightItem,
                                                                               relationship.getRelationshipType(),
                                                                               false);
 
+        if (leftRelationships.contains(relationship)) {
+            leftRelationships.remove(relationship);
+        }
+        if (rightRelationships.contains(relationship)) {
+            rightRelationships.remove(relationship);
```
tdonohue: These if statements should be preceded by a comment to describe why we are removing the relationship from the list.  I don't understand the logic here.  Could we add some Unit Tests to prove out this is working as expected?

dspace-api/src/test/java/org/dspace/content/RelationshipServiceImplPlaceTest.java
```diff
@@ -0,0 +1,374 @@
...
+     */
+    @After
+    @Override
+    public void destroy() {
+        context.abort();
+        super.destroy();
+    }
+
+    @Test
+    public void addMetadataAndRelationshipTest() throws Exception {
```
tdonohue: Either break up this method into separate tests, or please add inline comments.  This method is hard to follow as you create new content midway through and then run more tests.  It's difficult for me to see what exactly you are testing for in each test without understanding all 50 lines of this method.

dspace-api/src/test/java/org/dspace/content/RelationshipServiceImplPlaceTest.java
```diff
@@ -0,0 +1,374 @@
...
+        assertMetadataValue(authorQualifier, contributorElement, dcSchema, "test, three", null, 3, list.get(3));
+        assertMetadataValue(authorQualifier, contributorElement, dcSchema, "test, four", null, 4, list.get(4));
+        assertMetadataValue(authorQualifier, contributorElement, dcSchema, "familyNameTwo, firstNameTwo",
+                            "virtual::" + relationshipTwo.getID(), 5, list.get(5));
+        assertThat(relationshipTwo.getLeftPlace(), equalTo(5));
+
+    }
+
+    @Test
+    public void AddMetadataAndRelationshipWithSpecificPlaceTest() throws Exception {
```
tdonohue: Again, method is too long, and you add new content at several stages.  Please break this up into several smaller tests (with more descriptive names)...or add descriptive inline comments

dspace-api/src/test/java/org/dspace/content/RelationshipServiceImplPlaceTest.java
```diff
@@ -0,0 +1,374 @@
...
+        assertMetadataValue(authorQualifier, contributorElement, dcSchema, "test, four", null, 4, list.get(4));
+        assertMetadataValue(authorQualifier, contributorElement, dcSchema, "familyNameTwo, firstNameTwo",
+                            "virtual::" + relationshipTwo.getID(), 5, list.get(5));
+        assertThat(relationshipTwo.getLeftPlace(), equalTo(5));
+
+    }
+
+
+    @Test
+    public void AddAndRemoveMetadataAndRelationshipsTest() throws Exception {
```
tdonohue: Same as above.  Too large and no comments

dspace-api/src/test/java/org/dspace/content/RelationshipServiceImplPlaceTest.java
```diff
@@ -0,0 +1,374 @@
...
+        metadataValueService.delete(context, metadataValueToRemove);
+
+        context.restoreAuthSystemState();
+
+        List<MetadataValue> list = itemService
+            .getMetadata(item, dcSchema, contributorElement, authorQualifier, Item.ANY);
+        assertThat(list.size(), equalTo(2));
+
+        // Check that these places are still intact after the deletion as the place doesn't get updated until an
+        // item.update has been called
```
tdonohue: This is a better inline comment.  But, where do we call `item.update` below to ensure the place *does* get updated after the deletion?

tdonohue (commented):
@benbosman : I gave this a quick review today. I'm confused by several of the code changes here (see below).  As the new test methods are very large an uncommented, it's difficult to follow them as well.  I'd appreciate either inline comments in large test methods, or find a way to break them up into smaller these methods (where plausible).

dspace-api/src/main/java/org/dspace/content/RelationshipServiceImpl.java
benbosman: The place -1 implies it's not set and should be appended at the end.
The length of the arrays is also not necessarily the correct place, e.g. if there are plain text authors present

A comment has been included at https://github.com/DSpace/DSpace/pull/2551/files/e56e71a90338000666a2ce012c3849fc56d0c639#diff-2e89fe18482e7ac43a1a4e901a8105b7R104 to explain this

dspace-api/src/main/java/org/dspace/content/dao/impl/RelationshipDAOImpl.java
```diff
@@ -86,9 +86,9 @@ public int findRightPlaceByRightItem(Context context, Item item) throws SQLExcep
         List<Relationship> list = list(context, criteriaQuery, false, Relationship.class, -1, -1);
         list.sort((o1, o2) -> o2.getLeftPlace() - o1.getLeftPlace());
         if (!list.isEmpty()) {
-            return list.get(0).getLeftPlace();
+            return list.get(0).getLeftPlace() + 1;
```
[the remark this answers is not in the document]
benbosman: we'll check this out

dspace-api/src/test/java/org/dspace/content/RelationshipServiceImplPlaceTest.java
benbosman: it happens implicitly in https://github.com/DSpace/DSpace/pull/2551/files/e56e71a90338000666a2ce012c3849fc56d0c639#diff-56b0380ff7b66daeff6012e659df5863R246 but it can be duplicated explicitly as well if that makes it easier to understand

dspace-api/src/main/java/org/dspace/content/RelationshipServiceImpl.java
Raf-atmire: These relationships here aren't removed from the actual list, they're only removed in this temporary list so that we can perform a proper place calculation further on in this method. I do agree that it was a bit unclear, so I'll add a comment stating this. This is particularly useful incase this method is called for a delete use case, so that we can update the relationships' places properly without the relationship given interfering with the place calculation

dspace-api/src/main/java/org/dspace/content/dao/impl/RelationshipDAOImpl.java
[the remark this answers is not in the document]
benbosman: This has been solved.
The `testGetNextRightPlace` method has been created to verify this works correctly

dspace-api/src/main/java/org/dspace/content/RelationshipServiceImpl.java
benbosman: This has been clarified using inline comments

dspace-api/src/test/java/org/dspace/content/RelationshipServiceImplPlaceTest.java
benbosman: Inline comments have been included

dspace-api/src/test/java/org/dspace/content/RelationshipServiceImplPlaceTest.java
benbosman: Inline comments have been included

dspace-api/src/test/java/org/dspace/content/RelationshipServiceImplPlaceTest.java
benbosman: Inline comments have been included

paulo-graca (approved):
The changes look ok to me. And didn't found any issues (besides the CSV import) within my tests. My thanks to @benbosman and @Raf-atmire for this contribution!

dspace-api/src/test/java/org/dspace/content/RelationshipMetadataServiceTest.java
```diff
@@ -138,5 +138,41 @@ public void testGetRelationshipMetadata() throws Exception {
         assertThat(list.get(1).getMetadataField().getElement(), equalTo("isAuthorOfPublication"));
         assertThat(list.get(1).getAuthority(), equalTo("virtual::" + relationship.getID()));
 
+    }
+
+    @Test
+    public void testGetNextRightPlace() throws Exception {
```
tdonohue: Could we add a corresponding test for `findNextLeftPlaceByLeftItem`?  I just noticed it seems to be missing.

tdonohue (commented):
@benbosman : This is looking good overall now.  Just one (very minor) missing test...see below.  Once that's ready, I think this can be merged.

tdonohue (approved):
👍 Thanks @benbosman and @Raf-atmire.  Looks good to me!

PR #2114, opened 2018-07-10
DS-3946: start-handle-server script fails to remove the lock file

toniprieto:
Pull request for: https://jira.duraspace.org/browse/DS-3946

PR #2552, opened 2019-10-15
DS-3946: start-handle-server script fails to remove the lock file (master port)

nwoodward:
This PR ports the fix in https://github.com/DSpace/DSpace/pull/2114 to 7.x.

Jira issue: https://jira.duraspace.org/browse/DS-3946


nwoodward:
Closing this since another PR with the same fix was merged.

https://github.com/DSpace/DSpace/pull/2929

PR #2556, opened 2019-10-18
DS-4029: adds missing keys in Messages.properties for search filters

nwoodward:
Jira issue: https://jira.duraspace.org/browse/DS-4029


kshepherd (approved):
+1 obvious fix, merging immediately as this is a very simple change
thanks @nwoodward !

PR #2559, opened 2019-10-21
DS-4370 - Allow Structure Builder to import beneath a parent community 

kshepherd:
JIRA: https://jira.duraspace.org/browse/DS-4370
Fixes #7709 

This improvement allows a user to specify a parent community handle or ID as an argument to the structure builder. The structure document will then be imported beneath that community rather than at the root level of the repository.
A new integration test has been written to test this case.

This work was developed by The Library Code GmbH with support of Deutsches Institut für Urbanistik (DIFU) / German Institut of Urban Affairs


kshepherd:
not sure why the Travis test is failing, my local tests were working - I'll push fixes soon

tdonohue (approved):
👍 Thanks @kshepherd !  This looks good.  The provided tests prove it's working properly.

PR #2565, opened 2019-11-04
DS-4377 Fix Sherpa RoMEO layout

AndreaJenisSaroni:
Fixes #7716
Fix for Sherpa RoMEO layout, please see https://jira.duraspace.org/browse/DS-4377 for related screens


abollini (approved):
changes look good to me. They are already applied to DSpace-CRIS. The reason to put the sherpa-romeo on the left is to have it before the upload also when displayed on small screen

kshepherd (approved):
+1 looks good, works in DSpace-CRIS and is same as merged #2567 for 6.x

PR #2566, opened 2019-11-05
[DS-3444] JSPUI must keep shibboleth attributes on session renewal

lap82:
Fixes #6797
Cherry-pick of PR #2358 developed by The Library Code GmbH


kshepherd (approved):
merging this port of #2358, thanks @lap82 !
(not a direct cherry pick as the description suggests - as 5.x uses primitive arrays of group IDs and 6.x uses lists of UUIDs, but it is functionality equivalent)

PR #2567, opened 2019-11-05
DS-4377 Fix Sherpa RoMEO layout

AndreaJenisSaroni:
Fixes #7716
Fix for Sherpa RoMEO layout, please see https://jira.duraspace.org/browse/DS-4377 for related screens.


abollini (approved):
it looks good to me, see also https://github.com/DSpace/DSpace/pull/2565

kshepherd (approved):
+1, looks good (and i've seen the patch working in DSpace-CRIS)
thanks @AndreaJenisSaroni !

PR #2568, opened 2019-11-05
DS-4379 Stop adding relations as metadata for new items

cwilper:
This fixes a bug where metadatavalue rows were incorrectly added on CSV import, when the item was new, and it had relationships to another item. The relationships should only be persisted in the database relationship table, not metadatavalue.

https://jira.duraspace.org/browse/DS-4379


paulo-graca (approved):
This PR fixes the issue DS-4379. Thank you @cwilper for this patch!

tdonohue (approved):
👍 Looks like an obvious fix to the described bug.  Thanks @cwilper

PR #2574, opened 2019-11-11
[DS-4382] BTE must check value pairs if no qualifier is configured

pnbecker:
https://jira.duraspace.org/browse/DS-4382
#7721

BTE checks for every field if it is configured in input-types.xml before it imports any data using that field. Within the check, BTE does not respect qualifies as set in value pairs. This PR solves this bug.

BTE is deprecated in DSpace 7, but if I see it correctly, it will be used by the submission workflow. If that is right, this PR should be ported to DSpace 7.

This fix was developed by [The Library Code](https://www.the-library-code.de) with support of the Deutsches Institut für Urbanistik (German Institute of Urban Affairs).


abollini:
I want to note that we have agreed to remove BTE before to cut 7.0 for this reason this PR is likely to be closed without merge in the next weeks. I'm just keep it open until the BTE removal is complete

abollini (changes requested):
flag as rejected as BTE will be removed in the next weeks

PR #2579, opened 2019-11-13
CRUD for collection item template

jonas-atmire:
REST API implementation for the rest contract changes mentioned in:

* DSpace/Rest7Contract#80
This has recently been slightly altered by the following as well
* DSpace/Rest7Contract#85

#### Quick overview:
This extends the codebase with the CRUD functionality for the collection item template


KevinVdV (changes requested):
Made an initial review & found a few small things that require changing.

dspace-server-webapp/src/main/java/org/dspace/app/rest/CollectionItemtemplateController.java
```diff
@@ -0,0 +1,164 @@
...
+     */
+    @PreAuthorize("hasPermission(#uuid, 'COLLECTION', 'WRITE')")
+    @RequestMapping(method = RequestMethod.POST)
+    public ResponseEntity<ResourceSupport> createTemplateItem(HttpServletRequest request,
+                                                              @PathVariable UUID uuid,
+                                                              @RequestBody(required = false) JsonNode itemBody)
+            throws SQLException, AuthorizeException {
+
+        if (itemBody == null) {
+            throw new BadRequestException("The new item should be included as json in te body of this request");
```
mspalti: Minor typo: 'the body'

dspace-server-webapp/src/test/java/org/dspace/app/rest/ItemTemplateRestControllerIT.java
```diff
@@ -0,0 +1,391 @@
...
+
+    private ObjectMapper mapper;
+    private String adminAuthToken;
+    private Collection childCollection;
+    private ItemRest testTemplateItem;
+    private String patchBody;
+
+    @Before
+    public void createStructure() throws Exception {
+        context.turnOffAuthorisationSystem();
```
mspalti: Would it be a good idea to also restoreAuthSystemState? The tests below pass so I guess it's not necessary!

mspalti (commented):
@jonas-atmire , the code looks good to me.  I added two minor comments.

dspace-server-webapp/src/test/java/org/dspace/app/rest/ItemTemplateRestControllerIT.java
jonas-atmire: @mspalti As the tests are 'reset' to a general predefined stats, as per this setup method: https://github.com/DSpace/DSpace/blob/6e56c8bad15274da523b5f6a819b8f67030ee541/dspace-server-webapp/src/test/java/org/dspace/app/rest/test/AbstractIntegrationTestWithDatabase.java#L98-L106, there is no actual need to do this.

mspalti (approved):

KevinVdV (approved):
I would just like to request 2 small tweaks to variables that are duplicated, once these are removed I approve.

dspace-server-webapp/src/main/java/org/dspace/app/rest/repository/CollectionRestRepository.java
```diff
@@ -70,9 +70,14 @@
     @Autowired
     private BitstreamService bitstreamService;
 
+    @Autowired
+    private ItemConverter itemConverter;
```
KevinVdV: This variable is now unused & can be removed.

KevinVdV (approved):
There is just one small thing left before I can merge this PR, with the latest merge in master a single variable has become unused. Remove it & I can merge it.

PR #2582, opened 2019-11-14
[DS-4385] Replace <prerequisites> with enforcer rules in POM

mwoodiupui:
https://jira.duraspace.org/browse/DS-4385
Maven 3 says that we are abusing `<prerequisites>`.  Use the enforcer plugin instead.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -16,10 +16,6 @@
         <url>http://www.dspace.org</url>
     </organization>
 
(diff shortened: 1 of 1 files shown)


tdonohue (approved):
👍 Looks good to me. Thanks @mwoodiupui .  Merging this immediately as it's a minor refactor to POM.

PR #2589, opened 2019-11-19
Entities: Search relationships total fix

benbosman:
The totals for the `/api/core/relationships/search/byLabel` endpoint has a problem
If there are multiple relationship types with the requested label, the total is not correct.

This has been tested and ITs have been included to verify this


tdonohue (approved):
👍 This is a very obvious bug. Thanks @benbosman for including the tests to prove it now works.

PR #2594, opened 2019-11-21
Bug in GET collections

benbosman:
Similar to https://github.com/DSpace/DSpace/pull/2486, we discovered a problem in /server/api/core/collections
The same collection is returned multiple times if there are multiple dc.title values

This PR contains:
* A fix similar to https://github.com/DSpace/DSpace/pull/2486
* An IT to proof this resolves the problem


tdonohue (approved):
👍 This looks good to me, and looks like the same implementation as #2486 (which fixed this bug for Communities).  This one fixes it for Collections.

So, merging immediately.

PR #2596, opened 2019-11-22
[DS-4390] rewrote ScriptLauncher to include scripts from other config…

Raf-atmire:
Rewrote the ScriptLauncher so that it now checks the Scripts defined in the scripts.xml file as DSpaceRunnable beans as well before falling back on the Launcher.xml scripts
Also fixed issues with tests that arose due to this change
Jira: https://jira.lyrasis.org/browse/DS-4390


KevinVdV (approved):
Looks good to me 👍

tdonohue (approved):
👍 This looks good to me, thanks @Raf-atmire !  

I've tested this locally, and with this PR applied, I can now run `./dspace index-discovery` commands from the commandline again.  This was *not* previously possible as the #2529 PR removed `index-discovery` from our `launcher.xml` config, in order to move that script to the REST API's "scripts" endpoint.

One *very minor* bug I discovered in my testing.  While `./dspace index-discovery` commands work on commandline, if I just run `./dspace` from commandline I get a list of all the commandline scripts in `launcher.xml`.  This list unfortunately does **NOT** include `index-discovery` as an option, since it's no longer listed in the `launcher.xml`.  So, we need to find a way to ensure scripts that move the the REST API are still listed in the help info on the commandline.  I'll move this message to the JIRA ticket (DS-4390), as we'll keep that ticket open until this additional minor bug is resolved.

Merging as-is.  Thanks again @Raf-atmire !

PR #2597, opened 2019-11-23
DS-4391 Get modified date from the right store

cwilper:
Fixes https://jira.lyrasis.org/browse/DS-4391

This also tightens up the contract and improves javadocs a bit, so callers know what to expect.

### DIFF
--- a/dspace-api/src/main/java/org/dspace/storage/bitstore/BitstreamStorageServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/storage/bitstore/BitstreamStorageServiceImpl.java
@@ -16,6 +16,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 
+import javax.annotation.Nullable;
+
 import org.apache.commons.collections4.MapUtils;
 import org.apache.logging.log4j.Logger;
 import org.dspace.authorize.AuthorizeException;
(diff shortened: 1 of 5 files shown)


dspace-api/src/main/java/org/dspace/storage/bitstore/BitstreamStorageServiceImpl.java
```diff
@@ -326,15 +328,16 @@ public void cleanup(boolean deleteDbRecords, boolean verbose) throws SQLExceptio
         }
     }
 
-    public Long getLastModified(Bitstream bitstream) {
-        Map wantedMetadata = new HashMap();
-        wantedMetadata.put("modified", null);
-        try {
-            wantedMetadata = stores.get(incoming).about(bitstream, wantedMetadata);
```
cwilper: This was the problematic line...note how it hardcodes incoming. This PR corrects this call to use getStoreNumber() so the request is correctly delegated.

mwoodiupui:
A test procedure:
1. Install a fresh instance.
1. Create a Community and a Collection.
1. Ingest an Item.
1. Edit config/spring/api/bitstore.xml:
    1. add a second local store with a second directory and a new `key`
    1. alter the `incoming` property of the BitstreamStorageService to the value of that key.
    1. if you are doing this through the GUI, restart the `server` webapp to re-read the configuration.
1. Ingest another Item.  You now have two assetstores with one item in each.  Both items are in the same Collection.
1. If the server webapp. is not running, start it.
1. Fetch each item's bitstream and see that we no longer get a 500 error.

mwoodiupui (approved):
Tested as described above.  This seems to work.  I examined the code and it looks right.

tdonohue (approved):
Looks good to me!

PR #2602, opened 2019-11-28
DS-4397 speed up controlled vocabulary loading (6.x)

helix84:
significant speed improvement with 10k entries

https://jira.lyrasis.org/browse/DS-4397

Fixes #7735


nwoodward (approved):
:+1: Tested and confirmed it works. I noticed a smaller performance improvement with the given sample dictionary, but that may have to do with my testing.

alanorth:
Tested with some controlled vocabularies with ~1,500 items on our DSpace 6.3 instance and it at least doesn't break anything. Looks good to me. :+1:

benpal:
Will it work in XMLUI in DSpace 6.3? Or, what to do with vocabulary-support.js?

PR #2605, opened 2019-11-28
[DS-4393] Discovery clean index fix DSpace 5

KevinVdV:
Fixes #7731
The "index-discovery" command when run without argument or with "-c" will retrieve all documents from the solr index, check if they are still in the database & if not remove them.

There is a bug in this process however, the retrieval of "all" the documents will only retrieve the first 10 check those & then stop.


kshepherd:
just temporarily closing and re-opening this PR to force Travis CI to rebuild (on the new service)

kshepherd (approved):
+1 merging this port of PRs already merged for master and 6.x

PR #2607, opened 2019-11-28
[DS-4393] Discovery clean index fix

KevinVdV:
Fixes #7731

The "index-discovery" command when run without argument or with "-c" will retrieve all documents from the solr index, check if they are still in the database & if not remove them.

There is a bug in this process however, the retrieval of "all" the documents will only retrieve the first 10 check those & then stop.


benbosman (approved):
The code looked good to me.

I have tested this by taking an old DB and a solr index from a much newer DB containing much more items to ensure the findIndexableObject will encounter many items which don't exist in the DB

Initial state:
```json
"search.resourcetype":[
        "2",1121,
        "8",224,
        "3",16,
        "9",12,
        "10",8,
        "4",5,
        "11",4],
```

After executing "index-discovery -c" with the master code, max 10 records per type were deleted:
```
"search.resourcetype":[
        "2",1111,
        "8",217,
        "3",6,
        "9",2]
```

After executing "index-discovery -c" with this branch:
```
[
        "2",311,
        "8",5]
```

This matches the expected behavior.

In both use cases, an explicit solr commit was necessary

tdonohue (approved):
👍 The code here looks correct. Thanks @KevinVdV !

PR #2610, opened 2019-12-03
Updating Relationships through JSON put causes place alterations to be broken

Raf-atmire:
This PR fixes a minor error where right was used instead of left
It also includes an IT to verify the fix


paulo-graca (approved):
Everything looks ok!

tdonohue (approved):
👍 Looks good to me. Thanks for catching & fixing this @Raf-atmire

PR #2611, opened 2019-12-05
DS-4380 Update build to work with JDK 11

cwilper:
# Update as of 2020/01/09

This PR is a starting point but I don't have much time to spend on this currently. For anyone taking it on, here's what I would advise:

* Start from latest master (with Kevin's latest dependency updates)
* Look at the changes in this PR and incorporate them or something close to them
* Get tests working again.

By far, the most difficult part is going to be dealing with `jmockit`. It's an extensively used library in historic "unit" tests for DSpace, and the old version no longer works with Java 11. Upgrading to the latest version is one path forward, but note, the API has changed significantly and certain features are no longer available. Another approach is to try to switch the legacy tests away from jmockit and have them use the latest mockito instead.  IMO, moving to away from jmockit and toward more "pure" unit tests with mockito seems preferable for consistency with newer tests, but requires more significant refactoring.

# Original Description

> NOTE: Tests are currently NOT expected to pass because a) they need updates to work with newer jmockit version, and b) test server needs update to build with JDK 11.

https://jira.lyrasis.org/browse/DS-4380

This updates the following in order to support builds with Java 11:

- Requiring Java 11 to build
- Manually including JAXB as it is no longer bundled with JDK 11
- Updating jacoco to a version that supports JDK 11
- Updating jmockit to a version that supports JDK 11

The following remain to be done:

- Re-enabling errorprone for the build to work with JDK 11
- Updating tests to work with newer version of jmockit

### DIFF
--- a/dspace-api/pom.xml
+++ b/dspace-api/pom.xml
@@ -50,10 +50,23 @@
                 <configuration>
                     <debug>true</debug>
                     <showDeprecation>true</showDeprecation>
-                    <compilerArguments>
-                        <processor>org.hibernate.jpamodelgen.JPAMetaModelEntityProcessor</processor>
-                    </compilerArguments>
-
+                    <annotationProcessorPaths>
+                        <annotationProcessorPath>
+                            <groupId>org.hibernate</groupId>
+                            <artifactId>hibernate-jpamodelgen</artifactId>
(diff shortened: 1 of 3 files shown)


cwilper:
Closing; replaced by https://github.com/DSpace/DSpace/pull/2654

PR #2613, opened 2019-12-06
Avoid setting the virtual metadata language to ANY (*)

benbosman:
This is a minor change to the virtual metadata

This used to create:
```
"dc.contributor.author": [
      {
        "value": "Demirguc-Kunt, Asli",
        "language": "*",
        "authority": "virtual::399",
        "confidence": -1,
        "place": 0
      },
      {
        "value": "Maksimovic, Vojislav",
        "language": null,
        "authority": null,
        "confidence": -1,
        "place": 1
      }
    ]
```

The `"language": "*"` will now become `"language": null`


tdonohue (approved):
👍 This seems fine to me. I'm going to merge this immediately, as it's a simple one-liner.  I agree that language should be null when unspecified (which is how it was in XMLUI and JSPUI as well)

PR #2615, opened 2019-12-09
Get entity type performance improvement

benbosman:
This is a minor change to improve the performance of the EntityServiceImpl.getType() function.
This function is called regularly while it doesn't require virtual metadata.


tdonohue (approved):
👍 This looks fine to me, and seems like a very simple obvious fix.

benbosman:
was marked as 1 APPROVAL, merging

PR #2617, opened 2019-12-09
DS-4401 Enforce relational place ordering when place direction is known

AndrewZachWood:
From: https://jira.lyrasis.org/browse/DS-4401

Small improvement to enforce ascending order by leftward or rightward place when calling `findByItemAndRelationshipType` with the `isLeft` param.

The need for this fix came about when I requested the relationships from an item that had around 1,800~ relationships and received them in an mixed order which would cause a bad user experience when you're dependent on relational ordering.

### DIFF
--- a/dspace-api/src/main/java/org/dspace/content/dao/impl/RelationshipDAOImpl.java
+++ b/dspace-api/src/main/java/org/dspace/content/dao/impl/RelationshipDAOImpl.java
@@ -145,11 +145,13 @@ public class RelationshipDAOImpl extends AbstractHibernateDAO<Relationship> impl
                     .where(criteriaBuilder.equal(relationshipRoot.get(Relationship_.relationshipType),
                             relationshipType),
                            criteriaBuilder.equal(relationshipRoot.get(Relationship_.leftItem), item));
+            criteriaQuery.orderBy(criteriaBuilder.asc(relationshipRoot.get(Relationship_.leftPlace)));
         } else {
             criteriaQuery
                     .where(criteriaBuilder.equal(relationshipRoot.get(Relationship_.relationshipType),
                             relationshipType),
                             criteriaBuilder.equal(relationshipRoot.get(Relationship_.rightItem), item));
+            criteriaQuery.orderBy(criteriaBuilder.asc(relationshipRoot.get(Relationship_.rightPlace)));
         }
         return list(context, criteriaQuery, true, Relationship.class, limit, offset);
     }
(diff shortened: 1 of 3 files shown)


benbosman (approved):
I've verified this functionality in master:
* /api/core/items/<:uuid>/relationships was already correct
* /api/core/relationships/search/byLabel?label=<:label>&dso=<:uuid> doesn't take the place into account (tested by using PUT on the place)

This uses findByItemAndRelationshipType, so this should solve the problem (didn't test the new branch but it looks good)

benbosman (changes requested):
As discussed in today's meeting, I've now also tested this specific branch.

It seems this doesn't solve the order on /api/core/relationships/search/byLabel?label=<:label>&dso=<:uuid> yet.
It ensures the order is correct in `org.dspace.content.dao.impl.RelationshipDAOImpl#findByItemAndRelationshipType(org.dspace.core.Context, org.dspace.content.Item, org.dspace.content.RelationshipType, boolean, java.lang.Integer, java.lang.Integer)` but /api/core/relationships/search/byLabel?label=<:label>&dso=<:uuid> actually uses `org.dspace.content.dao.impl.RelationshipDAOImpl#findByItemAndRelationshipType(org.dspace.core.Context, org.dspace.content.Item, org.dspace.content.RelationshipType, java.lang.Integer, java.lang.Integer)`. For that method, there's of course no way to determine the relevant order

How can this also be made available for `/api/core/relationships/search/byLabel?label=<:label>&dso=<:uuid>`:
The org.dspace.app.rest.repository.RelationshipRestRepository#findByLabel() could request the findByItemAndRelationshipType function for left or for right. Whether left or right should be used actually depends on which side contains the label in the parameter. So the query can be `relationshipService.findByItemAndRelationshipType(context, item, relationshipType, relationshipType.getLeftwardType().equals(label), pageable.getPageSize(), pageable.getOffset())`

AndrewZWood:
@benbosman I believed I've covered the requested changes in the latest commit. I've updated RelationshipRestRepository#findByLabel to determine and include the direction of the `label` parameter for `relationshipService.findByItemAndRelationshipType()` when a DSO is specified in the request. I've also updated the RelationshipRestRepositoryIT#findRelationshipByLabelTest to respect the improvement and check the ordering of the relationships returned in the response.

benbosman (approved):
Thanks @AndrewZWood 

I've tested it again, it now solves the order on /api/core/relationships/search/byLabel?label=<:label>&dso=<:uuid>.

I've verified this with paging as well, and size 0 page 0 returned me the items with place 0 to 4, size 0 page 1 returned me the items with place 5 to 9, size 0 page 2 returned me the items with place 10 to 14, …

This can also be seen in the UI using https://github.com/DSpace/dspace-angular/pull/532
The master REST returns https://www.dropbox.com/s/j4z6lobouhj1959/Schermafdruk%202019-12-12%2011.10.53.png?dl=1 while this branch returns https://www.dropbox.com/s/5uk3lexj4z9v7k0/Schermafdruk%202019-12-12%2011.12.04.png?dl=1

benbosman:
Was marked as 1 APPROVAL, merging

PR #2620, opened 2019-12-10
[DS-3879] Enable XInclude for some XML configuration documents

mwoodiupui:
Fixes #7226
https://jira.lyrasis.org/browse/DS-3879
Allow the submission configuration to be broken up into multiple files.  These are large documents with groups of related elements, which documents are meant to be manually edited.  They could benefit from modularization.


mwoodiupui:
Might be useful for concentrating e.g. fields specific to OpenAire in a separate document, for clarity and locality.  #2608

paulo-graca:
I tend to agree with this. This is helpful!

mwoodiupui:
This will be somewhat less wieldy than I had imagined, due to a very incomplete implementation of XPointer in Xerces-J.  I've provided an example by pulling the journal forms out to an external document and `xi:include`ing each individually, which seems to be the only way this will work.

PR #2621, opened 2019-12-10
OpenAIRE4 adding missing field person.identifier

paulo-graca:
There is a missing generic person identifier field (without specifying a type).


tdonohue (approved):
👍 This looks fine to me. Thanks @paulo-graca for the quick fix

PR #2622, opened 2019-12-11
Default Entities relationship types used for the demos

benbosman:
This is the default configuration we've used for the tests and in the UI for importing entity types and relationship types


paulo-graca (approved):
Thank you @benbosman for adding this and for making the change on the openaire4-relationships file!

tdonohue (approved):
👍 Looks correct to me & this is only (optional) configuration changes.  Thanks @benbosman

PR #2638, opened 2020-01-09
Bug in ViewEvents

jonas-atmire:
Based on the following angular related issue/comment:
https://github.com/DSpace/dspace-angular/pull/532#issuecomment-566167627

The rest response contained a property: id: null
1. The ID is not relevant in this case
2. A null ID should not be returned.

This PR handles the following:
- Remove 'id' field from the ViewEventRest response

Similar to HarvesterMetadataRest, this has been ignored, so REST does not return that id (without changing any other functionality)


paulo-graca (approved):
The code changes look ok to me. And testing without no noticeable errors and correctly logging events. Thank you @jonas-atmire for adding this fix!

tdonohue (approved):
👍 Looks fine to me as well.

PR #2641, opened 2020-01-16
relationship type rest controller bug fix

benbosman:
This is a bug fix for `/api/core/entitytypes/{id}/relationshiptypes` for which the PR was apparently forgotten

Currently https://dspace7.4science.cloud/server/#https://dspace7.4science.cloud/server/api/core/entitytypes/4/relationshiptypes contains a few properties which are not relevant since the relationshiptypes are being retrieved (and not the property of the entity type). This causes errors in Angular.
The fix was deployed on https://dspace7-entities.atmire.com/server/#https://dspace7-entities.atmire.com/server/api/core/entitytypes/4/relationshiptypes where the properties are no longer displayed.

It was forgotten to create the PR for this change


dspace-server-webapp/src/main/java/org/dspace/app/rest/RelationshipTypeRestController.java
```diff
@@ -64,32 +65,28 @@
      * @param id        The ID of the EntityType objects that we'll use to retrieve the RelationshipTypes
      * @param response  The response object
      * @param request   The request object
-     * @return          The wrapped resource containing the list of RelationshipType objects as defined above
+     * @return The wrapped resource containing the list of RelationshipType objects as defined above
```
paulo-graca: please update with the missing params.

paulo-graca (approved):
Thank you @benbosman !

tdonohue (approved):
👍 Looks good to me too. Thanks @benbosman

PR #2644, opened 2020-01-17
Fix for DS-4422

paulo-graca:
Fix for issue reported in https://jira.lyrasis.org/browse/DS-4422


tdonohue (approved):
Tiny, obvious fix.  Thanks @paulo-graca !

PR #2645, opened 2020-01-17
POC: Using Spring MVC patch 

mspalti:
This PR explores using Spring patch support for non-metadata patch operations. Metadata patch operations don't appear to be possible using Spring, so this PR continues to use the existing json patch strategy for metadata.

The idea for this PR grew out of discussion in https://github.com/DSpace/DSpace/pull/2591 which implements a new patch strategy that operates on the dspace business object.


mspalti:
Changes in this PR do not apply to workflow or workspace patches.

dspace-server-webapp/src/test/java/org/dspace/app/rest/EPersonRestRepositoryIT.java
```diff
@@ -615,7 +615,7 @@ public void patchUsingStringForBoolean() throws Exception {
         List<Operation> ops = new ArrayList<Operation>();
 
         // String should be converted to boolean.
-        ReplaceOperation replaceOperation = new ReplaceOperation("/canLogin", "true");
```
mspalti: Unfortunately there are two incorrect paths in the Rest contract for eperson patches (canLogin and certificate). This will need to be changed, or perhaps the EpersonRest model can be changed instead.  For the purposes of this POC I just updated the integration tests.

mspalti:
I rebased and made some modifications that allow all ITs to pass. 

The behavior of Spring patch `apply` changed after I rebased: the method now requires the current rest model (e.g. `Class<ItemRest>`) as a parameter instead of one of it's parent classes. There's a pre-existing method on repositories that provides us with the Class.  (I haven't tacked down the explanation for this change, but presumably it is best to use the rest model anyway.)

Validation of patch requests and error handling are done in the `PatchUtils` class and repository classes. Some of this could perhaps be improved. Previously the work was done in the individual patch operation classes (which have been removed here).

PR #2652, opened 2020-01-23
[DS-4409] Remove GeoIP download Ant target, reconfigure for external provision

mwoodiupui:
https://jira.lyrasis.org/browse/DS-4409
Ported #2632 to dspace-6_x.


tdonohue (approved):
👍 Changes look fine to me (matches with the `master` PR).  I haven't tested though

kshepherd (approved):
+1 looks good to me, i just added one small commit to update the error messages in SolrLoggerServiceImpl (more explicit note to see install instructions, and make sure it referes to the correct cfg file), thanks @mwoodiupui 
We definitely do need clear doco in the wiki for this, i found it a bit of a pain to go through the process of Find appropriate geoipupdate .deb -> Register Account -> Make free licence -> update config -> update db, etc, and I can imagine people on less common OS's might have more trouble, too

PR #2677, opened 2020-02-17
Non-admin User can't create new item via REST

qzhao-ulb:
REST interface should support user to create a new item in a collction. 
But if the user is not an administrator of this collection, it failed, the system checks the authority for adding item in this collection with WRITE action. 
I think it should check that with action ADD, then it allows normaler user to add new item in this collection.


abollini (approved):
code looks good to me, I have restarted travis as an error was occurred during the build

PR #2680, opened 2020-02-19
Read only versioning endpoints

KevinVdV:
PR for the following rest contract: https://github.com/DSpace/Rest7Contract/pull/97


tdonohue (commented):
@KevinVdV : I've tested & reviewed this today.  First off, it looks to work as described in https://github.com/DSpace/Rest7Contract/pull/97  I can see version / version histories of Items that have versions previously created, and Items without versions give a 204 response.

The code here looks good as well overall, though I'm not sure I understand why the `VersionRestController` is necessary (see note inline below).  If there's a reason why this Controller needs to exist, then please enhance the JavaDocs to describe how it differs from `EPersionVersionLinkRepository` (as currently the two seem duplicative to me).

Also, please rebase this PR as it has minor merge conflicts. 

Once both of these are cleaned up, I think this can be merged.

KevinVdV:
@tdonohue The "VersionRestController" & "EPersionVersionLinkRepository" were deleted, we were exposing the entire EPerson record which is not something we want (in DSpace 6 we currently only expose the name).

We are now exposing only the name of the submitter for admins (or to the public if the versioning is configured in that manner).
These changes also required some changes to the rest contract, which can be found here: https://github.com/DSpace/Rest7Contract/pull/112

tdonohue (approved):
👍 Looks great to me.  I've re-reviewed and re-tested and it's working as described. Thanks for the updates here @KevinVdV and @Raf-atmire !

PR #2682, opened 2020-02-19
Default to full projection for workspace item POSTs

cwilper:
This is a workaround for the issue discussed on dev-sprint 2010/02/18-19. Angular currently expects embeds here, and this gets it quickly working again.


tdonohue (approved):
👍  I've tested this and it fixes the immediate problem.  

NOTE for the future. This PR was a "quick fix" workaround to a larger issue described in more detail in https://github.com/DSpace/dspace-angular/issues/581.  It is necessary because the Submission UI cannot function without embedded content, and `POST` requests do not yet accept the `projection=full` parameter.

PR #2691, opened 2020-02-25
Metadata PATCH when mixing plain text and relationship

benbosman:
While working with the submission, we noticed a few problems with metadata PATCH when mixing plain text and relationship
This contains fixes for those PATCH requests and ITs
The pure plain-text PATCHes are also still working correctly, there were ITs for those use cases


paulo-graca (commented):
It looks ok to me. I'm going to test it.

paulo-graca:
I'm getting this error on angular side when I try to access submission form (the same doesn't occurs with master branch):

image

benbosman:
@paulo-graca Thanks for the review.
This is most likely due to the fact that this branch hasn't been merged with master for a while.
I've merged it now, and tested locally that it works with the latest Angular code

paulo-graca:
Thank you @benbosman! I will redo my tests.

paulo-graca (commented):
I would like to give a thumbs up to this PR because all code changes look ok to me, but I didn't quite understand the reasons to justify this PR, neither I have a full workable setup to validate them. I'm still encountering the issue reported here: https://github.com/DSpace/dspace-angular/pull/531#issuecomment-564094326, so I can't have plain text and entities working to properly test it.

paulo-graca:
@benbosman is this PR related with issue: https://github.com/DSpace/dspace-angular/issues/817 ?

benbosman:
@paulo-graca I've updated and tested this branch and this doesn't seem related
The issue also seems to occur prior to sending something to REST

dspace-api/src/main/java/org/dspace/content/RelationshipMetadataValue.java
```diff
@@ -57,4 +59,8 @@ public boolean equals(Object obj) {
         }
         return super.equals(obj);
     }
+
+    public int getRelationshipId() {
```
tdonohue, line 63: Please add JavaDocs to describe this method.  It looks like it's just extracting the RelationshipID corresponding to the current RelationshipMetadataValue?

dspace-api/src/main/java/org/dspace/content/service/DSpaceObjectService.java
```diff
@@ -202,8 +202,8 @@
      * @param values    the values to add.
      * @throws SQLException if database error
      */
-    public void addMetadata(Context context, T dso, String schema, String element, String qualifier, String lang,
-                            List<String> values) throws SQLException;
+    public List<MetadataValue> addMetadata(Context context, T dso, String schema, String element, String qualifier,
```
tdonohue, line 205: Could you update the JavaDocs for all these modified methods to have a `@return` annotation now & describe what is returned?

dspace-server-webapp/src/test/java/org/dspace/app/rest/PatchMetadataIT.java
```diff
@@ -0,0 +1,803 @@
...
+        // Make sure we grab the latest instance of the Item from the database before adding a regular author
+        WorkspaceItem publication = workspaceItemService.find(context, publicationItem.getID());
+        itemService.addMetadata(context, publication.getItem(),
+                "dc", "contributor", "author", Item.ANY, authorsOriginalOrder.get(0));
+        workspaceItemService.update(context, publication);
+
+        context.restoreAuthSystemState();
+
+        // Create a relationship between publication and person 1
+        MvcResult mvcResult = getClient(adminToken).perform(post("/api/core/relationships")
```
tdonohue, line 155: This `post` needs to be updated to cleanup after itself.  Remember the (newer) best practices to use an `andDo` in a try/finally, similar to this test: https://github.com/DSpace/DSpace/blob/main/dspace-server-webapp/src/test/java/org/dspace/app/rest/RelationshipRestRepositoryIT.java#L278

dspace-server-webapp/src/test/java/org/dspace/app/rest/PatchMetadataIT.java
```diff
@@ -0,0 +1,803 @@
...
+        regularMetadata.add(authorsOriginalOrder.get(2));
+        regularMetadata.add(authorsOriginalOrder.get(3));
+        itemService.addMetadata(context, publication.getItem(),
+                "dc", "contributor", "author", null, regularMetadata);
+        workspaceItemService.update(context, publication);
+
+        context.restoreAuthSystemState();
+
+        // Create a relationship between publication and person 2
+        mvcResult = getClient(adminToken).perform(post("/api/core/relationships")
```
tdonohue, line 178: Again, this `post` needs to be updated to cleanup after itself.  In this scenario, since there are two posts here, you could have a try/finally around essentially most of this test method & do cleanup of both after the assertions below

dspace-server-webapp/src/test/java/org/dspace/app/rest/PatchMetadataIT.java
```diff
@@ -0,0 +1,803 @@
...
+                        (org.springframework.data.rest.webmvc.RestMediaTypes.TEXT_URI_LIST_VALUE))
+                .content("https://localhost:8080/server/api/core/items/" + publicationItem.getItem().getID() + "\n" +
+                                "https://localhost:8080/server/api/core/items/" + personItem2.getID()))
+                .andExpect(status().isCreated())
+                .andReturn();
+
+        publication = workspaceItemService.find(context, publicationItem.getID());
+        List<MetadataValue> publicationAuthorList =
+                itemService.getMetadata(publication.getItem(), "dc", "contributor", "author", Item.ANY);
+        assertThat(publicationAuthorList.size(), equalTo(5));
```
tdonohue, line 190: Just a note, this sort of assertion is easier to read/understand (and shorter) if you use `assertEquals()` instead of the longer `assertThat(...., equalTo())`.  It's not a required change, but I find it slightly easier to read.

dspace-server-webapp/src/test/java/org/dspace/app/rest/PatchMetadataIT.java
```diff
@@ -0,0 +1,803 @@
...
+    }
+
+    /**
+     * A method to create a workspace publication containing 5 authors: 3 regular authors and 2 related Person items.
+     * The authors are added in a specific order:
+     * - "Whyte, William": Regular author
+     * - "Dahlen, Sarah": Related Person
+     * - "Peterson, Karrie": Regular author
+     * - "Perotti, Enrico": Regular author
+     * - "Linton, Oliver": Related Person
```
tdonohue, line 104: It looks (to me) like we are always testing with mixed content...some Entities, and some textual authors.  Could we add a few basic tests to verify this patching also works for *only textual authors*?

tdonohue (changes requested):
@benbosman : The code here looks good overall.  However, I'm unclear how to manually test this (i.e. is there a bug that is reproducible before applying this PR which should now be fixed).  The ITs however look to be good (though I asked for minor improvements & some tests to prove patching plain text authors works fine too).  So, overall, as long as the ITs are updated to prove this refactor works as expected, then I think that does provide some proof it works (even if it's hard to verify manually)

More comments inline below.

dspace-api/src/main/java/org/dspace/content/RelationshipMetadataValue.java
benbosman: The JavaDocs have been created

dspace-api/src/main/java/org/dspace/content/service/DSpaceObjectService.java
benbosman: The JavaDocs have been created

dspace-server-webapp/src/test/java/org/dspace/app/rest/PatchMetadataIT.java
benbosman: The cleanup code has been included

dspace-server-webapp/src/test/java/org/dspace/app/rest/PatchMetadataIT.java
benbosman: The cleanup code has been included

dspace-server-webapp/src/test/java/org/dspace/app/rest/PatchMetadataIT.java
benbosman: The assertEquals is now used

dspace-server-webapp/src/test/java/org/dspace/app/rest/PatchMetadataIT.java
benbosman: Another set of test for only textual authors has been included

paulo-graca (approved):
I was able to change metadata using this PR.
Thank you @benbosman !

tdonohue (approved):
👍 Re-reviewed and my prior feedback was addressed. I also tested this PR with the UI, and found no change in behavior with editing Community/Collection/Item metadata or editing/updating WorkspaceItem metadata.

PR #2696, opened 2020-02-27
DS-4291 Properly show results for 0-9 link in browse

### LINKED ISSUE #7631: [DS-4291] Browse Starts-With 0-9 only returns results for 0
<i>Imported from JIRA <a href="https://jira.lyrasis.org/browse/DS-4291">[DS-4291]</a> created by kingkrimmson</i>
<p>Currently when browsing by Author, if you click the "0-9" link at the top, it is only showing results for 0 instead of 0-9.</p>

<p>Under the hood, this is being done by a solr facet query prefix, which doesn't support ranges and so is currently only returning results for 0. However, it is fairly quick (even on a large repo) to perform separate queries with prefixes 0-9.</p>

aprilherron:
When browsing by Author, if you click the "0-9" link at the top, it is only showing results for 0 instead of 0-9.

Currently, only a facet query prefix for "0" is being used, so no other results are shown. Because facet query prefixes don't support ranges, the simplest solution is to perform separate queries and combine the results for 0-9. This fix performs the additional queries and combines their results and totals.

This is the 7.x version of #2463 

Fixes #7631

### DIFF
--- a/dspace-api/src/main/java/org/dspace/browse/BrowseEngine.java
+++ b/dspace-api/src/main/java/org/dspace/browse/BrowseEngine.java
@@ -381,7 +381,8 @@ public class BrowseEngine {
             // this is the distinct table constrained to either community or collection
             dao.setTable(browseIndex.getDistinctTableName());
 
-            dao.setStartsWith(StringUtils.lowerCase(scope.getStartsWith()));
+            dao.setStartsWith("0".equals(scope.getStartsWith()) && !scope.getOrder().equals("ASC") ?
+                    "9" : StringUtils.lowerCase(scope.getStartsWith()));
             // remind the DAO that this is a distinct value browse, so it knows what sort
             // of query to build
             dao.setDistinct(true);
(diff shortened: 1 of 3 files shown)


tdonohue:
@KingKrimmson : Just a note that this PR looks to be failing in Travis because of checkstyle issues.  Checkstyle validation is required on `master`, see https://wiki.lyrasis.org/display/DSPACE/Code+Style+Guide

tdonohue:
@KingKrimmson : Any chance this could be updated so it can be applied to 7.x?

alanorth:
I don't think this affects DSpace 7+ Angular interface anymore? The browse by author does not break down the results by number or letter anymore. See screenshot from DSpace demo running 7.6.1:

Screenshot 2024-03-07 at 08-15-27 DSpace Repository Browsing by Author-fs8

tdonohue:
Closing as this appears obsolete per @alanorth 's [comment above](https://github.com/DSpace/DSpace/pull/2696#issuecomment-1982365422)

PR #2699, opened 2020-03-02
DS-4073 fix. FindByValue should pass in value, not qualifier (dspace-6_x port)

nwoodward:
Port of https://github.com/DSpace/DSpace/pull/2454 to fix the same bug in DSpace 6.


kshepherd (approved):
+1 merging now as this is port of merged #2454 master PR ,thanks @nwoodward

PR #2700, opened 2020-03-03
Simple Search / Did you mean: Link fix

ghost:
I recently noticed that the link of the did you mean section was empty. It shows a recommended word but it doesn't insert it into the link.


kshepherd:
Hi @GodmanGen 
I can't reproduce an actual problem with this, and I can see that the query value is handled, but it's with a javascript event handler rather than including in the URL via JSP:
`jQ( "#spellCheckQuery").click(function(){
			jQ("#query").val(jQ(this).attr('data-spell'));
			jQ("#main-query-submit").click();
		});`

If I am missing something and there is a problem with spellcheck links in JSPUI, please open a JIRA issue explaining the expected and actual behaviour with some instructions to help us reproduce the problem. Thanks!

GodmanGen:
Yeah! You are correct. I didn't notice the javascript handler since at the project I'm working on, the link is always empty. I will check it now. Thank you for your quick reply!

GodmanGen:
Also tested on the online demo and it works as is.

PR #2710, opened 2020-03-10
[DS-2715] porting ORCID Integration for JSPUI (6.x) 

lap82:
This PR addressed also resolution for DS-4439

### DIFF
--- a/dspace-api/src/main/java/org/dspace/authority/orcid/Orcidv2AuthorityValue.java
+++ b/dspace-api/src/main/java/org/dspace/authority/orcid/Orcidv2AuthorityValue.java
@@ -7,6 +7,7 @@
  */
 package org.dspace.authority.orcid;
 
+import org.apache.commons.lang.ObjectUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.solr.common.SolrDocument;
 import org.apache.solr.common.SolrInputDocument;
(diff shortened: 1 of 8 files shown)


abollini (approved):
found some time today to check it. Thanks for this contribution it looks good to me

kshepherd:
Doing the "close+reopen" trick to force the new Travis CI build to run again

abollini:
@kshepherd travis confirm that it works, should we merge?

kshepherd (commented):
I've given this a test now - sorry for my previous comment, that was my fault - I forgot to add in the authority index consumer!

Everything looks to be working well! There are a few UX issues with the lookup form but those are pre-existing and the functionality now matches XMLUI.

The fix for DS-4439 is also a welcome fix.

kshepherd (approved):

nwoodward (approved):

kshepherd:
Merging now, thanks @lap82 !

aatanasijevic:
Not sure that is a issue but  "Name Authority author lookup" popup after accept return this "will be generated::orcid::XXXXX" in authority field instead orcid number.

kshepherd:
@aatanasijevic this is how new authority records are created in submission or item edits - when the item is next posted/updated and the authority key indexed, the indexer checks to see if it begins with that prefix and if it does, creates a new record. if you try it, once you've submitted the item or updated an existing one, you should see the authority key UUID in teh form. (note, you'll never see the ORCID ID there - though you can see it in the Solr index when you look up that authority key UUID)

You can see more info here (though some of it may be out of date) https://wiki.lyrasis.org/display/DSDOC6x/ORCID+Integration#ORCIDIntegration-DifferentpossibleusecasesforIndex-authorityscript

alanorth:
Note that this also needs a documentation update on [ORCID Integration](https://wiki.lyrasis.org/display/DSDOC6x/ORCID+Integration). I think we can simply remove this warning now?

Screenshot 2022-05-09 at 09-39-32 ORCID Integration - DSpace 6 x Documentation - LYRASIS Wiki

kshepherd:
@alanorth good catch, i've done this

PR #2730, opened 2020-04-01
DS-4342 improve performance when retrieving items from a collection (7.x)

nwoodward:
Port of https://github.com/DSpace/DSpace/pull/2516


tdonohue (approved):
👍 Looks good.  Merging this immediately as it's a forward-port of #2516 and it only impacts the deprecated v4-6 REST API.

PR #2742, opened 2020-04-08
bump up pdfbox version on 6.x to match main branch

nwoodward:


J4bbi:
thanks for the PR @nwoodward 

I think the current version is 2.0.24, go for that?

alanorth:
To be clear, as I had to look this up myself: in [DSpace 6, PDFBox is used by the filter-media script](https://wiki.lyrasis.org/display/DSDOC6x/Mediafilters+for+Transforming+DSpace+Content) to both extract full text (which most people should be using) and to generate thumbnails (which was [added in DSpace 6.3](https://wiki.lyrasis.org/display/DSDOC6x/Release+Notes#ReleaseNotes-6.3ReleaseNotes)). This updates the library to the latest minor version (from 2.0.2 to 2.0.24), which likely [includes _hundreds_ of performance, stability, and security updates](https://downloads.apache.org/pdfbox/2.0.24/RELEASE-NOTES.txt), and perhaps even improved text extraction, etc.

I tested pdfbox versions 2.0.2 and 2.0.24 on our DSpace 6.3 repository, forcing the filter-media to run on a collection with 250 PDFs and comparing the output:

```console
$ grep -c '^File:' /tmp/pdfbox-2.0.2.log 
257
$ grep -c '^FILTERED' /tmp/pdfbox-2.0.2.log 
257
$ grep -c '^File:' /tmp/pdfbox-2.0.24.log 
257
$ grep -c '^FILTERED' /tmp/pdfbox-2.0.24.log 
257
```

I can't speak to the quality or contents of the filtered text, but the number of items filtered without error is the same in each version. So looks good to me! :+1: 

@nwoodward @J4bbi note that we likely need to update fontbox to 2.0.24 as well, as it is provided by the pdfbox project and uses the same release scheme.

nwoodward:
@alanorth My thoughts exactly. As this is a patch release, we're likely pulling in lots of bug fixes and improvements without much concern for incompatible changes.

J4bbi (approved):

alanorth:
@nwoodward wondering if we shouldn't bump this again before the DSpace 6.4 release since this PR was merged ages ago. There are metric tons of updates on the PDFBox 2.0.x branch:

- 2.0.25 release notes: https://issues.apache.org/jira/secure/ReleaseNote.jspa?projectId=12310760&version=12350196
- 2.0.26 release notes: https://issues.apache.org/jira/secure/ReleaseNote.jspa?projectId=12310760&version=12350925

nwoodward:
@alanorth I definitely agree that a newer patch release would pull in some improvements. The reason I went with 2.0.24 is because it's what was in main, and that's still the case today (https://github.com/DSpace/DSpace/blob/main/pom.xml#L42). I could submit PRs to update it in both branches, though I can probably only test it out on a DSpace 6 instance with content.

alanorth:
In terms of DSpace 6.4-SNAPSHOT it just kills me that we bumped the version of so many dependencies years ago and then never released it. Meanwhile the upstream projects have made point releases patching bugs, security issues, etc. :cry:

I have a huge DSpace 6 repository full of content. I could test an updated pdfbox if you made a PR. Then we just let Tim come along and say "port to main" ;)

PR #2743, opened 2020-04-10
DS-4475 Add support for Controlled Vocabulary

Micheleboychuk:
## References
https://jira.lyrasis.org/browse/DS-4475
https://github.com/DSpace/Rest7Contract/pull/120

## Description
The PR introduce a new controller, unfortunately the search methods for the authorityEntry cannot be really implemented as search methods in the repository class as the authorityEntry are not normal rest object with their own endpoint but instead "nested resources" inside the authority.

To keep the code as uniform as possible we have introduced a generic controller for authorityEntry able to expose "these nested search methods" that are so annotated in the link repository class "as usual".

## Instructions for Reviewers
The default configuration contains the srsc authority that can be used for testing the new endpoints. Please check the IT or the REST contract PR for the relevant changes to verify.

## Checklist
- **N/A** If my PR includes new, third-party dependencies (in any `pom.xml`), I've made sure their licenses align with the [DSpace BSD License](https://github.com/DSpace/DSpace/blob/master/LICENSE) based on the [Licensing of Contributions](https://wiki.lyrasis.org/display/DSPACE/Code+Contribution+Guidelines#CodeContributionGuidelines-LicensingofContributions) documentation.
- [x] If my PR modifies the REST API, I've linked to the REST Contract page (or open PR https://github.com/DSpace/Rest7Contract/pull/120) related to this change.

### DIFF
--- a/dspace-server-webapp/src/test/java/org/dspace/app/rest/AuthorityRestRepositoryIT.java
+++ b/dspace-server-webapp/src/test/java/org/dspace/app/rest/AuthorityRestRepositoryIT.java
@@ -359,6 +359,17 @@ public class AuthorityRestRepositoryIT extends AbstractControllerIntegrationTest
                              .andExpect(jsonPath("$.page.totalElements", Matchers.is(0)));
     }
 
+    @Test
+    public void srscSearchTest() throws Exception {
+        String tokenAdmin = getAuthToken(admin.getEmail(), password);
+        getClient(tokenAdmin).perform(get("/api/integration/authorities/srsc/entries/search"))
+                             .andExpect(status().isOk())
+                             .andExpect(jsonPath("$._links.byParent.href", Matchers.containsString(
+                                                 "api/integration/authorities/srsc/entries/search/byParent")))
+                             .andExpect(jsonPath("$._links.top.href", Matchers.containsString(
+                                                 "api/integration/authorities/srsc/entries/search/top")));
+    }
+
     @Override
     public void destroy() throws Exception {
         AuthorityServiceFactory.getInstance().getAuthorityIndexingService().cleanIndex();
--- a/dspace-server-webapp/src/test/java/org/dspace/app/rest/AuthorityRestRepositoryIT.java
+++ b/dspace-server-webapp/src/test/java/org/dspace/app/rest/AuthorityRestRepositoryIT.java
@@ -250,6 +250,35 @@ public class AuthorityRestRepositoryIT extends AbstractControllerIntegrationTest
           .andExpect(jsonPath("$.page.totalElements", is(12)))
           .andExpect(jsonPath("$.page.totalPages", is(3)))
           .andExpect(jsonPath("$.page.number", is(0)));
+
+        //second page
+        getClient(tokenAdmin).perform(get("/api/integration/authorities/srsc/entries/search/top")
+                 .param("page", "1")
+                 .param("size", "5"))
+           .andExpect(status().isOk())
+           .andExpect(jsonPath("$._embedded.authorityEntries", Matchers.containsInAnyOrder(
+               AuthorityEntryMatcher.matchAuthority("SCB16", "TECHNOLOGY"),
+               AuthorityEntryMatcher.matchAuthority("SCB17", "FORESTRY, AGRICULTURAL SCIENCES and LANDSCAPE PLANNING"),
+               AuthorityEntryMatcher.matchAuthority("SCB18", "MEDICINE"),
+               AuthorityEntryMatcher.matchAuthority("SCB19", "ODONTOLOGY"),
+               AuthorityEntryMatcher.matchAuthority("SCB21", "PHARMACY")
+               )))
+           .andExpect(jsonPath("$.page.totalElements", is(12)))
+           .andExpect(jsonPath("$.page.totalPages", is(3)))
+           .andExpect(jsonPath("$.page.number", is(1)));
+
+        // third page
+        getClient(tokenAdmin).perform(get("/api/integration/authorities/srsc/entries/search/top")
+                 .param("page", "2")
+                 .param("size", "5"))
+           .andExpect(status().isOk())
+           .andExpect(jsonPath("$._embedded.authorityEntries", Matchers.containsInAnyOrder(
+               AuthorityEntryMatcher.matchAuthority("SCB22", "VETERINARY MEDICINE"),
+               AuthorityEntryMatcher.matchAuthority("SCB23", "INTERDISCIPLINARY RESEARCH AREAS")
+               )))
+           .andExpect(jsonPath("$.page.totalElements", is(12)))
+           .andExpect(jsonPath("$.page.totalPages", is(3)))
+           .andExpect(jsonPath("$.page.number", is(2)));
     }
 
     @Test
@@ -269,6 +298,7 @@ public class AuthorityRestRepositoryIT extends AbstractControllerIntegrationTest
     @Test
     public void srscSearchByParentFirstLevelPaginationTest() throws Exception {
         String tokenAdmin = getAuthToken(admin.getEmail(), password);
+        // first page
         getClient(tokenAdmin).perform(get("/api/integration/authorities/srsc/entries/search/byParent")
                  .param("id", "SCB14")
                  .param("page", "0")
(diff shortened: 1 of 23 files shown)


KevinVdV:
I notice a lot of classes related to the authority framework were adjusted to support the controlled vocabulary. In DSpace 6 the controlled vocabularies didn't use the authority framework at all. It also shouldn’t store anything in the authority metadata. The purpose of this feature previously was just a way to limit the number of values that can be entered in a certain field (same for example how dropdowns in the input forms work).

abollini:
Hi @KevinVdV 
the controlled vocabulary was an alone feature but also an option for the authority framework since long time (at least version 5 https://github.com/DSpace/DSpace/blob/dspace-5_x/dspace-api/src/main/java/org/dspace/content/authority/DSpaceControlledVocabulary.java )

When the submission was developed for DSpace 7 it was decided to deal withdropdown and controlled vocabulary as authority choice to provide a single uniform way to manage them instead than build three separate features. This PR is only about minor tweeks to the authority endpoint to allow a better support for hierarchical authority as the controlled vocabulary

benbosman:
@abollini I've included a list of problems if the DSpace 6 implementation would no longer be supported. I hope this helps clarify the importance of the functionality as it exists in DSpace 6

-   The item metadata is not compatible with the DSpace 6 data. Items submitted in DSpace 6 don't use an authority ID while items with DSpace 7 would suddenly require an authority ID. This has a direct impact on the data being stored and DSpace currently treats authority values different from non-authority values on many parts (think about search, browse, CSV export, ...).
-   The authority ID doesn't add any value, but is useless because it doesn't contain the source. There's no way to link back to the authority entry based on the authority ID. An authority ID should also contain the authority-name (the fact that it is retrieved from the srsc file) to ensure the link to <https://github.com/DSpace/Rest7Contract/blob/master/authorities.md#authority-entry-values> can be built again
-   Authority control configuration has always been based on [DSpace config](https://github.com/DSpace/DSpace/blob/master/dspace/config/dspace.cfg#L1506-L1507), but this config has now been bypassed and authority values are created without the necessary configuration
-   The DSpace 6 submission UI supports [browsing the tree top down](https://www.dropbox.com/s/4dpnjl7mieritj0/Schermafdruk%202020-05-05%2015.44.26.png?dl=1) and [filtering in the tree while still being aware of the tree structure](https://www.dropbox.com/s/t90d4s376203xnz/Schermafdruk%202020-05-05%2015.44.40.png?dl=1). In DSpace 7 neither feature is supported
-   Archived items should still support searching and facets based on the leaf or parent node of the chosen value (including the discovery facets). The topic facet configured at e.g. <https://openknowledge.worldbank.org/discover> will display and filter on "Finance and Financial Sector Development" for items containing e.g. "Finance and Financial Sector Development :: Concessional Finance and Global Partnerships". Also searching for the parent in the plain query field should still work

abollini:
hi @benbosman 
thanks for the summary about the difference between the current DSpace7 approach and one way, probably the most common, to use controlled vocabulary in the previous version.
It would be useful to put these point in one or more ticket as I don't think they are appropriate as feedback for such PR that is just addressing few aspects as detailed in the DS-4475 (if we found the name of the ticket confusing we can fix it).

That said, just few inline feedback to your points

> The item metadata is not compatible with the DSpace 6 data. Items submitted in DSpace 6 don't use an authority ID while items with DSpace 7 would suddenly require an authority ID. This has a direct impact on the data being stored and DSpace currently treats authority values different from non-authority values on many parts (think about search, browse, CSV export, ...).

good point, we should decide if we want to force such transition (my preference) providing migration script or if we should support both way as in the previous versions. Please note that version 5 & 6 also have support for controlled vocabulary via authority and in such case the authority id was stored https://github.com/DSpace/DSpace/blob/dspace-5_x/dspace-api/src/main/java/org/dspace/content/authority/DSpaceControlledVocabulary.java#L195

> The authority ID doesn't add any value, but is useless because it doesn't contain the source. There's no way to link back to the authority entry based on the authority ID. An authority ID should also contain the authority-name (the fact that it is retrieved from the srsc file) to ensure the link to https://github.com/DSpace/Rest7Contract/blob/master/authorities.md#authority-entry-values can be built again
there is nothing new in that, this is how the authority framework works since ever. It could be useful to store the source in the authority ID as well but currently (dspace < 7) to make use of the authority you need to discover which authority has provided it looking to the metadata configuration

> Authority control configuration has always been based on DSpace config, but this config has now been bypassed and authority values are created without the necessary configuration

this is inline with the change done also for value pair since the first developments of the new submission and noted in the configuration changes
https://wiki.lyrasis.org/display/DSPACE/Configuration+changes+in+the+submission+process

> The DSpace 6 submission UI supports browsing the tree top down and filtering in the tree while still being aware of the tree structure. In DSpace 7 neither feature is supported

This will be supported on the Angular side, as noted in the JIRA ticket this is mainly a backport and cleanup of what already exists in DSpace-CRIS 7. If useful I will ask @atarix83 to anticipate screen or make a short demo about the "final result on the angular side"

> Archived items should still support searching and facets based on the leaf or parent node of the chosen value (including the discovery facets). The topic facet configured at e.g. https://openknowledge.worldbank.org/discover will display and filter on "Finance and Financial Sector Development" for items containing e.g. "Finance and Financial Sector Development :: Concessional Finance and Global Partnerships". Also searching for the parent in the plain query field should still work

this require analysis but it should be possible to configure discovery to either ignore the authority or better makes use of it still providing support for the hierarchical facets. The use of an authority ID will allow us to support translation.

Again, I suggest to move these discussion outside this specific PR that have a very much limited scope

dspace-server-webapp/src/test/java/org/dspace/app/rest/AuthorityRestRepositoryIT.java
```diff
@@ -210,6 +211,165 @@ public void retrieveSolrValueTest() throws Exception {
...
+        String tokenAdmin = getAuthToken(admin.getEmail(), password);
+        getClient(tokenAdmin).perform(get("/api/integration/authorities/srsc/entries/search/byParent")
+                             .param("id", "WRONG_ID"))
+                             .andExpect(status().isOk())
+                             .andExpect(jsonPath("$.page.totalElements", Matchers.is(0)));
+    }
+
+    @Test
+    public void srscSearchTest() throws Exception {
+        String tokenAdmin = getAuthToken(admin.getEmail(), password);
```
tdonohue, line 364: It looks like all these tests currently only run as an Admin user. We'd need tests that also run as a normal authenticated user (which should have permissions too), and also tests to verify that an anonymous user has no access.

tdonohue (commented):
@Micheleboychuk and @abollini : I gave this a quick code review today just to give some feedback on the code itself (obviously we'll discuss the broader topics in tomorrow's meeting).  I've found a lot of missing JavaDocs & tests.  There's also areas of the code I don't quite understand (see inline questions).

Finally, I think this PR needs rebasing on `master` as it looks like it uses code that was valid for an older version of Spring HATEOAS.  After merging #2720 we are now on Spring HATEOAS v1.0, which means this PR currently doesn't compile if I try and pull it into a local copy of `master`.  So, I cannot test it at this time.

benbosman (changes requested):
Hi @abollini,

If desired, this discussion can move a ticket, assuming of course this PR is closed or placed on hold until it's sorted out.
The discussion points are definitely important for this PR because the PR seems to be adding new functionality (different from the DSpace 6 functionality) while the normal DSpace 6 behavior is not implemented yet and you seem to assume it's no longer applicable. The tree browse has never been part of the (rarely used) authority control version of the taxonomies. Assuming the taxonomies can be migrated to authority control without support for the original functionality is of course lacking any backwards compatibility

The same and even more problems apply for the dropdowns you're referring to, for which I've created a [ticket](https://jira.lyrasis.org/browse/DS-4498) already. I did not involve it here because this PR doesn't alter the dropdowns. That discussion should definitely remain part of the ticket.

atarix83:
>  If useful I will ask @atarix83 to anticipate screen or make a short demo about the "final result on the angular side"

Here some of screenshot that should help understand the final result
Schermata da 2020-05-07 15-25-19
Schermata da 2020-05-07 15-18-53
Schermata da 2020-05-07 15-26-25

tdonohue:
I'm flagging this as `work in progress` as this PR either requires updating or replacement based on the new REST Contract at https://github.com/DSpace/Rest7Contract/pull/128.  Until then this is unreviewable.

If this PR should just be closed for now, that's fine too. Just let me know, and we can use the ticket to track progress on this work.

abollini:
replaced by https://github.com/DSpace/DSpace/pull/2792

PR #2755, opened 2020-05-04
[DS-3940] New SHERPA - fix for incorrect metadata (publisher policy) URI

kshepherd:
## References
Fix to recently merged #2739 PR - after the change to handle multiple responses in JSPUI, the metadata (policy) URI needed updating to be in the correct scope
Original JIRA: https://jira.lyrasis.org/browse/DS-3940


nwoodward (approved):
:+1: Tested and confirmed it works. The journal policy URL is correct now.

PR #2758, opened 2020-05-07
Shibboleth Single Logout support

### LINKED ISSUE #7799: [DS-4464] Shibboleth Single Logout (SLO) support
<i>Imported from JIRA <a href="https://jira.lyrasis.org/browse/DS-4464">[DS-4464]</a> created by paulo_graca</i>
<p>The DSpace 6 and lower implementations don't support CAS Logout (or Single LogOut). We use a national identification service, that also integrates with a government id service, and upon logout, all sessions should be terminated (across all service providers).</p>

<p>As referred in <a href="https://github.com/DSpace/DSpace/pull/2651#issuecomment-585816548" class="external-link" target="_blank" rel="nofollow noopener">https://github.com/DSpace/DSpace/pull/2651#issuecomment-585816548</a> there is this need. In the PR there are also some relevant comments from <a href="https://jira.lyrasis.org/secure/ViewProfile.jspa?name=benbosman" class="user-hover" rel="benbosman">Ben Bosman</a> and <a href="https://jira.lyrasis.org/secure/ViewProfile.jspa?name=mwood" class="user-hover" rel="mwood">Mark H. Wood</a> of how this feature could be working. Mark, if I understood correctly, proposes two different sign outs, one locally and the other CAS specific. And Ben somehow proposes the same, by having the presence of a specific configuration to control if CAS logout is enabled or not.</p>

<p> </p>

<p>This is what we (RCAAP from FCT|FCCN) did to have it working on DSpace repositories:</p>

<p>Basically, in terms of the Service Provider, we use the shibboleth's notify parameter to configure the IdP logout url:<br/>
 <a href="https://wiki.shibboleth.net/confluence/display/SP3/Notify" class="external-link" target="_blank" rel="nofollow noopener">https://wiki.shibboleth.net/confluence/display/SP3/Notify</a></p>

<p><tt><Notify Channel="front" Location="https://[MY-SERVER]/logout" /></tt></p>

<p>When the logout event in DSpace is performed, then the Notify is triggered.<br/>
 Some exchanges will occur between the SP and the IdP and an indication to IdP to clean the shibboleth authentication will also occur.</p>

<p>In our case we will have something like this: <tt>[https://[MY-SERVER]/shibboleth.sso/Logout?notifying=1&index=1|https://[my-server]/shibboleth.sso/Logout?notifying=1&index=1]</tt></p>

<p>To invoke the logout. We also clean local cookies too.</p>

paulo-graca:
## References
* https://jira.lyrasis.org/browse/DS-4464 
* Fixes #7799
* https://github.com/DSpace/Rest7Contract/pull/124

## Description
This PR intends to add Shibboleth Single Logout (SLO) support to DS7 based at the frontend channel (using notify). For more info about SLO, and front channel please check: https://wiki.shibboleth.net/confluence/display/SHIB2/NativeSPNotify

## Instructions for Reviewers
* PLEASE NOTE for testing this you should configure the shibboleth your server and perhaps the IdP in order to have this feature working properly. This implementation depends on those configurations. You have more documentation about this on https://wiki.shibboleth.net. In our particular case we added <Notify Channel="front" Location="https://[MY-SERVER]/server/auth/logout" /> on our Service Provider to use the front channel.
* It introduces the concept of LogOutPage for all authentication methods as already exists for loginpage
* We choose to reuse the current /authn/logout controller, since we consider the logout process should be only one. We added support for extra required params (action and return)
* Shibboleth authentication was also changed to retrieve a page that can be configured
* For shibboleth it was created a specific configuration url `authentication-shibboleth.lazysession.logouturl`. If this configuration is defined, then it will be used for the logout page (for testing purposes I suggest to add it to your local.cfg)
* A new logout success filter was created to support the logout page redirection


gfernandez598:
> This PR needs to be refactored. Meanwhile, authentication suffered several changes that weren't reflected here.

I was testing the PR within DSpace 7.5 and actually I am using it in 7.6. A basic integration with Shibboleth SP 3 works.
Now I am trying to improve DSpace UI Login because it breaks at logout event.

In case it's helpful, I attach the changes that I apply to the respective DSpace versions.
[patch-2758.zip](https://github.com/DSpace/DSpace/files/13044706/patch-2758.zip)

Successes!

PR #2761, opened 2020-05-18
DS-4506 Add a limit on the maximum value allowed for the parameter rpp (resuts per page) in browse and search pages.

### LINKED ISSUE #7840: [DS-4506] There is no max limit in rpp parameter of OpenSearchController::search
<i>Imported from JIRA <a href="https://jira.lyrasis.org/browse/DS-4506">[DS-4506]</a> created by santit96</i>
<p>Users can set this parameter to the value they want, both in search and browse pages. So, they could retrieve all items in the repository in a query if a large value is set. This may cause a DOS if an intentional attack is performed.</p>

santit96:
## References
*  [DS-4506](https://jira.lyrasis.org/browse/DS-4506)
* Fixes #7840

## Description
Add a limit on the maximum value allowed for the parameter rpp p (resuts per page) in browse and search pages.

## Instructions for Reviewers
* In solr, a variable MAX_RESULTS_ALLOWED to control the max rpp value was created.
* In xmlui browse page now there is a check for the entered value is not greater than the maximum value shown in the select used to set the parameter.

* Now if a user set the rpp with a value greater than the maximum value allowed, then those changes are not taken into account and the parameter is set with the max value allowed

### DIFF
--- a/dspace-api/src/main/java/org/dspace/discovery/SolrServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/discovery/SolrServiceImpl.java
@@ -110,6 +110,8 @@ public class SolrServiceImpl implements SearchService, IndexingService {
 
     public static final String VARIANTS_STORE_SEPARATOR = "###";
 
+    public static final int MAX_RESULTS_ALLOWED = 100;
+
     @Autowired(required = true)
     protected ContentServiceFactory contentServiceFactory;
     @Autowired(required = true)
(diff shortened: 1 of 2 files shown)


alanorth:
This is a good catch. It doesn't seem to work for Discovery "view more" (search-filter) links in XMLUI?

- http://localhost:8080/search-filter?field=author&filterorder=COUNT → returns authors 1–10
- http://localhost:8080/search-filter?field=author&filterorder=COUNT&rpp=15000 → returns authors 1–15000

I don't use browse indexes so I haven't tested that.

P.S. this probably requires a documentation update if it is merged.

PR #2771, opened 2020-05-25
DS-4504 Performance issues with finding Collections a user can submit to

Micheleboychuk:
## References
https://jira.lyrasis.org/browse/DS-4504
restContract: https://github.com/DSpace/Rest7Contract/pull/127
## Description
In this PR has been
1)  created SolrServiceIndexCollectionSubmittersPlugin to add information in the solr document 
     related to a collection about which epersons/groups  have ADD permissions on it
2) Updated methods findAuthorizedByCommunity and findAuthorized to use solr query
3) created some IT to proof that the implementation is robust and work properly also with 
    community and collection admins and with a deeper hierarchy
4) small updates to restContract
## Checklist

### DIFF
--- a/dspace-api/src/main/java/org/dspace/content/CollectionServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/content/CollectionServiceImpl.java
@@ -17,11 +17,13 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.MissingResourceException;
+import java.util.Set;
 import java.util.UUID;
 
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.logging.log4j.Logger;
+import org.apache.solr.client.solrj.util.ClientUtils;
 import org.dspace.app.util.AuthorizeUtil;
 import org.dspace.authorize.AuthorizeConfiguration;
 import org.dspace.authorize.AuthorizeException;
@@ -40,6 +42,13 @@ import org.dspace.core.Context;
 import org.dspace.core.I18nUtil;
 import org.dspace.core.LogManager;
 import org.dspace.core.service.LicenseService;
+import org.dspace.discovery.DiscoverQuery;
+import org.dspace.discovery.DiscoverResult;
+import org.dspace.discovery.IndexableObject;
+import org.dspace.discovery.SearchService;
+import org.dspace.discovery.SearchServiceException;
+import org.dspace.discovery.indexobject.IndexableCollection;
+import org.dspace.eperson.EPerson;
 import org.dspace.eperson.Group;
 import org.dspace.eperson.service.GroupService;
 import org.dspace.eperson.service.SubscribeService;
@@ -99,6 +108,9 @@ public class CollectionServiceImpl extends DSpaceObjectServiceImpl<Collection> i
 
     @Autowired(required = true)
     protected CollectionRoleService collectionRoleService;
+
+    @Autowired(required = true)
+    protected SearchService searchService;
 
     protected CollectionServiceImpl() {
         super();
@@ -907,4 +919,66 @@ public class CollectionServiceImpl extends DSpaceObjectServiceImpl<Collection> i
         return role;
     }
 
+    @Override
+    public List<Collection> findAuthorizedCollectionsInSOLR(String q, Context context, Community community,
+        int offset, int limit) throws SQLException, SearchServiceException {
+
+        List<Collection> collections = new ArrayList<Collection>();
+        StringBuilder query = new StringBuilder();
+        DiscoverQuery discoverQuery = new DiscoverQuery();
+        discoverQuery.setDSpaceObjectFilter(IndexableCollection.TYPE);
+        discoverQuery.setStart(offset);
+        discoverQuery.setMaxResults(limit);
+        DiscoverResult resp = resultSolrQuery(context, query, discoverQuery,community, q);
+        for (IndexableObject solrCollections : resp.getIndexableObjects()) {
+            Collection c = ((IndexableCollection) solrCollections).getIndexedObject();
+            collections.add(c);
+        }
+        return collections;
+    }
+
+    @Override
+    public int countAuthorizedCollectionsInSOLR(String q, Context context, Community community)
+        throws SQLException, SearchServiceException {
+
+        StringBuilder query = new StringBuilder();
+        DiscoverQuery discoverQuery = new DiscoverQuery();
+        discoverQuery.setDSpaceObjectFilter(IndexableCollection.TYPE);
+        DiscoverResult resp = resultSolrQuery(context, query, discoverQuery,community,q);
+        return (int)resp.getTotalSearchResults();
+    }
+
+    private DiscoverResult resultSolrQuery(Context context, StringBuilder query, DiscoverQuery discoverQuery,
+                                   Community community, String q) throws SQLException, SearchServiceException {
+
+        EPerson currentUser = context.getCurrentUser();
+        if (!authorizeService.isAdmin(context)) {
+            Group anonymousGroup = groupService.findByName(context, Group.ANONYMOUS);
+            String anonGroupId = "";
+            if (anonymousGroup != null) {
+                anonGroupId = anonymousGroup.getID().toString();
+            }
+            query.append("submit:(g").append(anonGroupId);
+            if (currentUser != null) {
+                query.append(" OR e").append(currentUser.getID());
+            }
+            Set<Group> groups = groupService.allMemberGroupsSet(context, currentUser);
+            for (Group group : groups) {
+                query.append(" OR g").append(group.getID());
+            }
+            query.append(")");
+            discoverQuery.addFilterQueries(query.toString());
+        }
+        if (community != null) {
+            discoverQuery.addFilterQueries("location.comm:" + community.getID().toString());
+        }
+        if (StringUtils.isNotBlank(q)) {
+            StringBuilder buildQuery = new StringBuilder();
+            String escapedQuery = ClientUtils.escapeQueryChars(q);
+            buildQuery.append(escapedQuery).append(" OR ").append(escapedQuery).append("*");
+            discoverQuery.setQuery(buildQuery.toString());
+        }
+        DiscoverResult resp = searchService.search(context, discoverQuery);
+        return resp;
+    }
 }
--- a/dspace-api/src/main/java/org/dspace/content/CollectionServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/content/CollectionServiceImpl.java
@@ -924,12 +924,11 @@ public class CollectionServiceImpl extends DSpaceObjectServiceImpl<Collection> i
         int offset, int limit) throws SQLException, SearchServiceException {
 
         List<Collection> collections = new ArrayList<Collection>();
-        StringBuilder query = new StringBuilder();
         DiscoverQuery discoverQuery = new DiscoverQuery();
         discoverQuery.setDSpaceObjectFilter(IndexableCollection.TYPE);
         discoverQuery.setStart(offset);
         discoverQuery.setMaxResults(limit);
-        DiscoverResult resp = resultSolrQuery(context, query, discoverQuery,community, q);
+        DiscoverResult resp = retrieveAuthorizedCollections(context, discoverQuery,community, q);
         for (IndexableObject solrCollections : resp.getIndexableObjects()) {
             Collection c = ((IndexableCollection) solrCollections).getIndexedObject();
             collections.add(c);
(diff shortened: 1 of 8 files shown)


dspace-server-webapp/src/main/java/org/dspace/app/rest/repository/CollectionRestRepository.java
```diff
@@ -174,19 +178,70 @@ public CollectionRest findOne(Context context, UUID id) {
...
+                Collection c = ((IndexableCollection) solrCollections).getIndexedObject();
+                collections.add(c);
+            }
+            return converter.toRestPage(collections, pageable, tot , utils.obtainProjection());
+        } catch (SQLException | SearchServiceException e) {
             throw new RuntimeException(e.getMessage(), e);
         }
     }
 
+    private DiscoverResult discoverAuthorizedCollections(Pageable pageable, String q, Context context, Community com)
```
tdonohue, line 194: This entire method seems like we may want to consider moving it into the `dspace-api`.  It's a bit odd to me to have a RestRepository class building detailed Solr queries, which makes me wonder if we need to consider moving this into the `CollectionIndexFactoryImpl` or somewhere similar (and then just calling that new method from this RestRepository class)?

dspace/solr/search/conf/schema.xml
```diff
@@ -259,6 +259,8 @@
     <field name="taskfor" type="string" indexed="true" stored="true" omitNorms="true" multiValued="true" docValues="true" />
    
     <field name="has_content_in_original_bundle" type="string" indexed="true" stored="true" omitNorms="true" multiValued="true" docValues="true" />
+
+    <field name="submit" type="string" indexed="true" stored="true" omitNorms="true" multiValued="true" docValues="true" />
```
tdonohue, line 263: We may want to add a comment here to explain the difference between "submit" and the existing "submitter" index (as it may be confusing in the future).  Obviously this index is to track *who has submit permissions*, while "submitter" seems to be *who submitted an item*.  They are similarly named, so we should add a comment here to describe them.

tdonohue (commented):
@Micheleboychuk and @abollini : Gave this a really quick code review (haven't tested yet) and added a few minor comments inline below.  Mostly, the code looks good though.

dspace-server-webapp/src/main/java/org/dspace/app/rest/repository/CollectionRestRepository.java
```diff
@@ -165,7 +168,8 @@ public CollectionRest findOne(Context context, UUID id) {
 
     @SearchRestMethod(name = "findAuthorizedByCommunity")
     public Page<CollectionRest> findAuthorizedByCommunity(
-        @Parameter(value = "uuid", required = true) UUID communityUuid, Pageable pageable) {
+        @Parameter(value = "uuid", required = true) UUID communityUuid, Pageable pageable,
+        @Parameter(value = "query") String q) {
```
KevinVdV, line 172: What is the use case for this new "query" parameter ? Because it isn't in the rest contract: https://github.com/DSpace/Rest7Contract/blob/733d81061f1dbcd3049ed0d80130746418a4b3a1/collections.md#findauthorizedbycommunity

KevinVdV (changes requested):
Did a quick review of the code, didn't find any issues (except for the things that Tim already noted), but I do have a question about a new parameter that is now included.

But I might have found some other pottential issues, more with the design than with the code:

- Any changes made to resourcepolicies won't trigger modify events so the discovery won't notice changes on a collection & thus won't reindex it.
- If you add / remove an administrator group to a community that has collections the SolrServiceIndexCollectionSubmittersPlugin class won't pick up on these changes (because of the issue mentioned above but also modify events will only force a reindex on the community that is changed, not in the underlying collections)

dspace-server-webapp/src/main/java/org/dspace/app/rest/repository/CollectionRestRepository.java
abollini: thanks to point to the rest contract. Indeed we need to create a PR to document this change. The need for a query parameter was noted and agreed in the discussion about how to address the performance issue https://wiki.lyrasis.org/display/DSPACE/Collection+dropdown+in+submission

dspace-server-webapp/src/main/java/org/dspace/app/rest/repository/CollectionRestRepository.java
abollini: you are right. Should we move it to the `CollectionService`? should we name it findAuthorizedCollectionsInSOLR

abollini: @Micheleboychuk please refactor the code adding the method in the CollectionService as by comment above. We have discussed that briefly in yesterday meeting and @tdonohue was just fine to move it in the dspace-api layer

abollini:
@KevinVdV @tdonohue the consideration about direct changes to the resourcepolicies or the post creation/removal of community admin groups are right. They are a bit "edge" situation that not occur in normal workflow (usually there is no direct use of resourcepolicy to set the submitter group but instead the UI and the default "COLLECTION_x_SUBMITTER" group is used.
Also the creation/removal of a community admin group is not something that happen ofter on a live repository, these are organisation decision that usually are taken in the early stage of the project and revisited at specific time but not frequently.

We have already a similar discussion open https://github.com/DSpace/Rest7Contract/pull/116
that was started from this comment https://github.com/DSpace/DSpace/pull/2707#discussion_r393832946
Moreover, please note that we also have the same "issue" in dspace 6 (and in the master) about the search of restricted items. As only the read policy of the items are processed this mean that
- if the policy are manipulated directly the item will be not reindexed
- admin permission on collection/community and the eventual changes to these groups are not taken in consideration

For such reason I would like to document this additional use case together with the other in a single separated ticket

dspace-api/src/main/java/org/dspace/content/CollectionServiceImpl.java
```diff
@@ -907,4 +919,66 @@ public Group createDefaultReadGroup(Context context, Collection collection, Stri
         return role;
     }
 
+    @Override
+    public List<Collection> findAuthorizedCollectionsInSOLR(String q, Context context, Community community,
+        int offset, int limit) throws SQLException, SearchServiceException {
+
+        List<Collection> collections = new ArrayList<Collection>();
+        StringBuilder query = new StringBuilder();
```
KevinVdV, line 927: This StringBuilder object is created here, but isn't used until the resultSolrQuery() method. Can you move the creation of the variable to the resultSolrQuery() method.

dspace-api/src/main/java/org/dspace/content/CollectionServiceImpl.java
```diff
@@ -907,4 +919,66 @@ public Group createDefaultReadGroup(Context context, Collection collection, Stri
...
+            collections.add(c);
+        }
+        return collections;
+    }
+
+    @Override
+    public int countAuthorizedCollectionsInSOLR(String q, Context context, Community community)
+        throws SQLException, SearchServiceException {
+
+        StringBuilder query = new StringBuilder();
```
KevinVdV, line 944: Same as above, move the variable to the resultSolrQuery() method.

dspace-api/src/main/java/org/dspace/content/CollectionServiceImpl.java
```diff
@@ -907,4 +919,66 @@ public Group createDefaultReadGroup(Context context, Collection collection, Stri
...
+        DiscoverQuery discoverQuery = new DiscoverQuery();
```
KevinVdV, line 945: As we only need the total results for this one can you call on:
`discoverQuery.setMaxResults(0);`

Because the default here is 10 & we don't need to retrieve these objects from the database if we need the total count.

dspace-api/src/main/java/org/dspace/content/CollectionServiceImpl.java
```diff
@@ -907,4 +919,66 @@ public Group createDefaultReadGroup(Context context, Collection collection, Stri
...
+        discoverQuery.setDSpaceObjectFilter(IndexableCollection.TYPE);
+        DiscoverResult resp = resultSolrQuery(context, query, discoverQuery,community,q);
+        return (int)resp.getTotalSearchResults();
+    }
+
+    private DiscoverResult resultSolrQuery(Context context, StringBuilder query, DiscoverQuery discoverQuery,
```
KevinVdV, line 951: Can we rename the method name here ? This method is used specifically to retrieve authorized collections a user can submit in.

dspace-api/src/main/java/org/dspace/content/CollectionServiceImpl.java
```diff
@@ -907,4 +919,66 @@ public Group createDefaultReadGroup(Context context, Collection collection, Stri
...
+                                   Community community, String q) throws SQLException, SearchServiceException {
+
+        EPerson currentUser = context.getCurrentUser();
+        if (!authorizeService.isAdmin(context)) {
+            Group anonymousGroup = groupService.findByName(context, Group.ANONYMOUS);
```
KevinVdV, line 956: Can we leave out the call to the anonymousGroup group here ? Anonymous will never be granted submit rights (as you always need a logged in user), so this database call is not needed.

KevinVdV (changes requested):
Made another review of the code, some small things that require cleanup. Could we perhaps also add an additional test where we don't assign the user straight in the submitter group, but rather add him to a group that is present in the submitter group. So you would have the following setup:
- SubmitterGroup
- ChildGroupOfSubmitterGroup
- User (member in the child group)

abollini:
in yesterday meeting we have agreed to deal with the discussion about solr in a different issue/PR, I have created the ticket https://jira.lyrasis.org/browse/DS-4524 for that.
The other feedback in this PR have been addressed, I would appreciate a re-review @KevinVdV  @tdonohue

dspace-server-webapp/src/main/java/org/dspace/app/rest/repository/CollectionRestRepository.java
```diff
@@ -174,19 +175,26 @@ public CollectionRest findOne(Context context, UUID id) {
...
+                                              Math.toIntExact(pageable.getOffset()),
+                                              Math.toIntExact(pageable.getOffset() + pageable.getPageSize()));
+            int tot = cs.countAuthorizedCollectionsInSOLR(q, context, com);
+            return converter.toRestPage(collections, pageable, tot , utils.obtainProjection());
+        } catch (SQLException | SearchServiceException e) {
             throw new RuntimeException(e.getMessage(), e);
         }
     }
 
     @SearchRestMethod(name = "findAuthorized")
```
tdonohue, line 188: I'm just now realizing this `@SearchRestMethod` seems mis-named as well.  We are *not* finding all Collections you are authorized to Read or edit or similar.  This is very specific to collections you can *submit to*.  Shouldn't this be called `findSubmitAuthorized`  or `findAuthorizedForSubmit`?

tdonohue (commented):
@Micheleboychuk and @abollini : Overall, this looks good code-wise.  However, I have some minor requests to cleanup the method names / terminology.  See inline below.  I've realized in re-reviewing this that we may want to make the method names more specific....they often say "authorized" (which could mean any level of authorization), when we are specifically looking only for **submit** authorization (not read or edit or admin).

dspace-server-webapp/src/main/java/org/dspace/app/rest/repository/CollectionRestRepository.java
abollini: https://github.com/DSpace/Rest7Contract/pull/127

dspace-api/src/main/java/org/dspace/content/service/CollectionService.java
```diff
@@ -354,4 +356,34 @@ public void canEdit(Context context, Collection collection, boolean useInheritan
...
+     *                         an autosuggest feature over the collection name
+     * @param context          DSpace Context
+     * @param community        parent community
+     * @param offset           the position of the first result to return
+     * @param limit            paging limit
+     * @return                 discovery search result objects
+     * @throws SQLException              if something goes wrong
+     * @throws SearchServiceException    if search error
+     */
+    public List<Collection> findAuthorizedCollectionsInSOLR(String q, Context context, Community community,
```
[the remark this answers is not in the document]
abollini: these renaming have been applied also to the rest contract, see https://github.com/DSpace/Rest7Contract/pull/127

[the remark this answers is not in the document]
abollini: As there are already IT that proof that the whole stack and so the SolrPlugin works as expected we would like to have that merged as is. We have tried to add some unit test in `CollectionTest` but it is more complicated than expected as the dspace-api doesn't support yet test in SOLR. This was also noted in this PR https://github.com/DSpace/DSpace/pull/2744  so we suggest to merge this PR as is, solve the issue with the solr testing in and later (https://jira.lyrasis.org/browse/DS-4526 ) add these extra test in the CollectionTest class

tdonohue (approved):
👍 This now looks good to me.  I haven't tested it manually, but the ITs seem to be testing that all is working as expected.  I'm OK with leaving out the Unit Tests for now as the ITs (in `dspace-server-webapp`) are performing the same tests we'd expect out of Unit Tests at the `dspace-api` layer

tdonohue:
@KevinVdV : Can you re-review this?  It looks to me like your initial feedback has all been addressed.  If you can review this week, great, otherwise I'll merge as-is on Friday (as this looks good enough to me)

KevinVdV (approved):
Performed another review everything looks good to me !

PR #2867, opened 2020-07-17
DS-4527: Authorizations endpoint and X-On-Behalf-Of headers

### LINKED ISSUE #2810: DS-4527: Authorizations endpoint should also recognize Authorization and X-On-Behalf-Of headers
This is a duplicate of [DS-4527](https://jira.lyrasis.org/browse/DS-4527)

Estimate from release plan: TBD -- needs further investigation
Expressing interest: TBD

Per discussion in Slack on June 15, 2020 between Art Lowel (Atmire), Andrea Bollini (4Science) and myself.

Currently, the /api/authz/authorizations/search/object endpoint returns a list of matching authorizations only for the individual specified as the "eperson" parameter (or anonymous users if the "eperson" parameter is missing).

This would mean that for the client (e.g. Angular UI) to check the permissions of the currently logged in user, it would need to ensure the current eperson's UUID is passed as a param to this method. While that is not complex, it's duplicative as the current user information already exists in the Authorization header passed on every request (after logged in).

Ideally, this endpoint should default to reading current user information from the Authorization and/or X-on-behalf-of (for "log in as" functionality) headers, so that the client only needs to use the "eperson" parameter if a different user's permissions need to be retrieved.

It should be noted that this would change current behavior, as currently a missing "eperson" param means that the backend should return anonymous access permissions. Because of this, we may want to pass a eperson=anonymous parameter for this scenario.

The overall suggested logic is as follows:

Use eperson param if it exists (highest priority). A value of eperson=anonymous can be used to check anonymous permissions while logged in.
Use X-on-behalf-of header if exists & no eperson param
Use Authorization header if exists, and none of the above
If nothing above exists, then no one is logged in, so it'd be an anonymous user asking about anonymous permissions. (Or we could choose to throw an error here and only support anonymous queries by using eperson=anonymous)
This effort is not yet estimated & may need further investigation before implementation

benbosman:
Authorizations endpoint should also recognize Authorization and X-On-Behalf-Of headers

## References
* Link to [JIRA](https://jira.lyrasis.org/browse/DS-4527) ticket
* Fixes #2810 

## Description
This fixes two problems with the Authorizations endpoint:
* The eperson parameter is no longer required, it will use the eperson from the authentication header
* If the X-On-Behalf-Of header is present, it will use the eperson from that header

## Instructions for Reviewers
Best to review this by requesting whether the user has authorization without specifying the eperson parameter.
Make sure to also verify it using the log-in-as functionality

## Checklist

### DIFF
--- a/dspace-server-webapp/src/main/java/org/dspace/app/rest/repository/AuthorizationRestRepository.java
+++ b/dspace-server-webapp/src/main/java/org/dspace/app/rest/repository/AuthorizationRestRepository.java
@@ -242,25 +242,27 @@ public class AuthorizationRestRepository extends DSpaceRestRepository<Authorizat
      */
     private EPerson getUserFromRequestParameter(Context context, UUID epersonUuid)
             throws AuthorizeException, SQLException {
+
         EPerson currUser = context.getCurrentUser();
-        EPerson user = currUser;
-        if (epersonUuid != null) {
+
+        if (epersonUuid == null) {
+            // no user is specified in the request parameters, check the permissions for the current user
+            return currUser;
+
+        } else {
+            // a user is specified in the request parameters
+
             if (currUser == null) {
                 throw new AuthorizeException("attempt to anonymously access the authorization of the eperson "
                         + epersonUuid);
-            } else {
-                // an user is specified in the request parameters
-                if (!authorizeService.isAdmin(context) && !epersonUuid.equals(currUser.getID())) {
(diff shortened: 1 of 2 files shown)


tdonohue (approved):
👍 This looks good to me. I've tested authentication on both the backend & frontend using this PR & everything still works.  Also tested impersonating another user and that still works.  The updated ITs here look good & the actual code changes are less than 10 lines

PR #2896, opened 2020-07-21
Improve indexing performance (uncache relationship metadata entities)

aprilherron:
## References
* [JIRA DS-4425](https://jira.lyrasis.org/browse/DS-4425): Indexing Performance is very slow since refactor. ITs take twice as long

## Description
Indexing in DSpace 7 (particularly in repositories with large numbers of relationships) has gotten increasingly slow. In addition, processing becomes increasingly slower as indexing continues.

After investigating the cause of this, it was discovered that, for every item, all of its relationships _and_ related entities were kept in the hibernate cache for the rest of the job. To fix this issue, relationship entities and the related entity (but not the invoked entity) are uncached after the respective `RelationshipMetadataValue` has been created.

Note: This should also improve performance for any repeated calls to `itemService.getMetadata()`.

## Instructions for Reviewers
1. Test repo with adequate number of items/relationships
2. Test indexing performance
3. Apply patch
3. Repeat indexing test performance

List of changes in this PR:
* Relationship entities and related DSO entities are properly uncached after use in `getRelationshipMetadata()`

## Checklist

### DIFF
--- a/dspace-api/src/main/java/org/dspace/content/RelationshipMetadataServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/content/RelationshipMetadataServiceImpl.java
@@ -51,6 +51,12 @@ public class RelationshipMetadataServiceImpl implements RelationshipMetadataServ
                     fullMetadataValueList
                         .addAll(findRelationshipMetadataValueForItemRelationship(context, item, entityType,
                                 relationship, enableVirtualMetadata));
+                    if (item.getID().equals(relationship.getLeftItem().getID())) {
+                        context.uncacheEntity(relationship.getRightItem());
+                    } else {
+                        context.uncacheEntity(relationship.getLeftItem());
+                    }
+                    context.uncacheEntity(relationship);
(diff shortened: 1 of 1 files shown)


mwoodiupui:
Another example showing that Context has the wrong lifetime for bulk database updates.  We need a way to create additional `Session`s that can be used to operate on objects with short-lived interest during long-lived procedures, discarding them quickly while the overall procedure plods along on a long-lived read-only Session.

cwilper:
Closing; we found this approach has negative side effects that can't easily be avoided. 

In the example we saw, it caused Hibernate to report lazy load errors when iterating through items in a search result, when the items were related to each other. This was due to de-caching both entities when the first item was iterated, then when the second item was iterated, it was missing from Hibernate cache. We could force-reload each item as  we iterate them, but that just creates a new performance problem. And there are likely many other places in the code where entities are iterated like this. Removing both from the cache is just too aggressive a thing to do in the general case.

PR #2897, opened 2020-07-21
#2893 Move builders from dspace-server-webapp to dspace-api.

### LINKED ISSUE #2893: Make some testing infrastructure available to more of our projects.
`dspace-server-webapp` contains some nice Builder classes for testing, which would be useful in other projects.  They should move to `dspace-api` so that they can be shared more widely.

Both projects contain `org.dspace.statistics.MockSolrLoggerServiceImpl` and `org.dspace.solr.MockSolrServer`.  The one in `dspace-server-webapp` spins up an embedded Solr instance while the other does not.  This would be useful in both places (and maybe more).  Plus, it's hard to understand what is going on with the identically-named classes without pondering some obscure classpath magic.  These should be consolidated in `dspace-api` so that a single fully-featured pair can be shared.

mwoodiupui:
## References
#2893

## Description
Move DSO Builders from dspace-server-webapp to dspace-api and share.

## Instructions for Reviewers

List of changes in this PR:
* Moved most Builders to dspace-api.  REST-specific code was not moved.
* **NOTE**:  most Builders moved from package `org.dspace.app.rest.builder` to `org.dspace.builder`.
* Took up some recommendations from code quality tools.
* Renamed an IT from AuthorizeConfigIntegrationTest to AuthorizeConfigIT because the old name made it run with the unit tests.
* **NOTE**:  this changes the default Maven treatment of tests from "do not build or run" (maven.test.skip) to "build but do not run" (skipTests).
* `dspace-api:org.dspace.app.bulkedit.MetadataExportTest` now uses Builder classes instead of creating DSOs "by hand" to demonstrate that the newly-common Builder classes now work in both `dspace-api` and `dspace-server-webapp`.

How to test:
Run all test suites.  They should show the same results as before.

## Checklist


pom.xml
```diff
@@ -475,29 +475,33 @@
...
+                <skipTests>true</skipTests>
             </properties>
         </profile>
 
-        <!-- Skip Integration Tests by default, but allow override on
-             command line by setting property "-DskipITs=false" -->
+        <!-- Skip running Integration Tests by default, but allow override on
+             command line by setting property "-DskipITs=false".
+             ITs are always built, because we have modules that depend
+	     on common test infrastructure. -->
```
tdonohue, line 499: Again, tiny alignment issue cause of a tab on the last line

tdonohue (changes requested):
@mwoodiupui : Looks great overall!  I just had a few minor comments/questions inline (and minor cleanup/fixing to your example of using Builders in `dspace-api`).  Honestly, though this is excellent to see us be able to move Builders up to dspace-api, so thanks for your work on this!

**NOTE TO SELF**: When this gets merged, we'll need to warn the DSpace 7 team to rebase & change the package of all Builders in currently open REST PRs.

dspace-api/src/test/java/org/dspace/app/bulkedit/MetadataExportIT.java
```diff
@@ -14,58 +14,60 @@
...
         context.turnOffAuthorisationSystem();
-        Community community = communityService.create(null, context);
-        Collection collection = collectionService.create(context, community);
-        WorkspaceItem wi = workspaceItemService.create(context, collection, true);
+        Community community = CommunityBuilder.createCommunity(context).build();
+        Collection collection = CollectionBuilder.createCollection(context, community).build();
+        WorkspaceItem wi = WorkspaceItemBuilder.createWorkspaceItem(context, collection).build();
         Item item = wi.getItem();
         itemService.addMetadata(context, item, "dc", "contributor", "author", null, "Donald, Smith");
         item = installItemService.installItem(context, wi);
```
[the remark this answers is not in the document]
mwoodiupui: Thank you.  That's much neater.  I'm not at all familiar with *using* the Builder classes yet....

dspace-api/src/test/java/org/dspace/discovery/MockSolrSearchCore.java
```diff
@@ -22,4 +22,8 @@ public void afterPropertiesSet() throws Exception {
         solr = null;
     }
 
+    public void reset() {
+        // This method intentionally left blank.
```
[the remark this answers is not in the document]
mwoodiupui: I added it because AbstractIntegrationTestWithDatabase calls it.

There used to be two MockSolrSearchCore classes, one in dspace-api and one in dspace-server-webapp.  I think that when I merged them, I kept the wrong one.  I'll try it the other way.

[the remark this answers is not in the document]
mwoodiupui: Thank you for asking.  This opened an interesting can of worms.  I hadn't copied enough from dspace-server-webapp, which incorrectly allowed this to succeed.

dspace-api/src/test/java/org/dspace/statistics/MockSolrLoggerServiceImpl.java
```diff
@@ -62,18 +64,28 @@ public void afterPropertiesSet() throws Exception {
...
         return new CityResponse(city, new Continent(), country, location, new MaxMind(), postal,
-                                country, new RepresentedCountry(), new ArrayList<>(0),
-                                new Traits());
+                                                 country,  new RepresentedCountry(), new ArrayList<>(0),
+                                                 new Traits());
+    }
+
+    /** Remove all records. */
+    public void reset() {
+        mockSolrServer.reset();
```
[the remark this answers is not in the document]
mwoodiupui: It's called from AbstractIntegrationTestWith Database.  I've tried to improve the JavaDoc.

dspace-api/pom.xml
```diff
@@ -537,6 +618,35 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>org.apache.lucene</groupId>
+            <artifactId>lucene-core</artifactId>
+            <version>${solr.client.version}</version>
```
tdonohue, line 624: Should this have a `<scope>test</scope>` as well? 

I'm not exactly sure why this new dependency was added, as it doesn't appear on `main` branch at all.  I'm assuming it's a dependency convergence error in the test environment?

dspace/modules/additions/pom.xml
```diff
@@ -194,6 +203,35 @@
             </exclusion>
          </exclusions>
       </dependency>
+        <dependency>
+            <groupId>org.apache.lucene</groupId>
+            <artifactId>lucene-core</artifactId>
+            <version>${solr.client.version}</version>
```
tdonohue, line 209: Same as above, should this have  `<scope>test</scope>` as well?

dspace-api/pom.xml
```diff
@@ -323,6 +325,11 @@
             <artifactId>apache-jena-libs</artifactId>
             <type>pom</type>
         </dependency>
+        <dependency>
+            <groupId>commons-cli</groupId>
+            <artifactId>commons-cli</artifactId>
+            <version>1.4</version>
```
tdonohue, line 331: Should this change be in the Parent POM (or maybe just removed)?  It looks like in the Parent POM (`[src]/pom.xml`) we already have commons-cli version 1.3.1 right now...So, it's unclear why we have a different version specified here.

dspace-api/pom.xml
```diff
@@ -537,6 +618,35 @@
...
+        <dependency>
+            <groupId>org.apache.lucene</groupId>
+            <artifactId>lucene-analyzers-stempel</artifactId>
+            <version>${solr.client.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.xmlbeans</groupId>
+            <artifactId>xmlbeans</artifactId>
+            <version>2.6.0</version>
```
tdonohue, line 648: This new dependency is also a bit odd...we don't have this one specified on `main` branch.  What this a dependency conversion fix?

dspace-api/pom.xml
```diff
@@ -717,7 +827,7 @@
 
         <dependency>
             <groupId>org.xmlunit</groupId>
-            <artifactId>xmlunit-matchers</artifactId>
+            <artifactId>xmlunit-core</artifactId>
```
tdonohue, line 830: Same here...why the switch to -core?  This seems new as well as it's not on `main` branch.

tdonohue (changes requested):
@mwoodiupui : Overall, the code changes look good to me.  However, I had some questions about a number of the newly added dependencies in POMs.  It seems like there's several unexplained new dependencies here...and at least one seems like it might be in the wrong POM (should possibly be moved up to the Parent POM).

It's possible there's very good reasons for all this, but I'd appreciate it if you took a second look and/or maybe add in some comments in the POMs (if the reasons are easily explainable that way).

dspace-api/pom.xml
mwoodiupui: All of the classes that we actually use are in -core, not -matchers.  `mvn dependency:analyze` pointed this out.  I don't know why we were using -matchers.

dspace-api/pom.xml
mwoodiupui: PoiWordFilter imports from xmlbeans.  `mvn dependency:analyze` showed xmlbeans as "used undeclared".  Even though it seemed to work in practice without the declaration, I tend to tidy these up whenever I become aware of them.  If I don't know why it worked, then I don't know why it might stop working.

dspace-api/pom.xml
mwoodiupui: Thank you.  It should be declared here (since this project uses it) and updated in the parent (since it is already managed there).  I'll do that.

dspace-api/pom.xml
mwoodiupui: org.dspace.app.util.IndexVersion (in src/main) imports five classes from this artifact.  I probably fixed it up while using `mvn dependency:analyze` to track down an unrelated dependency puzzle.  It should be default-scoped.

dspace/modules/additions/pom.xml
mwoodiupui: I recall that, once I fixed the above, I had to add it here too, to get `additions` to build.

dspace-api/pom.xml
tdonohue: Ah, thanks for clarification! This seems correct then

dspace-api/pom.xml
tdonohue: Yes, I see that now after looking closer at that PoiWordFilter. This is correct then...it just is slightly unrelated to this PR, but that's OK to clean this up :)

dspace-api/pom.xml
tdonohue: Ah, OK then. This seems fine then.  I was thrown off by the fact that lucene-core is in the midst of a bunch of other `test` dependencies.  At some point, we might want to put all the test dependencies together, separate from the non-test ones.  But, that can come in a followup PR.

tdonohue (approved):
👍 Looks good to me now!  Thanks @mwoodiupui !  I've also run a quick test that the commands in the README (for running individual ITs or Unit Tests) still work, and they do.  So, this looks like a great addition.

tdonohue:
@mwoodiupui : I just realized I caused a few minor IT conflicts for you after merging a separate PR.  Could you do a quick rebase here & then I'll merge this PR next?

PR #2964, opened 2020-09-11
DS-4551 Close FileInputStreams

MW3000:
## References
* Fixes [DS-4551](https://jira.lyrasis.org/browse/DS-4551)

## Description
Closes `FileInputStream`s immediately after they are no longer in use, to avoid an excessive number of open files when processing a large upload. In the original code, only the last `FileInputStream` was closed, because `is.close()` was called outside of the loop that opens the streams instead of within the loop. I just moved the `is.close()` to the appropriate place.

To verify the problem and the result _from outside_ you can profile open files for the process. Description of how to do so can be found in [DS-4551](https://jira.lyrasis.org/browse/DS-4551).


pnbecker (approved):
Obvious fix. +1

pnbecker:
merged as being tiny and obvious.

PR #2977, opened 2020-09-21
[DS-2569] Fail gracefully if CC-API is down

J4bbi:
## Description
Currently getLicenses returns null instead of a collection of 0 or more CCLicenses if there is any exception when making an API call to the CC licenses API.

The API was down last week and I sent a query to CC who responded that it's a legacy API that shouldn't have been running in the first place!

## Instructions for Reviewers
This PR just creates a check for a null value to the getLicenses result and if it is null then it prints out a warning to the submitter that licenses could not be obtained. (This error message could be put into messages.xml)

To reproduce an error it's simple to change the value of _cc.api.rooturl_ in dspace/local.cfg to an incorrect value.

It is likely that the API has not got much left so maybe we need to start thinking about this more longterm.


kshepherd:
@J4bbi looks great, i think we should put the error text into the i18n catalog too

kshepherd (approved):
+1 good fix! thanks @J4bbi

PR #2985, opened 2020-09-28
DS-4552 Update crosswalk according to new guidelines

J4bbi:
## Description
The crosswalk for DOI minting currently mints using HTTP and the subdomain "dx.doi.org" whereas the official guidelines recommends HTTPS and "doi.org" https://support.datacite.org/docs/datacite-doi-display-guidelines.

This PR simply changes the value to use HTTPS and drop the subdomain. Additionally it stores this as a variable to avoid repetition but at a slight cost of readability.


kshepherd (approved):
+1 easy win and complies with doi guidelines

kshepherd:
added port to main as the crosswalk exists there too (as well as the EZID crosswalk)
there may be other places to change, too, though we don't want to affect resolution or detecting of existing dc.identifier.uri values with dx.doi.org prefixes... I'll make a note on the JIRA issue https://jira.lyrasis.org/browse/DS-4552

PR #2986, opened 2020-09-29
[DS-4552] Revert PR#2985 "DS-4552 Update crosswalk according to new guidelines"

kshepherd:
JIRA: https://jira.lyrasis.org/browse/DS-4552

Reverts DSpace/DSpace#2985 - as per notes on issue, it may well be too risky to try and update these prefixes in a minor 6.x release
Many sites will have existing items with dc.identifier.uri values using the deprecated prefix, other crosswalk mapping and transformation rules for the deprecated prefix. Given this risk and the fact that the dx.doi.org is deprecated but not removed (and it is highly unlikely that a URI prefix for a persistent URI service will stop working even in the long-term), I've spoken to other devs and Hrafn and think we should revert the previous quick change in the Datacite crosswalk


J4bbi (approved):

kshepherd:
@J4bbi (original PR author) has approved and agreed - we will play it safe for now and plan prefix refactor for a future release

PR #2999, opened 2020-10-12
Cleaning up the cache of Builders after each test

Raf-atmire:
## Description
Currently the map of builders is not cleaned after every test run. This caused the issue of the delete being called on a builder for which there was no object anymore since the previous test had deleted it.
Since this delete was being called, we're creating a Context with the try-with-resources. Since there's no object anymore, the context.complete() will not be called. But because we're using a try-with-resources, the context will be closed and therefor aborted. Since contexts are shared on the same thread, this causes the general context to be aborted and not committed. If this happened in the first Builder after a test, any manual cleanup through services in that test would be lost as the context would never be committed. This caused for data inconsistencies.

## Instructions for Reviewers

List of changes in this PR:
* Added a cleanup to the map of Builders after each test run so that 'dead' builders aren't kept and ran when they don't contain a valid Object anyway.

### How to test this PR

* Boot up the code without this PR and create a test in which you create a RegistrationData object. Then delete the RegistrationData object afterwards in that very same test
* Create a test in which you create an EPerson through a Builder.
* Create a third test in which you check that the list of RegistrationData fetched through a findall at the beginning of the test is empty.

Execution of the tests in order for this problem to occur has to be:
1. The test that creates the EPerson through the builder
2. The test that manually creates and deletes the RegistrationData object (Creation through a REST call, deletion through a service.delete)
3. The test that checks that the list of RegistrationData from a findAll is empty

## Checklist

### DIFF
--- a/dspace-api/src/test/java/org/dspace/AbstractIntegrationTestWithDatabase.java
+++ b/dspace-api/src/test/java/org/dspace/AbstractIntegrationTestWithDatabase.java
@@ -192,6 +192,8 @@ public class AbstractIntegrationTestWithDatabase extends AbstractDSpaceIntegrati
             // Reload our ConfigurationService (to reset configs to defaults again)
             DSpaceServicesFactory.getInstance().getConfigurationService().reloadConfig();
 
+            AbstractBuilder.cleanupBuilderCache();
+
             // NOTE: we explicitly do NOT destroy our default eperson & admin as they
             // are cached and reused for all tests. This speeds up all tests.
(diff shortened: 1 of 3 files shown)


tdonohue (approved):
👍 Code looks correct to me.  Tests pass (and seemingly ITs are *faster now*?).  The explanation here makes sense to me & I agree the Builder map needs to be reinitialized from scratch between tests just like everything else.  Thanks @Raf-atmire !

PR #3079, opened 2020-12-07
3078 Change EntityTypeLabelRestController get by label to return REST…

### LINKED ISSUE #3078: EntityTypeLabelRestController get by `Label` should return as REST Resource
(Related PR https://github.com/DSpace/dspace-angular/issues/964 )

**Describe the bug**
Currently requesting `/label/{entity-type-label}` from `EntityTypeLabelRestController` returns of `EntityTypeRest` type which will not include the `self` or `relationshiptypes` links that are commonly made use of on the Angular side. 
Submitting PR shortly.

AndrewZachWood:
Issue: https://github.com/DSpace/DSpace/issues/3078


tdonohue (approved):
👍 Code looks good & makes sense.  I tested this alongside https://github.com/DSpace/dspace-angular/pull/965 (corresponding Angular PR) and they work great!  Thanks @AndrewZWood !

PR #3087, opened 2020-12-10
DS-4563 Signal that the transaction should be committed

cwilper:
Fixes https://jira.lyrasis.org/browse/DS-4563 by setting the curation context to a value that will cause the curator to commit the transaction.

This fix has been tested successfully on a DSpace 5.4 repository.


kshepherd (approved):
+1 quick win

PR #3109, opened 2021-01-07
Limit externalSources per metadata field

Raf-atmire:
## References
* Fixes REST side for https://github.com/DSpace/dspace-angular/issues/943
* Related to https://github.com/DSpace/Rest7Contract/pull/153

## Description
This PR will add support for an "externalsources" xml tag for a "relation-field" xml tag in the submission forms. It'll be parsed by the Java code and it'll be set in the REST response so that Angular can use this field to limit their external sources lookup.
This "externalsources" field can be multivalued and can be empty as well, it's currently set to optional in the .dtd file.
We've also added a relation-field tag on the PeopleStep of the submissions with pubmed as externalsource. This will also provide Angular with an easier way of testing/using their fix for this issue

## Instructions for Reviewers

List of changes in this PR:
* Added support for "externalsources" tag for a "relation-field" tag in the submission forms
* Added an isPublicationOfAuthor relation-field for the PeopleStep

How to test this PR:
* Build and deploy the code locally
* Ask the submission forms for the PeopleStep config
* Verify that it shows pubmed in an "externalsources" field

## Checklist

### DIFF
--- a/dspace-api/src/main/java/org/dspace/app/util/DCInput.java
+++ b/dspace-api/src/main/java/org/dspace/app/util/DCInput.java
@@ -145,6 +145,7 @@ public class DCInput {
     private String relationshipType = null;
     private String searchConfiguration = null;
     private String filter;
+    private List<String> externalSources;
 
     /**
      * The scope of the input sets, this restricts hidden metadata fields from
@@ -226,6 +227,15 @@ public class DCInput {
         relationshipType = fieldMap.get("relationship-type");
         searchConfiguration = fieldMap.get("search-configuration");
         filter = fieldMap.get("filter");
+        externalSources = new ArrayList<>();
+        String externalSourcesDef = fieldMap.get("externalsources");
+        if (StringUtils.isNotBlank(externalSourcesDef)) {
+            String[] sources = StringUtils.split(externalSourcesDef, ",");
+            for (String source: sources) {
+                externalSources.add(StringUtils.trim(source));
+            }
+        }
+
     }
 
     /**
@@ -520,6 +530,10 @@ public class DCInput {
 
     public String getFilter() {
         return filter;
+    }
+
+    public List<String> getExternalSources() {
+        return externalSources;
     }
 
     public boolean isQualdropValue() {
--- a/dspace/config/item-submission.xml
+++ b/dspace/config/item-submission.xml
@@ -18,7 +18,7 @@
     <!-- for handle "default". -->
     <submission-map>
         <name-map collection-handle="default" submission-name="traditional"/>
-        <name-map collection-handle="10673/1623" submission-name="People" />
+        <name-map collection-handle="123456789/1161" submission-name="People" />
     </submission-map>
 
 
--- a/dspace/config/item-submission.xml
+++ b/dspace/config/item-submission.xml
@@ -18,6 +18,7 @@
     <!-- for handle "default". -->
     <submission-map>
         <name-map collection-handle="default" submission-name="traditional"/>
+        <name-map collection-handle="10673/1623" submission-name="People" />
     </submission-map>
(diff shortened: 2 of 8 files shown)


dspace/config/item-submission.xml
```diff
@@ -18,6 +18,7 @@
     <!-- for handle "default". -->
     <submission-map>
         <name-map collection-handle="default" submission-name="traditional"/>
+        <name-map collection-handle="123456789/1161" submission-name="People" />
```
paulo-graca, line 21: This shouln't be versioned.

dspace-api/src/main/java/org/dspace/app/util/DCInput.java
```diff
@@ -226,6 +227,15 @@ public DCInput(Map<String, String> fieldMap, Map<String, List<String>> listMap)
         relationshipType = fieldMap.get("relationship-type");
         searchConfiguration = fieldMap.get("search-configuration");
         filter = fieldMap.get("filter");
+        externalSources = new ArrayList<>();
+        String externalSourcesDef = fieldMap.get("externalsources");
+        if (StringUtils.isNotBlank(externalSourcesDef)) {
+            String[] sources = StringUtils.split(externalSourcesDef, ",");
```
paulo-graca, line 233: can we have any example, or use case where we use more than one external source? I'm asking this because I didn't see in any place the case of defining multiple external sources (it's comma separated), like:

`<externalsources>crossref, datacite</externalsources>`

I think it could be useful.

paulo-graca (commented):
This PR works as expected, I'm getting (`"externalSources": ["orcid"]`):

```
{
  "id": "traditionalpageone",
  "name": "traditionalpageone",
  "rows": [{
      "fields": [{
          "input": {
            "type": "name"
          },
          "label": "Author",
          "mandatory": true,
          "repeatable": true,
          "mandatoryMessage": "At least one author (plain text or relationship) is required",
          "hints": "Add an author",
          "selectableRelationship": {
            "relationshipType": "isAuthorOfPublication",
            "filter": null,
            "searchConfiguration": "person",
            "nameVariants": "false",
            "externalSources": ["orcid"]
          },
          "selectableMetadata": [{
              "metadata": "dc.contributor.author",
              "label": null,
              "closed": false
            }],
          "languageCodes": []
        }]
    }, {
```

abollini (changes requested):
I agree with the review of @paulo-graca so once his feedback is processed it is +1 from me as well

dspace-api/src/main/java/org/dspace/app/util/DCInput.java
Raf-atmire: There's an example for this in the test's config: https://github.com/DSpace/DSpace/pull/3109/files#diff-47993096c9f4173f69f5e9722712ea54753666c578f5b86b691021d973ce1e19R66
It's verified to work properly in this test: https://github.com/DSpace/DSpace/pull/3109/files#diff-6009af8825c771129ab9a3b5b7e3c104715cb2690ba72ff3cd814f2d496c6f6dR611

dspace/config/item-submission.xml
Raf-atmire: Indeed, that was a remnant from a test case that we used. I'll remove this config

Raf-atmire:
@paulo-graca @abollini Thank you for the review! I've addressed both of the feedback points. Please let me know if anything else seems amiss.

paulo-graca (approved):
Thank you @Raf-atmire !

tdonohue (approved):
👍 Looks good to me.  This is already at essentially +3 (since @abollini said he's also +1 once @paulo-graca feedback is addressed & it has been)

PR #3111, opened 2021-01-08
Fix puzzling incorrect failure to match dspace.server.url

### LINKED ISSUE #3110: Poor URL matching causes improper failure with confusing message
**400 Bad Request** while rendering the initial page server-side.  The underlying error made no sense:  `java.lang.IllegalArgumentException: the supplied uri is not valid: http://ds7.wood.net/server/api/core/sites/2510e895-db7e-46b2-9a9e-b4fd92026156`.  The string *is* a valid HTTP URI, and fetching it with `curl` succeeds.  I traced the problem to the way that URIs are validated in `org.dspace.app.rest.utils.Utils.getBaseObjectRestFromUri`.  The problem was revealed because I had configured `dspace.server.url` with an explicit port 80, which is useless (for protocol `http`) but not incorrect.  This failure is incorrect, surprising, and time-consuming to understand.

The ultimate cause of the failure is that `getBaseObjectRestFromUri` compares URLs as strings rather than as serializations of complex structured objects (which is what URLs are).  I suggest that, to discover whether URL A is a prefix of URL B, one must compare protocol, host, port, and path separately.

**To Reproduce**
1. Configure `dspace.server.url` explicitly with the default port for its protocol.
2. Start up front end and back end, and browse to the front end.

**Expected behavior**
For example:  the URL `http://example.com:80/foo` should be recognized as a prefix of `http://example.com/foo/bar`.

mwoodiupui:
## References
* Fixes #3110

## Description
When validating a URL against `dspace.server.url`, compare the URLs field by field, not as simple strings.

## Instructions for Reviewers
List of changes in this PR:
* New utility method to compare URLs for prefix-ness.
* New unit test class to test new method.

## Checklist


mwoodiupui:
The first patch worked but was insufficient, because the old code was also making assumptions about string lengths when clipping off the leading "api/".  This has now been tested on the instance that first showed me the problem, and the problem is no longer evident.

dspace-server-webapp/src/test/java/org/dspace/app/rest/utils/URLUtilsTest.java
```diff
@@ -0,0 +1,67 @@
...
+        boolean isPrefix;
+
+        isPrefix = URLUtils.urlIsPrefixOf("http://example.com/path", "http://example.com/path");
+        assertTrue("Should match if all is equal", isPrefix);
+        isPrefix = URLUtils.urlIsPrefixOf("http://example.com:80/test", "http://example.com:80/test/1");
+        assertTrue("Should match if pattern path is longer", isPrefix);
+        isPrefix = URLUtils.urlIsPrefixOf("http://example.com:80/test", "http://example.com/test");
+        assertTrue("Should match if missing port matches default", isPrefix);
+
+        isPrefix = URLUtils.urlIsPrefixOf("http://example.com/path", "https://example.com/path");
```
abollini, line 55: can we add test for 
- http://example.com/path vs http://example.com/path/
- http://example.com/path vs http://example.com/path2
- http://example.com/path vs http://example.com/path2/sub

dspace-server-webapp/src/main/java/org/dspace/app/rest/utils/Utils.java
```diff
@@ -917,33 +931,48 @@ public Object getDSpaceAPIObjectFromRest(Context context, BaseObjectRest restObj
...
+        }
+
+        // Extract from the URI the category, model and id components.
+        // They start after the dspaceUrl/api/{apiCategory}/{apiModel}/{id}
+        URL dspaceUrlObject;
+        URL requestUrlObject;
+        try {
+            dspaceUrlObject = new URL(dspaceUrl);
+            requestUrlObject = new URL(uri);
+        } catch (MalformedURLException ex) {
```
abollini, line 945: we should think if this need to go before the prefix check. Eventually, the reason for that is that if the supplied uri is not a syntactically valid URI a BadRequest exception could be more appropriate than a potential 422 error reported if the URI is not valid or lead to a not appropriate object

abollini (approved):
thanks @mwoodiupui the code looks good to me, I have added two comments in line for discussion and improve the test suite

dspace-server-webapp/src/test/java/org/dspace/app/rest/utils/URLUtilsTest.java
mwoodiupui: Thank you.  Those are good suggestions.  Also, working on this revealed that I had the parameters reversed.

I took the opportunity to remove a "FIXME" by replacing the string comparison of paths with an element-by-element comparison which should be more accurate.

dspace-server-webapp/src/main/java/org/dspace/app/rest/utils/Utils.java
mwoodiupui: A good suggestion.  Done.

tdonohue (approved):
👍 Looks good to me as well. Thanks for this @mwoodiupui !  The tests (and recent improvements) all look good too

PR #3112, opened 2021-01-11
[DS-4452] Porting correct instantiation of default ignore fields (CSV…

kshepherd:
Porting correct instantiation of default ignore fields in DSpaceCSV from #2661 

## References
* Fixes [DS-4452](https://jira.lyrasis.org/browse/DS-4452)

## Description
This is a simple port of the very simple patch in #2661 that correctly instantiates the `String[] defaultValues` array with the default field names as an array of string, not a single concatenated string. It's already been merged in dspace-6_x


tdonohue (approved):
👍 Looks correct based on how array configurations now work & this was tested in #2661.  Merging immediately. Thanks for the port, @kshepherd !

PR #3135, opened 2021-01-27
Rename test to avoid to mess up the test env for the new clean ITs

abollini:
## References
* Workaround for #3055

## Description
We have hit random failure on ITs due to an inconsistent state of the test environment. This is more often the case when solr is involved but I guess that could happen also in other scenarios. The problem is that the cleanup of the test environment is guaranteed only extending `org.dspace.AbstractIntegrationTestWithDatabase` not extending `org.dspace.AbstractIntegrationTest` that is indeed in a completely separate tree.
The test that extends this later class are sometimes named according to the IT convention other time as UnitTest this influence which mvn plugin pick them during CI. Moreover, many dspace named Unit Test extending the previous AbstractIntegrationTest or also only the AbstractUnitTest are technically Integration Test so they have pick the wrong name...
unfortunately, making a complete cleanup of these tests is not possible right now due to the dspace 7 timeline so the best option in my opinion is to "give the wrong name" to the residual buggy IT so that they will not disturb the new IT written cleaner thanks to the new  `org.dspace.AbstractIntegrationTestWithDatabase`  and the `Builder` classes.

Unfortunately, I have had to disable the `org.dspace.content.crosswalk.QDCCrosswalkTest` that once renamed has started to fail due to conflict with other unit test. As this test is related to a minor feature I believe that it is worth to disable it instead than spent more hours (already tried to fix for a couple of hours).

## Instructions for Reviewers
A recent failing PR gives us a very clean way to identify the problem
https://github.com/DSpace/DSpace/pull/3133/checks?check_run_id=1778185064#step:5:591
the failing class that succeed in other run (https://github.com/4Science/DSpace/runs/1777980274?check_suite_focus=true ) and locally if run alone fail in a very initial not controversial check saying that nothing is in SOLR.
The reason for that is that the ITCurator test extending `org.dspace.AbstractIntegrationTest` is triggering this code that will disable completely solr for all the next test
```
    // Ensure all tests run with Solr indexing disabled
    disableSolrIndexing();
```

## Checklist

### DIFF
--- a/dspace-api/src/test/java/org/dspace/administer/StructBuilderTest.java
+++ b/dspace-api/src/test/java/org/dspace/administer/StructBuilderTest.java
@@ -49,18 +49,23 @@ import org.xmlunit.diff.Difference;
 /**
  * Tests of {@link StructBuilder}.
  *
+ * FIXME Due to a not optimal definition of UnitTest and IntegrationTest in
+ * DSpace see https://github.com/DSpace/DSpace/issues/3055 all the integration
+ * test that inherit from org.dspace.AbstractUnitTest without denying its init()
+ * method behavior MUST BE named as Unit Test
+ *
  * @author Mark H. Wood <mwood@iupui.edu>
  */
-public class StructBuilderIT
+public class StructBuilderTest
         extends AbstractIntegrationTest {
-    private static final Logger log = LoggerFactory.getLogger(StructBuilderIT.class);
+    private static final Logger log = LoggerFactory.getLogger(StructBuilderTest.class);
 
     private static final CommunityService communityService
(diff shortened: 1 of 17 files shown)


abollini:
I'm closing this PR as the fix in https://github.com/DSpace/DSpace/pull/3133/commits/f349e7f4ce605967d7b41a8366f0a6bda47c48c3 included in the PR https://github.com/DSpace/DSpace/pull/3133 seems to improve the situation a lot with less risk than the huge movement of test from our area to the other that is proposed here.
That said, I'm going to create a new PR in any case to move at least some "good" IT that are not named in the right way and two unit test that claiming to be IT can still generate random failure

PR #3165, opened 2021-02-25
[DS-4575] v.6 - Exposed more parameters for DBCP2

J4bbi:
Exposed more parameters for DBCP2 in the DSpace config.

Notably, maxConnLifeTime and removeAbandoned

See:

- #3070 
- #3080 
- #3162 

See also:
http://dspace.2283337.n4.nabble.com/Notes-on-PostgreSQL-connection-pooling-with-a-Tomcat-JNDI-resource-td4687149.html

### DIFF
--- a/dspace/config/dspace.cfg
+++ b/dspace/config/dspace.cfg
@@ -86,18 +86,41 @@ db.password = dspace
 #      so this may be set to ${db.username} in most scenarios.
 db.schema = public
 
-## Connection pool parameters
+## Database Connection pool parameters
+# DSpace wraps around the Apache Commons DBCP2 component, to read about its configuration
+# see: https://commons.apache.org/proper/commons-dbcp/configuration.html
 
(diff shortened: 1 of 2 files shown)


J4bbi:
Just a note for discussion.

This PR maps to config for DBCP2.

It enables the admin to switch one the built-in function to remove abandoned connections. By default the timeout for abandoned connections is 5 minutes (300 secs). 

After trying that out we've found that limit to be to short. Which is to say the library is closing valid idle connections on our instance it seems:

`2021-09-13 11:51:48,393 ERROR org.hibernate.engine.jdbc.spi.SqlExceptionHelper @ Connection org.postgresql.jdbc.PgConnection@5c1ab3b4 is closed.
2021-09-13 11:51:48,393 ERROR org.dspace.statistics.SolrLoggerUsageEventListener @ could not prepare statement`

I am sure this is an issue for tweaking.

PR #3196, opened 2021-03-18
Item not indexed due to authority value not being in a controlled vocabulary

### LINKED ISSUE #3177: Item not indexed due to authority value not being in a controlled vocabulary
**Describe the bug**
DSpace 7 backend has had an alteration that would allow us to specify a certain controlled vocabulary file for an input-field in the submission, the authority values added to the metadata values for this field would have to be present as IDs within this controlled vocabulary file.
However, if they're not present or if those authority values would ever be removed from that file, the Item which contains this metadatavalue would never be indexed in Solr again as Index Discovery would crash for this Item.

The submission-forms defines a dc.subject metadata field with the following vocabulary as its controlled vocabulary, as can be seen <a href="https://github.com/DSpace/DSpace/blob/main/dspace/config/submission-forms.xml#L205">here</a>
The file can be found <a href="https://github.com/DSpace/DSpace/blob/main/dspace/config/controlled-vocabularies/srsc.xml">here</a> and any authority value that's ever added to a metadatavalue with dc.subject as metadatafield, would have to be present in the id attributes of this file. So for example "VR110102" would be a valid authority value, but "authorityvalue_1" would be wrong. However, when specifying such a wrong authorityValue, the index discovery crashes.
The indexContent fails here: https://github.com/DSpace/DSpace/blob/f5aac1eefaaf969ca53d1bee5b2fdc68f5fd50fb/dspace-api/src/main/java/org/dspace/discovery/IndexEventConsumer.java#L210 through the following flow:
- <a href="https://github.com/DSpace/DSpace/blob/d5d3b55457385cf4820a420d8065ea2376545fdd/dspace-api/src/main/java/org/dspace/discovery/SolrServiceMetadataBrowseIndexingPlugin.java#L177">label call</a>
- retrieve <a href="https://github.com/DSpace/DSpace/blob/ed58d44329c67b1810851d9795364c7acbc00d0b/dspace-api/src/main/java/org/dspace/content/authority/ChoiceAuthorityServiceImpl.java#L194">choiceAuthority</a>
- get the <a href="https://github.com/DSpace/DSpace/blob/8fd328855a69c711ea48cecaff7c911788dc15ff/dspace-api/src/main/java/org/dspace/content/authority/DSpaceControlledVocabulary.java#L317">node</a> for this key

However since there can't be a node found for the authority value that doesn't exist within the file, it'll crash on a nullpointer exception here: https://github.com/DSpace/DSpace/blob/8fd328855a69c711ea48cecaff7c911788dc15ff/dspace-api/src/main/java/org/dspace/content/authority/DSpaceControlledVocabulary.java#L321
This will cause the indexing to break for this Item and it'll never be indexed.

This brings along another layer of issues. If we create Items within our repository with authority values that are present within the controlled vocabulary file today, but we alter the vocabulary file tomorrow and remove certain authority values, any items with those removed authority values would cease to be reindexed in discovery and would always remain broken and not show up in discovery.

**To Reproduce**
1. Enable authority control on dc.subject in the config
2. Create an Item with a dc.subject metadatavalue which has an authority that isn't represented in the respective controlled vocabulary file
3. Run index-discovery -i {handle} for this Item
4. Verify that it crashes and that the item isn't present in the search core

**Expected behavior**
The Item should be present in the search core

**Related work**
This was discovered during development of: https://github.com/DSpace/DSpace/pull/3176

Micheleboychuk:
## Description
Fixes #3177

## Checklist


dspace-api/src/main/java/org/dspace/content/authority/MetadataAuthorityServiceImpl.java
```diff
@@ -83,6 +83,11 @@ protected MetadataAuthorityServiceImpl() {
 
     }
 
+    @Override
+    public void reset() {
```
Raf-atmire, line 87: I feel like we should also clear the other map in this method. A reset method within a service should reset each variable that's caching information, this way there can't be any confusion later down the line as to why the other one isn't getting reset when this method is called. I'm referring to the `controlled` variable here

dspace-api/src/test/java/org/dspace/discovery/DiscoveryIT.java
```diff
@@ -273,6 +281,35 @@ public void solrRecordAfterDeleteTest() throws Exception {
...
+        metadataAuthorityService.reset();
+        try {
+            context.turnOffAuthorisationSystem();
+
+            parentCommunity = CommunityBuilder.createCommunity(context)
+                                              .withName("Parent Community").build();
+
+            Collection col1 = CollectionBuilder.createCollection(context, parentCommunity)
+                                               .withName("Collection 1").build();
+
```
Raf-atmire, line 296: Could we provide a starting state assertion here? I'd like to see `assertSearchQuery(IndexableItem.TYPE, 0);` being added here so that we can prove that we didn't have any Items present in SOLR before creating the one with the "non-existing" authority value.
This way, we're always sure that the item being indexed is the item within the test, since if it wasn't then the starting state assertion would have failed. This protects us from the use case where the actual indexing of this authority value would break, but the test could succeed because an Item in a previous test wasn't getting cleared correctly.

Raf-atmire (changes requested):
Hi @Micheleboychuk Thank you for the work! It looks great overall, I just have a few minor change requests, please check the inline comments. Could you have a look at this when you get the time? Thanks!

abollini:
@Raf-atmire can you recheck this please? your feedback should be addressed now as we have merged the main in this PR after that your PR https://github.com/DSpace/DSpace/pull/3176 was merged

Raf-atmire (approved):
Hi @Micheleboychuk @abollini Thank you for adding the check to the tests! This, combined with the merge, solves all my requests. +1 from me, thank you for the work!

PR #3212, opened 2021-03-31
Fix #3211 by depending directly on tika-parsers

### LINKED ISSUE #3211: Full-text indexing fails due to missing Tika library
**Describe the bug**
When adding or updating an item in the search core (via the discovery consumer, e.g. after a metadata edit occurs in the UI), full-text indexing fails, with an error in dspace.log containing the following snippet:

```
Caused by: java.lang.NoClassDefFoundError: org/apache/tika/parser/csv/TextAndCSVParser
        at org.dspace.discovery.indexobject.IndexFactoryImpl.writeDocument(IndexFactoryImpl.java:99) ~[dspace-api-7.0-beta5-SNAPSHOT.jar:7.0-beta5-SNAPSHOT]
```

**To Reproduce**
1. Create an item with a PDF
2. Run filter-media on it so that a fulltext bitstream is created in the TEXT bundle
3. Edit the item to modify the metadata
4. Notice that the indexing fails and a message like the above appears in dspace.log

**Expected behavior**
Full-text indexing should succeed without error.

cwilper:
Fixes #3211

## Description

This small PR changes the `tika-core` dependency of `dspace-api` to `tika-parsers`. When applied, full text indexing works from the discovery consumer again without error.

Note: The reason this probably went unnoticed is that `solr-cell`'s scope is set to `test` in `dspace-server-webapp`'s `pom.xml`. It appears this may have been done because because `solr-cell` brings in a lot of other dependencies that were unwanted in that context. However, for full text indexing to work in the context of the webapp, we do currently require `tika-parsers` to be deployed with it.

### DIFF
--- a/dspace-api/pom.xml
+++ b/dspace-api/pom.xml
@@ -632,7 +632,7 @@
         <!-- Used for full-text indexing with Solr -->
         <dependency>
             <groupId>org.apache.tika</groupId>
-            <artifactId>tika-core</artifactId>
+            <artifactId>tika-parsers</artifactId>
         </dependency>
         <dependency>
             <groupId>org.apache.lucene</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -1196,6 +1196,29 @@
                 <groupId>org.apache.tika</groupId>
                 <artifactId>tika-parsers</artifactId>
                 <version>1.24.1</version>
+                <!-- Exclude a few tika-parsers dependencies that we already use, with slightly different versions -->
+                <exclusions>
+                    <exclusion>
+                        <groupId>com.google.protobuf</groupId>
+                        <artifactId>protobuf-java</artifactId>
+                    </exclusion>
+                    <exclusion>
+                        <groupId>net.sf.ehcache</groupId>
+                        <artifactId>ehcache-core</artifactId>
+                    </exclusion>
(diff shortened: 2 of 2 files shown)


dspace-api/pom.xml
```diff
@@ -632,7 +632,7 @@
         <!-- Used for full-text indexing with Solr -->
         <dependency>
             <groupId>org.apache.tika</groupId>
-            <artifactId>tika-core</artifactId>
+            <artifactId>tika-parsers</artifactId>
```
mwoodiupui, line 635: We depend on both tika-parsers *and* tika-core.  For example, `org/dspace/discovery/indexobject/IndexFactoryImpl.java` imports `org.apache.tika.exception.TikaException`.

cwilper: True. I considered having both in the pom, but parsers already brings in core transitively, and the tika docs specifically said:

> If you want to use Tika to parse documents (instead of simply detecting document types, etc.), you'll want to depend on tika-parsers instead

mwoodiupui: It will work as-is.

mwoodiupui (approved):
This all looks reasonable.  (I had run into similar issues, with a similar fix, in an experimental patch.)

PR #3228, opened 2021-04-08
[DS-4579] Re-enable HTTP Ranges support

J4bbi:
Support for  HTTP Range requests was disabled way back in 2010-11 because "ome windows pdf viewers are incapable of handling this request.".

Supporting Range requests is however a rather standard expected functionality of web servers, especially for repositories serving large files. Moreover, Range requests are required in order to enable skipping or seeking when jumping along timelines of audio or video files.

If enabling Ranges requests by default can possibly break some behaviour or cause problems it would be possible to introduce a config boolean to be included in the if clause on line 699.

@anis-moubarik contributed to this PR

Don't think this supports spec 100% but is rather basic support.
See spec: https://tools.ietf.org/html/rfc7233

### DIFF
--- a/dspace-xmlui/src/main/java/org/dspace/app/xmlui/cocoon/BitstreamReader.java
+++ b/dspace-xmlui/src/main/java/org/dspace/app/xmlui/cocoon/BitstreamReader.java
@@ -694,38 +694,39 @@ public class BitstreamReader extends AbstractReader implements Recyclable
         // viewers are incapable of handling this request. You can
         // uncomment the following lines to turn this feature back on.
 
-//        response.setHeader("Accept-Ranges", "bytes");
-//        String ranges = request.getHeader("Range");
-//        if (ranges != null)
-//        {
-//            try
-//            {
-//                ranges = ranges.substring(ranges.indexOf('=') + 1);
-//                byteRange = new ByteRange(ranges);
-//            }
-//            catch (NumberFormatException e)
-//            {
-//                byteRange = null;
-//                if (response instanceof HttpResponse)
-//                {
-//                    // Respond with status 416 (Request range not
-//                    // satisfiable)
-//                    response.setStatus(416);
-//                }
-//            }
-//        }
+        response.setHeader("Accept-Ranges", "bytes");
+        String ranges = request.getHeader("Range");
+        if (ranges != null)
+        {
+            try
+            {
+                ranges = ranges.substring(ranges.indexOf('=') + 1);
+                byteRange = new ByteRange(ranges);
+            }
+            catch (NumberFormatException e)
+            {
+                byteRange = null;
+                if (response instanceof HttpResponse)
+                {
+                    //Respond with status 416 (Request range not
+                    //satisfiable)
+                    response.setStatus(416);
+                }
+            }
+        }
 
         try
         {
             if (byteRange != null)
             {
+                ByteRange actualByteRange = byteRange;
                 String entityLength;
                 String entityRange;
                 if (this.bitstreamSize != -1)
                 {
                     entityLength = "" + this.bitstreamSize;
-                    entityRange = byteRange.intersection(
-                            new ByteRange(0, this.bitstreamSize)).toString();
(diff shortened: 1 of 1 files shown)


J4bbi:
To test you can try curling a bitstream:

curl -I http://exampleserver.com/bitstream/handle/123456789/37556/example_video.mp4?sequence=1&isAllowed=y

and verify that Accept: Ranges is returned
also:

curl --header "Range: bytes=100-107" -I http://exampleserver.com/bitstream/handle/123456789/37556/example_video.mp4?sequence=1&isAllowed=y

see:
https://stackoverflow.com/questions/720419/how-can-i-find-out-whether-a-server-supports-the-range-header

kshepherd (commented):
Looks like a great quick win, i've noticed and done the same in other 6.x code bases outside of this branch - the comment about viewers not handling ranges is very out of date, and we're best off to implement this.

Just one question before my +1 - is there a reason we're not setting response status to 206 partial?

NOTE - we should definitely include documentation and a heads up about the change, as it can affect statistics / metrics when looking at apache logs, since all the 206 partial requests (if that is the response code) can sometimes trick systems looking for 2xx as 'successful hit'

J4bbi:
> Looks like a great quick win, i've noticed and done the same in other 6.x code bases outside of this branch - the comment about viewers not handling ranges is very out of date, and we're best off to implement this.
> 
> Just one question before my +1 - is there a reason we're not setting response status to 206 partial?
> 
> NOTE - we should definitely include documentation and a heads up about the change, as it can affect statistics / metrics when looking at apache logs, since all the 206 partial requests (if that is the response code) can sometimes trick systems looking for 2xx as 'successful hit'

Ahhhh oops I did it again. Of course, setting status to 206 is essential.

I c+p the code from a working branch onto a new DS branch and I omitted that part. Just me being too hasty.

kshepherd:
@J4bbi thanks, looks great - the failed LGTM java build test looked to be a temp failure copying some maven artifacts so I've re-run it and don't anticipate problems, but I'll hold off on my +1 till I can confirm all those tests

alanorth:
Haha this is pretty cool actually. It totally works. I tested the for the `Accept-Ranges` response header and then tried to download a bitstream in two separate requests with sequential ranges:

```console
$ curl -I 'http://localhost:8080/bitstream/handle/10568/114147/apache-solr-ref-guide-4.10.pdf?sequence=1&isAllowed=y'
HTTP/1.1 200 OK
Server: Apache-Coyote/1.1
Set-Cookie: JSESSIONID=FA7C06BA5252BFC753F392F20B3BA584; Path=/; HttpOnly
X-Cocoon-Version: 2.2.0
Vary: User-Agent
Last-Modified: Fri, 23 Jul 2021 12:00:27 GMT
Expires: Fri, 23 Jul 2021 13:14:56 GMT
Content-Disposition: attachment;filename="apache-solr-ref-guide-4.10.pdf"
Accept-Ranges: bytes
Content-Type: application/pdf;charset=ISO-8859-1
Content-Language: en-US
Content-Length: 9805564
Date: Fri, 23 Jul 2021 12:14:56 GMT
$ curl -s --header 'Range: bytes=0-90000' 'http://localhost:8080/bitstream/handle/10568/114147/apache-solr-ref-guide-4.10.pdf?sequence=1&isAllowed=y' > /tmp/omg.pdf
$ curl -s --header 'Range: bytes=90001-9805564' 'http://localhost:8080/bitstream/handle/10568/114147/apache-solr-ref-guide-4.10.pdf?sequence=1&isAllowed=y' >> /tmp/omg.pdf
```

Much to my amusement the PDF I downloaded in two partial range requests is identical to the original:

```console
$ sha256sum ~/Downloads/apache-solr-ref-guide-4.10.pdf /tmp/omg.pdf
b04b2daf6b5e845b4552815d13cab31405aa7597d74b57a279bddd6fbac22cab  /home/aorth/Downloads/apache-solr-ref-guide-4.10.pdf
b04b2daf6b5e845b4552815d13cab31405aa7597d74b57a279bddd6fbac22cab  /tmp/omg.pdf
```
Resounding thumbs up from me. :+1:

kshepherd (approved):

PR #3277, opened 2021-05-19
Delete item as collection admin & related issues

### LINKED ISSUE #3262: Error when permanently deleting an Item as a collection administrator
**Describe the bug**
(Copied from [Test Plan, Item T030](https://docs.google.com/spreadsheets/d/1BdK8SzytA7HoxOZgldCIHabMxeR4mYTXt_ynVT_-KR8/edit#gid=1865680815)):

> When trying to delete an item as a collection admin, I get a red message saying "An error occurred while deleting the item
".  The API returns a 403 status response

**To Reproduce**
1. Log in as a collection administrator
2. Go to an item page in the collection that you are administrator of
3. Click the pencil icon on the top right corner
4. Click "Permanently delete"
5. Confirm de action

**Expected behavior**
The item should be successfully deleted

MarieVerdonck:
## References
* Fixes #3262 

## Description
- Pre-authorize of `ItemRestRepository#delete` changed to actually check the user's permissions instead of restricting to admins
- IT's added to test functionality of collection admin being able to delete item of collection it is item to (& also items with bitstreams)

Some related issues that popped up:

### Issue - Collection admin can't delete item with bitstreams (resolved)
- Happens on `authorizeService.authorizeAction(context, bitstream, Constants.DELETE);` in `BitstreamServiceImpl#delete`
- => In `BitstreamServiceImpl#getParentObject` the bitstream doesn’t have any bundles anymore, since they’ve already been deleted => Can be fixed by moving `bundle.clearBitstreams()` to after bitstreams have been deleted in `BundleServiceImpl#delete`

### Issue - [BundleRestRepositoryIT#deleteBundle](https://github.com/DSpace/DSpace/blob/main/dspace-server-webapp/src/test/java/org/dspace/app/rest/BundleRestRepositoryIT.java) (resolved)

The above IT created an item with 2 bitstreams and then a separate bundle with those same 2 bitstreams in and then deletes the separate bundle, and assumes the two bitstreams will also be deleted. 
But this is not the case since the item's ORIGINAL bundle still contains the bitstreams, so they in itself should not be deleted, just removed from the separate bundle, see [here](https://github.com/atmire/DSpace/blob/w2p-79369_fix-error-when-permanently-deleting-item-as-collection-admin/dspace-api/src/main/java/org/dspace/content/BundleServiceImpl.java#L207). So the *assumption of this test* that the bitstreams in itself should also be removed, despite still being present in another bundle (ORIGINAL of item), *was false* and has been changed to just deleting the ORIGINAL bundle of the item.

Previously this worked because the `bundle.clearBitstreams()` in `bundleServiceImpl` was done *before* the bitstreams were deleted, so they then were already removed from the alternate bundle, and thus were only attached to 1 bundle (supposedly the one being removed, but in actuality it was ORIGINAL of item), so could be removed permanently.

### Issue - Deleted bitstream access with no auth token

We noticed there is a typo [here](https://github.com/DSpace/DSpace/blob/main/dspace-server-webapp/src/test/java/org/dspace/app/rest/ItemRestRepositoryIT.java#L1316) (biststreams instead of bitstreams) which inaccurately confirmed that a deleted bitstream access with no auth token gives a 404, likely because of the `BitstreamRestRepository#findOne` pre-authorize: `hasPermission(#id, 'BITSTREAM', 'METADATA_READ')`. So previously it returned a 404 not because the bitstream doesn't exist, but because the endpoint doesn't exist.
If a non logged in user access to a deleted/non-existent bitstream should return 404 this can be handled separately to this PR. Currently just checked it returns a 404 with the adminToken, since we just want to verify the bitstream is gone, regardless of access issues.

## Instructions for Reviewers
Check if a collection admin can permanently delete an item:
Backend:
- Make a non-admin user the collection admin of a testing collection => Add an item to that collection => perform `DELETE {{url}}/server/api/core/items/{uuid} with the user/password` of that collection admin user => should succeed
- Try the same with a user that is not the collection admin => Forbidden

Frontend: see steps to reproduce in issue: https://github.com/DSpace/DSpace/issues/3262

## Checklist

### DIFF
--- a/dspace-api/src/main/java/org/dspace/content/BundleServiceImpl.java
+++ b/dspace-api/src/main/java/org/dspace/content/BundleServiceImpl.java
@@ -501,10 +501,10 @@ public class BundleServiceImpl extends DSpaceObjectServiceImpl<Bundle> implement
 
         // Remove bitstreams
         List<Bitstream> bitstreams = bundle.getBitstreams();
-        bundle.clearBitstreams();
         for (Bitstream bitstream : bitstreams) {
             removeBitstream(context, bundle, bitstream);
         }
+        bundle.clearBitstreams();
 
         List<Item> items = new LinkedList<>(bundle.getItems());
         bundle.getItems().clear();
--- a/dspace-server-webapp/src/main/java/org/dspace/app/rest/repository/ItemRestRepository.java
+++ b/dspace-server-webapp/src/main/java/org/dspace/app/rest/repository/ItemRestRepository.java
@@ -147,7 +147,7 @@ public class ItemRestRepository extends DSpaceObjectRestRepository<Item, ItemRes
     }
 
     @Override
-    @PreAuthorize("hasAuthority('ADMIN')")
+    @PreAuthorize("hasPermission(#id, 'ITEM', 'DELETE')")
     protected void delete(Context context, UUID id) throws AuthorizeException {
         String[] copyVirtual =
             requestService.getCurrentRequest().getServletRequest()
--- a/dspace-server-webapp/src/test/java/org/dspace/app/rest/BundleRestRepositoryIT.java
+++ b/dspace-server-webapp/src/test/java/org/dspace/app/rest/BundleRestRepositoryIT.java
@@ -537,11 +537,7 @@ public class BundleRestRepositoryIT extends AbstractControllerIntegrationTest {
                     .build();
         }
 
-        bundle1 = BundleBuilder.createBundle(context, item)
-                .withName("testname")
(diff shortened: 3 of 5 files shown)


abollini (approved):
thanks @MarieVerdonck changes look good to me by code inspection

tdonohue (approved):
👍 Looks good. Thanks @MarieVerdonck and @bruno-atmire !  I've tested that the old bug no longer exists -- it doesn't, I can now delete Items as a Collection Admin (as long as the Item is owned by my Collection).

